PR #17, opened 2024-09-25
Update UX/UI from designs

clostao:
* Removes IPLD node explorer feature (as it was useful for early-PoC)
* Adds drag&drop files and folders and file upload management
* Refactor UI


clostao:
I've changed the main parts of the UX/UI but something are yet to be implemented

frontend/src/components/Files/RecentActivities.tsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+
+interface Activity {
+  user: string;
+  action: "uploaded" | "shared" | "deleted";
```
marc-aurele-besner, line 5: Is the term "deleted" accurate?
I'm assuming we are only deleting the cache version, so maybe something more like "flagged" or "reported"?
It could be misleading users in giving them the impression they can actually delete data...

frontend/src/components/common/FileTable.tsx
```diff
@@ -0,0 +1,215 @@
...
+import { ChevronDown, ChevronRight, Trash2 } from "lucide-react";
+import { UploadedObjectMetadata } from "../../models/UploadedObjectMetadata";
+import { ApiService } from "../../services/api";
+
+export const FileTable: FC<{ files: UploadedObjectMetadata[] }> = ({
+  files,
+}) => {
+  const [expandedRows, setExpandedRows] = useState<Set<string>>(new Set());
+
+  const toggleRow = (id: string) => {
```
marc-aurele-besner, line 13: These functions and render fn will likely benefit from memoization

frontend/src/components/Files/RecentActivities.tsx
clostao: Was more kind of a mock

marc-aurele-besner (approved):
LGTM 🚀

PR #23, opened 2024-09-27
Add object ownership management

clostao:


frontend/src/app/drive/layout.tsx
```diff
@@ -0,0 +1,96 @@
...
+  variable: "--font-geist-mono",
+  weight: "100 900",
+});
+
+export const metadata: Metadata = {
+  title: "Auto-Drive",
+  description: "Autonomys Drive",
+};
+
+export default function RootLayout({
```
marc-aurele-besner, line 32: Is the layout of this page completely different?
Otherwise, you could use`Layout` instead of `RootLayout` to use more of an island architecture and only add layout needed for this specific section.

marc-aurele-besner: I saw lower that you removed most of the code from the RootLayout, then here you could change for `Layout` and simply get rid of the already defined html and body tag and it should be gucci

frontend/src/app/page.tsx
```diff
@@ -1,42 +1,109 @@
...
+            disabled={isLoading}
+            className="bg-black hover:bg-gray-800 text-white font-bold py-3 px-6 rounded-full 
+                       border-2 border-transparent hover:border-white focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-900
+                       transition duration-300 ease-in-out transform hover:-translate-y-1 hover:scale-105
+                       flex items-center justify-center w-full max-w-xs"
+            aria-label="Sign in with Google"
+          >
+            {isLoading ? (
+              <div className="flex items-center justify-center">
+                <svg
```
marc-aurele-besner, line 58: Can we put svg either in a .svg or in a .tsx file with the name of what they are, to improve readability, please.

frontend/src/app/drive/layout.tsx
```diff
@@ -0,0 +1,96 @@
...
+                <ScopeSwitch />
+                <span className="text-sm font-medium">Global</span>
+                <div className="md:w-80">
+                  <SearchBar />
+                </div>
+              </div>
+            </header>
+            <div className="flex flex-1 overflow-hidden">
+              <aside className="w-16 md:w-64 space-y-4">
+                <a href="/drive" className="contents">
```
marc-aurele-besner, line 64: In next, especially when navigating to internal page, <Link should be use to avoid a full page reload

marc-aurele-besner (commented):
Awesome, a few comments and suggestion

jfrank-summit (commented):
I only reviewed the backend. I have a few comments/questions.

backend/src/api.ts
```diff
@@ -0,0 +1,292 @@
+import bodyParser from "body-parser";
+import cors from "cors";
+import express from "express";
+import multer from "multer";
+import { isJson } from "./utils/index.js";
```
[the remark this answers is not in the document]
clostao: Using es modules mandates to

backend/src/api.ts
```diff
@@ -0,0 +1,292 @@
...
+  cidToString,
+  IPLDNodeData,
+  OffchainMetadata,
+} from "@autonomys/auto-drive";
+import { decode } from "@ipld/dag-pb";
+import "dotenv/config.js";
+import { FolderTreeSchema, NodeWithMetadata } from "./models/index.js";
+import { transactionResultsRepository } from "./repositories/index.js";
+import { handleAuth } from "./services/authManager/express.js";
+import { uploadManager } from "./services/uploadManager/index.js";
```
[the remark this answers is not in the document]
clostao: I've separated into two different files: api and worker

backend/src/repositories/ownership.ts
```diff
@@ -0,0 +1,42 @@
+import { getDatabase } from "../drivers/sqlite.js";
+
+const setUserAsOwner = async (cid: string, userId: string) => {
+  const db = await getDatabase();
+
+  await db.run(
+    "INSERT INTO object_ownership (cid, user_id, is_admin, marked_as_deleted) VALUES (?, ?, ?, ?)",
```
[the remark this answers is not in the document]
clostao: You're right the primary key should be the pair (cid, user_id)

backend/src/api.ts
```diff
@@ -0,0 +1,292 @@
...
+    console.log(
+      `Setting Content-Disposition to attachment with filename: ${metadata.name}`
+    );
+    res.set("Content-Disposition", `attachment; filename="${metadata.name}"`);
+  }
+};
+
+const createServer = async () => {
+  const app = express();
+  const port = 3000;
```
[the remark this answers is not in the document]
clostao: I'll add it as an optional env variable having the default being 3000

backend/src/api.ts
```diff
@@ -0,0 +1,306 @@
...
+    res.set("Content-Disposition", `attachment; filename="${metadata.name}"`);
+  }
+};
+
+const createServer = async () => {
+  const app = express();
+  const port = Number(process.env.PORT) || 3000;
+
+  // Increase the limit to 10MB (adjust as needed)
+  const requestSizeLimit = process.env.REQUEST_SIZE_LIMIT || "10mb";
```
marc-aurele-besner, line 46: Can we add a .env or .env.sample (depending on if you also have .env that are true secret) so we have, somewhere in the repo, a list of all the .env defined in the code?

backend/src/drivers/sqlite.ts
```diff
@@ -21,6 +21,11 @@ const initSchema = async (db: Database) => {
   await db.run(
     "CREATE TABLE IF NOT EXISTS transactionResults (cid TEXT PRIMARY KEY, transaction_result TEXT, head_cid TEXT)"
   );
+
+  // Object Ownership
+  await db.exec(
+    "CREATE TABLE IF NOT EXISTS object_ownership (cid TEXT, user_id TEXT, is_admin BOOLEAN, marked_as_deleted BOOLEAN, PRIMARY KEY (cid, user_id))"
```
marc-aurele-besner, line 27: I'm not sure if I see the utility of having `is_admin` directly in the object_ownership
First, this property seems to never be use anywhere
Second, user role should be tied to the user object, and not the content they uploaded.

If today I'm an admin, there is chance that at a later date, I lost that role and become a simple user, that will be weird to still have files flag as admin and could lead to logic error

marc-aurele-besner: I think we could improve on `marked_as_deleted` to use a dated [soft delete approach](https://hasura.io/docs/2.0/schema/common-patterns/data-modeling/soft-deletes/) so a `deleted_at` and while we are at it, I believe adding a `created_at` (and for a table that could be edited `updated_at`)

backend/src/drivers/sqlite.ts
marc-aurele-besner, line 1: Do we have a users table?

marc-aurele-besner (commented):
Using email as user_id will create issue and security risk as soon as we implement multiple OpenID provider.
I have friends for example that share their twitter with their assistant, but not their Gmail. If we decide to implement twitter OAuth, now both flow will open the account.

clostao:
> Using email as user_id will create issue and security risk as soon as we implement multiple OpenID provider. I have friends for example that share their twitter with their assistant, but not their Gmail. If we decide to implement twitter OAuth, now both flow will open the account.

Sure, I thought that was good enough for the PoC but I feel like if we are introducing the subscription management for the PoC then its better to implement the correct user management already

backend/src/drivers/sqlite.ts
clostao: Not yet, but I'm gonna create it

backend/src/drivers/sqlite.ts
clostao: Makes sense

clostao: The updated_at, created_at I'll add it when I migrate DB to postgres

clostao:
@jfrank-summit @marc-aurele-besner I've updated the userId to a composed id (oauth_provider, oauth_user_id). The management of the CRUD of users will be in a new PR though

marc-aurele-besner (approved):
LGTM 🚀

jfrank-summit (approved):
LGTM

PR #30, opened 2024-10-01
Fix scope switch

clostao:
Scope switch based purely on the localStorage reactiveness didn't work well. I used an intermediate react state for enhancing how it works


clostao:
> Maybe worth considering using zustand to have a state management that can persist the state in localStorage if needed

I thought about using recoil that also supports that, I decided not includes just for this one, but for sure we should include It then.

Issue #5, opened 2024-09-18
[Front] Download folders from Backend


its-colby:

clostao:
Files are downloadable right now.

clostao:
How should folder download would work? Compressed to zip files?

its-colby:
Good question. Let's have folder download compress it to a zip file. That seems to be the standard for this kind of thing.

clostao:
Backend implementation at: https://github.com/autonomys/auto-drive/pull/28

PR #32, opened 2024-10-01
Add folder download to frontend

clostao:
Added folder download to frontend and couple small fixes


clostao:
Solves: https://github.com/autonomys/auto-drive/issues/5

frontend/src/app/page.tsx
```diff
@@ -16,7 +16,7 @@ export default function App() {
     // These are example parameters. Replace with your actual values
     const params = new URLSearchParams({
       client_id: process.env.NEXT_PUBLIC_GOOGLE_AUTH_CLIENT_ID!,
-      redirect_uri: "http://localhost:8080/api/auth/callback/google",
+      redirect_uri: `${window.location.origin}/api/auth/callback/google`,
```
marc-aurele-besner, line 19: I don't want to make it a blocker, if this work, this is fine with me.
But since next-auth is setup, why not simply using signIn('google') from next-auth when clicking on the google signin button?

marc-aurele-besner (approved):

frontend/src/app/page.tsx
clostao: Right, much cleaner solution

PR #36, opened 2024-10-02
Changed SQLite for Postgres 

clostao:
As @jfrank-summit suggested I'm changing sqlite for a more robust DB, PostgreSQL since we already have an instance for demos and its quite efficient.

I separated in a subproject the migrations with db-migrate. I have to create a subpackage because `db-migrate` cannot be used in a ES Module package.


marc-aurele-besner (approved):
Awesome, personally I'll also suggest adding Hasura, this maybe a bit lower priority but this allows using graphql in the frontend (with like apollo) and create permissions on table that are tied to the userId and the content of the JWT.

clostao:
Sounds good I'll explore It once we got the PoC phase done

Issue #9, opened 2024-09-18
[Front] View user content


its-colby:
- Enable users to see a dashboard view of all the files and subfolders within their root folder
- allow users to explore and click into these subfolders
- enable users to click and see the metadata of the files

PR #37, opened 2024-10-03
Updated entities and controllers

clostao:
This PR extracts the API controllers into a folder (for the moment just one), defines more clearly the object information entity and complies better with rest api schema.

This is the base for the #9 issue


marc-aurele-besner (approved):
LGTM

PR #38, opened 2024-10-03
Update to new entities and controllers

clostao:
I've integrated the changes in the backend at #37 and added the content view included it #9

Issue #27, opened 2024-10-01
[Front/Back] Shareable objects


clostao:

clostao:
@its-colby how a user shares an object with other? Should it be a list of all the users and select from there? The email may identify to several accounts: (google and discord account)

its-colby:
I will assume the term "object" means "files and folders".

**POC**
I see that you have tagged this issue as "POC". I believe, for the POC, we should only implement the following. Provide an option on the frontend to copy a link to a file or folder. Then, they can share this link with others. When the link is opened, it takes them to the relevant file or folder.

Every file and folder will be in plaintext on the POC and there will be no editing/creation IPNS names. Thus, technically, owners have no special permissions. So, "sharing", in terms of "sharing permissions", doesn't mean much.

**MVP**
For the MVP, we will likely not add sharing capabilities either. I don't imagine that sharing editing permissions will be something desired by our future partners. Nonetheless, we will have to wait and see after we talk to them.

If we did decide to have "sharing" in our MVP, I think it will likely work with the following UX. User A uses a messaging application to ask his friend for his handle. User B sends the handle (copy-pasted from profile window of Auto Drive frontend). User A uses Auto Drive frontend to share the relevant file/folder.

What will be this "handle"? I don't think it should be something users create. I think it should be the ID users receive within the OIDC system after completing registration.

clostao:
Pending to change the user-specified handle by a random id

its-colby:
> Pending to change the user-specified handle by a random id

Not sure I understand what you mean here. Do you mean, you plan to change the user-specified handle "to" (not "by") a random id?

clostao:
No, I mean that the current handle (user identificator) is setup by the user but you previously told me to update for an automatically created ID 

> What will be this "handle"? I don't think it should be something users create. I think it should be the ID users receive within the OIDC system after completing registration.

its-colby:
Understood

PR #39, opened 2024-10-04
Shareable objects and user handle management

clostao:
For sharing objects, some user-friendly id it's needed for sharing objects between users. Since email wouldn't work because it could represent several accounts I created a soft handle-based ID. With soft I mean that is updatable.

Enable object admins to share objects. Depends on #38 and implements #27


backend/src/useCases/ownership.ts
```diff
@@ -13,8 +13,18 @@ const setObjectAsDeleted = async (user: User, cid: string) => {
   await ownershipRepository.setObjectAsDeleted(user.provider, user.id, cid);
 };
 
+const getOwners = async (cid: string) => {
+  return ownershipRepository.getOwners(cid);
```
marc-aurele-besner, line 17: I'm not sure that I understand why these just forward to a other function with the same name?
Why not just using getAdmins from ownershipRepository where you need it?

Furthermore, declaring the same function name in two place in the same project make it a bit harder to search across the codebase and understand the flow

clostao: It's true that I could just call `ownershipRepository.getOwners(cid)` but I expected this function to get more complex in few iterations, e.g Once we have `users` table not returning `Ownership` entries. 

Generally speaking I prefer adding the two layers (repository and use case) because usually the use case functions get more complex within time and just from personal taste I prefer not to call a `B` repository in the context of entity `A`

backend/src/useCases/ownership.ts
```diff
@@ -1,20 +1,63 @@
...
+    console.log("users", users);
+
+    throw new Error("Inconsistent database state");
+  }
+
+  const safeUsers = users.map((user) => user!);
+
+  return safeUsers.map((user, index) => ({
+    handle: user.handle,
+    role: ownerships[index].is_admin ? "admin" : "viewer",
```
marc-aurele-besner, line 49: Could we move the roles into an enum somewhere that everything can access them, instead to have them hardcoded everywhere?
(and if we base our poc on google drive, then maybe we will want to add a commenter role in the future?)

backend/migrations/sqls/20241002111606-init-schema-up.sql
```diff
@@ -15,6 +15,13 @@ CREATE TABLE transaction_results (
   transaction_result jsonb,
   head_cid TEXT
 );
+-- Users
+CREATE TABLE users (
```
marc-aurele-besner, line 19: Do we see a use case for keeping user email?
For example, when an Alice share a file with Bob, we could send an email notification to Bob, or if Bob request access to a file Carl has uploaded?

backend/migrations/sqls/20241002111606-init-schema-up.sql
```diff
@@ -15,6 +15,13 @@ CREATE TABLE transaction_results (
...
+  oauth_provider TEXT,
+  oauth_user_id TEXT,
+  handle TEXT,
```
marc-aurele-besner, line 22: There does not appear to be any constraint to `handle`, I'm assuming we want that column to be unique

backend/src/useCases/users.ts
```diff
@@ -0,0 +1,72 @@
+import {
+  OAuthUser,
+  User,
+  userFromOAuth,
+  userFromTable,
+} from "../models/index.js";
+import { usersRepository } from "../repositories/index.js";
+
+const updateUserHandle = async (
```
marc-aurele-besner, line 9: I think we should add some input validation for handle, trimming, excluding problematic symbol, length min/max, ...

marc-aurele-besner (changes requested):

backend/src/useCases/ownership.ts
clostao: You're right

clostao: For the moment, I'll use only "admin" and "viewer" for simplifying a bit the first version

backend/migrations/sqls/20241002111606-init-schema-up.sql
clostao: Yes, it makes sense I thought about introducing other data such as full name, email, profile photo in other PR

backend/migrations/sqls/20241002111606-init-schema-up.sql
clostao: Thanks for the catch

marc-aurele-besner (approved):
Can't wait to park the @m handle 😂

backend/src/controllers/user.ts
```diff
@@ -18,6 +18,25 @@ userController.post("/@me/update", async (req, res) => {
...
+
+  if (handle.length > 32) {
+    return res.status(400).json({
+      error: "Handle must be 32 characters or less",
+    });
+  }
+
+  if (handle.length > 1) {
+    return res.status(400).json({
+      error: "Handle must be 1 characters or more",
```
marc-aurele-besner, line 36: this is not true I won't be able to park @m 😭

marc-aurele-besner (approved):

backend/src/controllers/user.ts
marc-aurele-besner: Actually the condition will make any handle above 1 length fail

marc-aurele-besner (changes requested):

marc-aurele-besner (approved):
LGTM 🚀

Issue #24, opened 2024-09-27
[Front] Disable Share and Download button, till upload is finish


marc-aurele-besner:
## [Front] Disable Share and Download button, till upload is finish

If pressing download before the upload is finish, it result in a error, and

PR #44, opened 2024-10-09
Fix upload status

clostao:
Upload status relied on `transactionResult` table for getting the number of nodes for a certain upload (rootCID). The problem is that since the object at first wasn't published on-chain a full uploaded object was shown in the frontend. Now `nodes` are related to a `rootCID` and `headCID`.

`RootCID`: The ID of an upload operation.
`HeadCID`: The ID of each of the files or folders uploaded within an upload operation.

This additionally solves #24 because there's no need to wait to on-chain upload for sharing or downloading.


marc-aurele-besner (dismissed):

marc-aurele-besner (approved):
LGTM

PR #45, opened 2024-10-10
UX Improvements

clostao:
I've added lots of small changes that improves the UX of the app. I'd recommend to review the PR by commit because every commit explains itself.


marc-aurele-besner (approved):
Awesome, LGTM 🚀

PR #46, opened 2024-10-11
Implement delete & restore objects

clostao:


backend/src/controllers/object.ts
```diff
@@ -203,6 +225,46 @@ objectController.get("/:cid/download", async (req, res) => {
...
+    res.sendStatus(200);
+  } catch (error: any) {
+    console.error("Error deleting object:", error);
+    res
+      .status(500)
+      .json({ error: "Failed to delete object", details: error.message });
+  }
+});
+
+objectController.post("/:cid/restore", async (req, res) => {
```
marc-aurele-besner, line 248: So if a user mark a file as deleted, obviously it's a shallow delete, but the cid still should disappear from the UI correct?
If so, when will the restore be in use? If the user upload a second time, the same file?

marc-aurele-besner (approved):
just one question on the restore flow

backend/src/controllers/object.ts
clostao: In the Trash view there's a button to restore. If the same user uploads the same file the file would be effectively restored. Since the entry in the `ownership` table is updated in case of conflict with primary key `(oauth_provider, oauth_user_id, cid)`

marc-aurele-besner (approved):

PR #53, opened 2024-10-14
Fix download content disposition

clostao:
Some characters are forbidden in file download so this parser should be used


marc-aurele-besner (approved):
LGTM

PR #58, opened 2024-10-17
Credit management (backend)

clostao:
Implemented backend changes described in #8:

Admin users management:

* Admin privileges are implemented adding a field named `"role"` to `users` table
* An admin could update a user's role
* Admins could see a list of all the user's 

Credit management:

* Added two more columns to `user` table: `download_credits` and `upload_credits` for limit user's activity.
* Users with `AdminRole` could increment upload and download credits.

Api Keys:

* Added a new table that keeps track of the ApiKeys and the userId. An apikey is a user itself and is not related with a certain oauth account. 
* Since api keys are user's they have a usage limit too


backend/migrations/sqls/20241016182736-add-observality-down.sql
```diff
@@ -0,0 +1,26 @@
...
+ALTER TABLE transaction_results DROP COLUMN updated_at;
+
+ALTER TABLE users DROP COLUMN created_at;
+ALTER TABLE users DROP COLUMN updated_at;
+
+ALTER TABLE object_ownership DROP COLUMN created_at;
+ALTER TABLE object_ownership DROP COLUMN updated_at;
+
+ALTER TABLE api_keys DROP COLUMN created_at;
+ALTER TABLE api_keys DROP COLUMN updated_at;
```
marc-aurele-besner, line 24: last_upload_at, last_download_at could be good to have
even a usage counter

marc-aurele-besner (commented):
This seems very cater to individual use and not organization use at all.
Even if it's maybe out of spec, I feel we should at least design the db and structure to support both.

I'll suggest (if out of spec) also creating a org table, and automatically creating a org for each users and tied the api to the org+user, this way we will be able to extend functionality more easily in the future

clostao:
Btw, I misunderstood sth in the issue #8 the limits are intended to be monthly limit instead of a fixed limit. The prev comment still applies

clostao:
Updates:

As we discussed internally, a single-member organization is created when a user "onboards" (signs up). Each organization has a subscription that has passed from having a counter of `downloadLimit` and `uploadLimit` to have a `limit` and a `granularity` (the frequency with which that limit is renewed).

For accounting the uploads and downloads there is the table `interactions` that it's used for calculating the remaining credits.

backend/migrations/sqls/20241016182736-add-observality-down.sql
clostao: Has been outdated since now there's an entry in table `interactions` for each r/w op

marc-aurele-besner (dismissed):
LGTM

marc-aurele-besner (approved):

PR #64, opened 2024-10-22
Search by files by name 

clostao:
Enable users to search files by their name through the search bar

### DIFF
--- a/frontend/src/models/ObjectSearchResult.ts
+++ b/frontend/src/models/ObjectSearchResult.ts
@@ -1,0 +1,4 @@
+export type ObjectSearchResult = {
+  cid: string;
+  name: string;
+};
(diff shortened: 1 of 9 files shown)


frontend/src/models/ObjectSearchResult.ts
marc-aurele-besner, line 1: It's kind of redundant to have these type on both side, at one point we should setup a local packages for fixtures that both the fe and be could import shared types and utils fn

marc-aurele-besner (approved):
LGTM

frontend/src/models/ObjectSearchResult.ts
clostao: Sure, type management between be and fe is always a great pain point

Issue #56, opened 2024-10-14
[Front — Enhancement] Metadata View


its-colby:
Currently, the metadata view (nodes, block depth, etc) is only available on hover. Please allow users to click into the metadata and view it as a full page. Perhaphs include a button next to "download" "delete" etc that says "properties" or something. This view should also have "question mark" icons next to each property that allow you to read a tooltip explaining, for example, what "minimum block depth" is

Issue #55, opened 2024-10-14
[Front — Small Addition] Sharing Files/Folders should include link


its-colby:
Sharing should not "exclusively" be by username. Similar to Google Drive, users should be able to copy a link or search by handle to share.

clostao:
How would this work along the download credit limitation feature. For implementing the monthly download limit we need the user to be authenticated for downloading a file (for reducing the user's download credits). If the user generates a link to the file the person that receives, should be able to download the file? If that's the case who would "pay" for that download?  

@its-colby

its-colby:
Not necessarily the case. If a non-registered user (or user without enough credits) views the link, they should be shown the appropriate error message. This will need to be done in the route handling on the backend. @clostao

its-colby:
In the Profile section, there should also be a "copy handle" button

Issue #54, opened 2024-10-14
[Front — Bug Fix] Font Colors


its-colby:
Fix UI font colors to be more visible (higher contrast) on Safari

clostao:
This should've been solved in #61

its-colby:
At least on Safari, it does not seem to be the case. But, perhaps I was using a cached version of the site.

clostao:
Please provide screenshots

Issue #50, opened 2024-10-14
[Front] Implementation of Diana's POC Figma Mock-Up


its-colby:

its-colby:
Things to add:

- When toggling between global and my files, the loading symbol, as well as messages like "there is nothing in your trash" are extremely transparent and hard to read on Safari.
- On Safari, the header (the toggle between global and my files + Auto Drive title/logo) is very hard to see. The font color is super transparent.
- On Safari, once the API key is generated, it is very transparent and too hard to read
- Once the API key is generated, I had to refresh the page in order to see it in "my keys" section. This should happen without me needing to refresh.
- "Delete" should be impossible on global files that aren't your own. When viewing global files/folders, the "delete" button should only show up for your own files/folders.
- Please make the "user" button appear right below "files", "shared", "trash". It should not appear at the very bottom. The user button should also show more information about the credit management system, such as more exact numbers for uploading/downloading history (like 4.853GB etc instead of rounding to 5GB, and number of files downloaded/uploaded, etc). 
@clostao

its-colby:
Additional Things:
When a user hovers over a particular file, they will see a UI card containing metadata information pop-up. This is good, however, it should be clickable. In other words, there should be a button that says "see more" or something along these lines. When clicked, this is where the "pin" button should be (I'm referring to another GitHub Issue about a pinning service). @clostao

clostao:
Could you provide screenshots please @its-colby

clostao:
I can see this in Safari: ![Image](https://github.com/user-attachments/assets/7639e784-bb28-42fb-920b-f1d46827766d) Is this problematic or you're seeing other thing?

its-colby:
<!-- Uploading "Screenshot 2024-10-22 at 9.01.04 PM.png"... -→

its-colby:
<!-- Uploading "Screenshot 2024-10-22 at 9.01.57 PM.png"... -→

PR #65, opened 2024-10-23
Small fixes from UI

clostao:
Should solved #56, #50, #55, #54

### DIFF
--- a/frontend/src/app/drive/layout.tsx
+++ b/frontend/src/app/drive/layout.tsx
@@ -92,6 +92,12 @@ export default function AppLayout({
                   <span className="hidden md:block">Trash</span>
                 </button>
               </InternalLink>
+              <InternalLink className="contents" href="/drive/profile">
+                <button className="flex items-center space-x-2 text-black hover:text-blue-600 mb-2">
+                  <UserIcon className="w-5 h-5" />
+                  <span className="hidden md:block">Profile</span>
+                </button>
+              </InternalLink>
               <RoleProtected roles={[UserRole.Admin]}>
                 <InternalLink className="contents" href="/drive/admin">
                   <button className="flex items-center space-x-2 text-black hover:text-blue-600 mb-2">
@@ -114,16 +120,6 @@ export default function AppLayout({
             <main className="flex-1 overflow-auto p-6">{children}</main>
           </UserEnsurer>
         </div>
-        <footer className="mt-8 flex justify-start w-16 md:w-64 flex-col gap-2 px-10 py-2">
-          <div className="flex">
-            <InternalLink href="/drive/profile">
-              <button className="flex space-x-2 text-gray-700 hover:text-blue-600">
-                <UserIcon className="w-5 h-5" />
-                <span className="hidden md:block">User</span>
-              </button>
-            </InternalLink>
-          </div>
-        </footer>
       </div>
     </div>
   );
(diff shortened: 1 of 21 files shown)


frontend/src/components/ApiKeysTable/index.tsx
```diff
@@ -34,9 +34,18 @@ export const ApiKeysTable = ({
 
   const nonDeletedApiKeys = apiKeys?.filter((apiKey) => !apiKey.deletedAt);
 
+  const onSuccess = useCallback(() => {
+    closeCreationModal();
+    window.location.reload();
```
marc-aurele-besner, line 39: Why do we need to force refresh the page?
There is likely a better way to do this with NextJS (`router.reload()`)

frontend/src/app/drive/search/[cid]/page.tsx
```diff
@@ -6,12 +6,13 @@ import { UploadedObjectMetadata } from "@/models/UploadedObjectMetadata";
...
 import { useEffect, useMemo, useState } from "react";
 import { useLocalStorage } from "usehooks-ts";
+import { useScopeStore } from "../../../../states/scope";
 
 export default function Page({ params: { cid } }: { params: { cid: string } }) {
   const [objectsMetadata, setObjectsMetadata] =
     useState<UploadedObjectMetadata[]>();
   const [error, setError] = useState<string>();
-  const [scope] = useLocalStorage<"user" | "global">("search-scope", "global");
+  const { scope } = useScopeStore();
```
Comment (marc-aurele-besner) [inline frontend/src/app/drive/search/[cid]/page.tsx:15]:
```suggestion
  const scope = useScopeStore((state) => state.scope);
```
You will gain better performance to not deconstruct the state and just get the exact property you need (otherwise you will get re-render on all the store update)

frontend/src/components/UploadedObjectInformation/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { ObjectDeleteModal } from "../Files/ObjectDeleteModal";
+
+export const UploadedObjectInformation = ({
+  object,
+}: {
+  object: UploadedObjectMetadata | null;
+}) => {
+  const [shareModalCid, setShareModalCid] = useState<string | null>(null);
+  const [deleteModalCid, setDeleteModalCid] = useState<string | null>(null);
+  const { user } = useUserStore();
```
marc-aurele-besner, line 21:
```suggestion
  const user = useUserStore((state) => state.user);
```
same reason as above

marc-aurele-besner (changes requested):

frontend/src/app/drive/layout.tsx
```diff
@@ -23,7 +23,7 @@ export default function AppLayout({
 }: Readonly<{
   children: React.ReactNode;
 }>) {
-  const { subscription } = useUserStore();
+  const subscription = useUserStore(({ subscription }) => subscription);
```
marc-aurele-besner, line 26:
```suggestion
  const subscription = useUserStore((state) => state.subscription);
```
https://zustand.docs.pmnd.rs/getting-started/introduction

marc-aurele-besner (changes requested):

frontend/src/components/ApiKeysTable/index.tsx
clostao: The method router.refresh() [doesn't reload the site it just refetch the page](https://github.com/vercel/next.js/discussions/62146#discussioncomment-8493677) and merges the new response with the previous one.

For that mechanism to actually reload the page, the API request should be on the server rather than in the client. That's not the case for some components of the application. I'll add an issue to update this structure but I think it'd be better to update it in another PR.

marc-aurele-besner (approved):
LGTM

PR #71, opened 2024-10-28
Add node publisher after unintended removal 

clostao:


backend/src/worker.ts
```diff
@@ -2,3 +2,4 @@ import { nodeMigrator } from "./services/nodeMigrator/index.js";
 import { onchainPublisher } from "./services/onchainPublisher/index.js";
 
 nodeMigrator.start();
+onchainPublisher.start();
```
jfrank-summit, line 5: The name of the PR is "remove node publisher" but this is adding a publisher. What am I missing?

clostao: A little confusing, it fixes the unintended removal of node publisher

jfrank-summit (approved):

PR #72, opened 2024-10-29
Client-Side Encryption & Compression

clostao:
This PR implements changes in auto-drive package in this [PR](https://github.com/autonomys/auto-sdk/pull/140) for enabling users to compress and encrypt with password in the client.

Some of methods in encryption.ts and compression.ts should be included in the auto-drive package but I prefer to have this ones yet here until the split of auto-drive in two takes places (few days).


clostao:
This PR is ready to be reviewed though is failing because the package used for @autonomys/auto-drive is not yet published

marc-aurele-besner (approved):
Awesome 👏

clostao:
I've updated the way encrypted folders are uploaded. Instead of uploading a folder with files encrypted we just packed the files on a zip an encrypt the zip file as any other file. The UX of encrypting a folder was confusing.

marc-aurele-besner (approved):

PR #73, opened 2024-10-29
Fix duplicated metadata in folder upload

clostao:


backend/src/useCases/objects/files.ts
```diff
@@ -278,11 +283,9 @@ const handleFolderUploadFinalization = async (
   user: User,
   uploadId: string
 ): Promise<string> => {
-  console.time("generateFolderArtifacts");
```
clostao, line 281: this is just the removal of some unintented logs

Issue #69, opened 2024-10-27
Improve performance of file list query and render


jfrank-summit:
Rendering is currently very slow, and will get more so as the the number of files uploaded increases. Some ideas (non-exhaustive) to improve performance:

- [ ] Implement pagination when querying the backend
- [ ] Only retrieve the necessary fields (it appears all chunk CIDs are fetched for each file)
- [ ] Implement batch requests rather than fetching each file metadata individually

PR #76, opened 2024-10-30
Improve files fetching

clostao:
Solves issue #69 

- Creates table paginator
- Paginates getRoots, getShareRoots, getDeletedRoots requests. 
- The type of these methods has been updated from CIDs (that posteriorly were one by one fetch its metadata) to ObjectSummary that basically composes only the needed data for table representation


marc-aurele-besner (approved):
LGTM

marc-aurele-besner (approved):

PR #77, opened 2024-10-31
 Improve files fetching backend

clostao:


marc-aurele-besner (approved):
scope could be defined as a enum, I see them hardcoded on both be and fe

furthermore, I think it will make a lot of sense to make a structure private package inside this monorepo that both fe and be can import as dependencies to move all the shared types, data structure, enum and const

clostao:
> scope could be defined as a enum, I see them hardcoded on both be and fe
> 
> furthermore, I think it will make a lot of sense to make a structure private package inside this monorepo that both fe and be can import as dependencies to move all the shared types, data structure, enum and const

I'll merge this PR though I'll be adding your suggestions ASAP

PR #84, opened 2024-11-01
Small updates on the UI

clostao:
It's mostly composes of UI Fixes:

* Update progress bar styling in credit limit tracker
* Added a log-out button
* Added global feed view
* Removed header scope switcher
* Translated header search bar into `global feed` and `my files` view
* Conditional rendering of buttons (e.g Not show remove button in the Trash view)

Some other fixes

* Restore file action didn't considered owner a user that marked a file as delete
* Search by cid or name with user scope did include shared files and that's not the desired behaviour

### DIFF
--- a/backend/src/useCases/objects/object.ts
+++ b/backend/src/useCases/objects/object.ts
@@ -56,7 +56,6 @@ const searchMetadataByName = async (
   limit: number = 5,
   filter: { scope: "user"; user: User } | { scope: "global" }
 ): Promise<MetadataEntry[]> => {
-  console.log("filter", filter);
   if (filter.scope === "user") {
     return metadataRepository.searchMetadataByNameAndUser(
       query,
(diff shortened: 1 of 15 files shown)


backend/src/useCases/objects/object.ts
```diff
@@ -56,6 +56,7 @@ const searchMetadataByName = async (
   limit: number = 5,
   filter: { scope: "user"; user: User } | { scope: "global" }
 ): Promise<MetadataEntry[]> => {
+  console.log("filter", filter);
```
marc-aurele-besner, line 59: do we still need this console?

marc-aurele-besner, line 59: do we still need this console?

marc-aurele-besner (dismissed):
LGTM

marc-aurele-besner (approved):

PR #85, opened 2024-11-05
Return CID in upload complete action

clostao:
Complete upload endpoint returned nothing (200 OK) we need the CID to be returned for use cases such as publish a file and automatically send a link to the file.


backend/src/controllers/upload.ts
```diff
@@ -5,6 +5,7 @@ import multer from "multer";
 import { FolderTreeFolderSchema } from "../models/objects/folderTree.js";
 import { uploadOptionsSchema } from "../models/uploads/upload.js";
 import { z } from "zod";
+import { cidToString } from "@autonomys/auto-drive";
```
marc-aurele-besner, line 8: This line is useless
Furthermore, can you add workflow to run eslint to catch these things as part of our ci/cd ?

backend/src/useCases/uploads/uploadProcessing.ts
```diff
@@ -21,6 +21,7 @@ import {
   uploadsRepository,
 } from "../../repositories/uploads/uploads.js";
 import { filePartsRepository } from "../../repositories/uploads/fileParts.js";
+import { CID } from "multiformats";
```
marc-aurele-besner, line 24: do we export this type in auto-sdk? if so we should export it, basically anything we need from multiformats should be exported in our package

marc-aurele-besner (changes requested):
We should have workflows for both BE and FE to run build and lint

backend/src/useCases/uploads/uploadProcessing.ts
clostao: No, they're not exported I'll add an issue on the sdk

clostao:
> We should have workflows for both BE and FE to run build and lint

Yeah, I'll add that in a another PR

marc-aurele-besner (approved):

PR #86, opened 2024-11-05
Add discord oauth provider

clostao:
* Adds discord oauth provider
* From 1 to N oauth providers now is plug & play adding new providers


frontend/src/app/api/auth/[...nextauth]/config.ts
```diff
@@ -1,26 +1,63 @@
...
+import DiscordProvider from "next-auth/providers/discord";
 export const authOptions: AuthOptions = {
   providers: [
     GoogleProvider({
       clientId: process.env.NEXT_PUBLIC_GOOGLE_AUTH_CLIENT_ID as string,
       clientSecret: process.env.GOOGLE_AUTH_CLIENT_SECRET as string,
       checks: [],
     }),
+    DiscordProvider({
+      clientId: process.env.NEXT_PUBLIC_DISCORD_AUTH_CLIENT_ID as string,
```
Comment (marc-aurele-besner) [inline frontend/src/app/api/auth/[...nextauth]/config.ts:12]: I don't seen any reason for the client id to be public, we should remove `NEXT_PUBLIC_` from the env name (same for Google above)

frontend/src/app/api/auth/[...nextauth]/config.ts
```diff
@@ -1,26 +1,63 @@
...
+        ? `${baseUrl}${_url}`
+        : _url.replace(_baseUrl, baseUrl);
+
+      return url;
+    },
   },
   secret: process.env.NEXTAUTH_SECRET,
 };
+
+declare module "next-auth" {
```
Comment (marc-aurele-besner) [inline frontend/src/app/api/auth/[...nextauth]/config.ts:43]: It's usually best practice to override/extend package types in a dedicated declaration file
for next-auth `next-auth.d.ts` in the root of the FE project

backend/src/services/authManager/providers/discord.ts
```diff
@@ -0,0 +1,33 @@
+import { OAuthUser } from "../../../models/users";
+
+interface DiscordUser {
+  id: string;
+}
+
+const getUserFromAccessToken = async (
```
marc-aurele-besner, line 7: Why you implemented your own fetch profile function when next auth has one build-in in the discord provider?

marc-aurele-besner (changes requested):

backend/src/services/authManager/providers/discord.ts
clostao: Does it? The only thing that I can see is the Discord provider that one of the fields of it's the `userinfo` that means the endpoint from which the user info is extracted.

clostao: Oh I see, I see these two cases as two different things:

What I'm doing in my code is checking that the `accessToken` that the user has sent is a correct discord user. So my input is an `accessToken` and the output is Discord's interface for Users. 

Meanwhile, the profile function it's a callback on how to treat the user interface (for mapping it to the desired user DTO the app needs, ensure certain consistency on the backend, etc.). Its input is the already retrieved User rather than the `accessToken`. 

Also, may be important to note is that the backend does not use next-auth it's only used in the frontend.

frontend/src/app/api/auth/[...nextauth]/config.ts
Comment (clostao) [reply frontend/src/app/api/auth/[...nextauth]/config.ts:12]: You're right, I took this from supabase oauth providers

backend/src/services/authManager/providers/discord.ts
marc-aurele-besner: You are right, I had not realized this file was in backend part 🙈, sorry 👍

marc-aurele-besner (approved):
Can we add a .env.sample or .env or define the env we use in the readme, at the moment if we clone the repo, we try to connect, we get an error and someone will need to dig in the code to find the missing env and know how they are call

frontend/README.md
```diff
@@ -1,8 +1,33 @@
...
+
+
+Execute this command for creating the enviroment file and then fulfill every missing entry.
+```
+cp .env.sample .env.local
+```
+
+Once, all this is setup you can launch it with:
 
 ```bash
```
marc-aurele-besner, line 32: it's missing the dependencies' installation step

marc-aurele-besner (dismissed):

marc-aurele-besner (approved):

marc-aurele-besner (approved):

Issue #49, opened 2024-10-14
[Front] Local Cache


its-colby:
Utilizing Local Storage, implement a local cache for files and folders, so that the user does not need to constantly query for frequently used files/folders.

clostao:
localStorage has a maximum size of 5MB per website. 

Might be a better option [IndexedDB](https://medium.com/@songusepython/the-best-way-of-browser-storage-more-than-localstorage-and-indexdb-d275cf78589a#:~:text=IndexedDB%20has%20a%20much%20larger,article%20is%20to%20introduce%20localForage!)

its-colby:
Yep, IndexedDB looks like the right approach. For now, you can use whatever is more convenient to develop, since its a low priority. But, long term, we will probably want IndexedDB. @clostao

PR #92, opened 2024-11-13
Local Files Cache

clostao:
As describe in issue #49, a cache for downloaded files has been implemented. Files of >150MB are not to be cached since it may provoke memory problems in some clients.


marc-aurele-besner (approved):
LGTM

PR #93, opened 2024-11-13
fix: search bar display

clostao:
In a previous PR I updated the height of FileDropZone this led to a not too nice UI display. I updated to this interface:

image


marc-aurele-besner (approved):
LGTM

PR #95, opened 2024-11-15
feat: add archiving data to nodes

clostao:
* Adds DSN information to publishing nodes
* Integrates with object mapping listener to get that info


marc-aurele-besner (approved):
LGTM

PR #97, opened 2024-11-17
File download cache

clostao:
- Adds a file download cache for both folders and files. 
- Fixes a bug in frontend file download (missing await) that led to not show password modal input if it's not the default

Uses a similar mechanism to the cache for saving files on the client though this are encrypted and are compressed.


marc-aurele-besner (dismissed):
LGTM

marc-aurele-besner (approved):
LGTM

PR #98, opened 2024-11-18
Add token refresh for google oauth

clostao:
Next-Auth doesn't manage token refresh so I'm handling manually. I did implement only Google since Discord lasts for 7 days that should be long enough before asking for re-auth


marc-aurele-besner (approved):
LGTM

PR #103, opened 2024-11-18
Cold pg-based download cache

clostao:
Similar to the memory cache at #97, files download are to be cached but instead of in memory & for a small size amount the this cache will save in the order of 100GB files (could be increased) for enabling the CDN network in the future.

This is a previous step for the integration with object fetcher to DSN when both caches (hot/cold) miss


marc-aurele-besner (approved):
LGTM

PR #107, opened 2024-11-19
Setup tests on backend

clostao:
Time to test most of the functionality: 

Added e2e testing (understood as based in user journeys rather than I/O function testing) at use case level. 

I used TestContainers for setting up a Postgres database


backend/src/useCases/uploads/uploads.ts
```diff
@@ -217,7 +217,7 @@ const createSubFolderUpload = async (
   const upload = await uploadsRepository.createUploadEntry(
     v4(),
     UploadType.FOLDER,
-    UploadStatus.PENDING,
+    UploadStatus.MIGRATING,
```
marc-aurele-besner, line 220: Not sure I understand what upload status `MIGRATING` mean?

clostao: Yeah, this is a tricky one. Basically, there are two different tables that saves IPLD nodes info. 

The first tracks the nodes that corresponds to each upload relating the uploadId and the nodes this is `uploads.blockstore`. This is used a cache-like table for enabling the processing of large files and its purpose ends at the upload completion.

The second relates the nodes with their `root_cid` the top parent in the hierarchy of an upload (e.g the root folder of a folder upload) and the `head_cid` the immediate parent (e.g if the node is file chunk it'd be the `head_cid` of the file). This table tracks the on-chain publishing and archiving data of a node.

-- 

Having this is mind when a file is uploaded (the moment a file is correctly transformed into IPLD format) the second table has to be populated with the nodes of the first one. This process is tag as `MIGRATING` since they're going from a table to another.

PR #111, opened 2024-11-27
Update onboarding page terms & conditions

clostao:
Simplifies the onboarding view from multiple to single step. More details in this [issue](https://github.com/orgs/autonomys/projects/32?pane=issue&itemId=87687717&issue=autonomys%7Cauto-drive%7C102)


marc-aurele-besner (commented):
I think we should have a single checkbox that regroup our ToS and Data Policy and other and the button that disable when the checkbox is checked, similar to the farming page or most flow.

marc-aurele-besner (approved):

PR #114, opened 2024-12-03
Improve loading table

clostao:
Between pagination updates the table was not showing any loading spinner and just the previous page. I added a loading spinner and some little improvements.

### DIFF
--- a/frontend/src/components/Files/ObjectDeleteModal.tsx
+++ b/frontend/src/components/Files/ObjectDeleteModal.tsx
@@ -35,7 +35,6 @@ export const ObjectDeleteModal = ({
     onError: (error) => {
       console.error('error', error);
     },
-    client: gqlClient,
   });
 
   const deleteObject = useCallback(() => {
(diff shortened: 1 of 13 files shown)


frontend/src/components/Files/ObjectDeleteModal.tsx
```diff
@@ -23,24 +23,19 @@ export const ObjectDeleteModal = ({
...
     },
     onError: (error) => {
       console.error('error', error);
     },
-    context: {
-      headers: {
-        Authorization: `Bearer ${session.data?.accessToken}`,
-      },
-    },
+    client: gqlClient,
```
marc-aurele-besner, line 38: One other approach will be to create a ApolloProvider and use it in layout to wrap your children's in the provider, so you won't need to pass in the client on each query

marc-aurele-besner (commented):
I think wrapping your gqlClient in a provider will simplify all these hooks 
smt like https://github.com/autonomys/astral/blob/36cf8b444af90288d9073d311a80c94c0724fdbf/explorer/src/providers/IndexersProvider.tsx#L105

frontend/src/components/Files/ObjectDeleteModal.tsx
clostao: Yeah, that's the best in this, I already had the ApolloProvider setup.

marc-aurele-besner (dismissed):
LGTM

frontend/src/components/Files/ObjectDeleteModal.tsx
clostao: Thanks for the comment btw!

marc-aurele-besner (approved):

PR #115, opened 2024-12-10
update: examples matching auto-drive cjs format

clostao:
Updates the auto-drive examples matching the update to `commonjs` format


marc-aurele-besner (approved):
Awesome

marc-aurele-besner (approved):

PR #125, opened 2024-12-21
Fix missing chunk during migration

clostao:
When a file is uploaded the IPLD nodes are generated and saved in a temporary table (`uploads.blockstore`) when the upload is completed by the user these nodes are migrated to the persistent table (`public.nodes`). 

If a user tries to fetch the file when this migration is not completed the fetch failed. This should fix that by searching in the temp table if the persistent table fails.


jfrank-summit (approved):
Quick turnaround 🚀

PR #126, opened 2024-12-23
Update to promise-based auto-drive pkg

clostao:
This PR is pending to the package to be published


examples/.env.test
```diff
@@ -1,2 +1,2 @@
-AUTO_DRIVE_API_KEY=
+AUTO_DRIVE_API_KEY=97cc977a535f4d89a3369ad5404f8ba0
```
marc-aurele-besner, line 1: This seems like a leak key, if it's a valid key, we should revoke it

clostao: Thanks for the catch, was a local API_KEY so no sec problem.

marc-aurele-besner (dismissed):
LGTM

marc-aurele-besner (approved):

Issue #121, opened 2024-12-19
[Front] Add a UI feature to request more storage uploads / download limits


EmilFattakhov:
This could be as simple as a Google form. What we want to collect is: the person/company contact details, the project they're working on (link and description), and what limits they require.

EmilFattakhov:
Resolved by https://github.com/autonomys/auto-drive/pull/127

PR #127, opened 2024-12-24
Add claim credits form

clostao:
Add a form link for claiming a credit limit increase. Solves [this issue](https://github.com/autonomys/auto-drive/issues/121)


marc-aurele-besner (approved):
Awesome

PR #131, opened 2025-01-09
Make sure the color of the FAQ question are dark

marc-aurele-besner:
## Make sure the color of the FAQ question is dark

I actually thought there were no questions/answers, but then I realized that it was just because the question text was white, like the background of the box.

This PR makes sure the color of the text is dark.

Some classes refer to dark: and use specific brand colors not defined in tailwind.config...
(This PR does not address these issues, since the design does not have a dark/light switch anyway)


clostao (approved):
Thanks!

PR #134, opened 2025-01-13
Fix search name

clostao:
Search by name is not retrieving objects correctly by name due to bad processing of the path param. `encodeUriComponent` solves this.


marc-aurele-besner (approved):
lgtm

PR #140, opened 2025-01-21
Separate auth system and backend

clostao:
Auto-Drive taurus & main-net instances will have a common user system. This PR separates the authentication system into the auth folder and updates backend and frontend for adapting to this change. 

Be aware of:

- Frontend interacts with auth system through HTTP API not GQL yet
- Users, Organizations and Api Keys are treated in auth system, subscriptions are managed in backend service


marc-aurele-besner (approved):
Awesome work 👍

PR #143, opened 2025-01-22
Update multichain frontend

clostao:
- Adds a dropdown for switching between networks
- On network change creates an API based on the links setup on `constants/networks.ts`
- Updates how these API/services are constructed


frontend/src/app/[chain]/page.tsx
```diff
@@ -0,0 +1,5 @@
+import { redirect } from 'next/navigation';
+
+export default function Page() {
+  redirect('/mainnet/drive');
```
Comment (marc-aurele-besner) [inline frontend/src/app/[chain]/page.tsx:4]: Should this redirect to /${chain}/drive instead?

marc-aurele-besner (dismissed):
Good job 👍

frontend/src/app/[chain]/page.tsx
Comment (clostao) [reply frontend/src/app/[chain]/page.tsx:4]: Not quite, but it's not correct should be sth like: `redirect('/${defaultNetwork}/drive')`

Comment (clostao) [reply frontend/src/app/[chain]/page.tsx:4]: Yup, thanks for the catch!

marc-aurele-besner (dismissed):
Look good, awesome work!

marc-aurele-besner (approved):

PR #149, opened 2025-01-27
Update persistent file cache

clostao:
Database-based cache is not as optimal as a file-based cache. This PR removes the database-based in exchange for the one based in files.

### DIFF
--- a/backend/src/config.ts
+++ b/backend/src/config.ts
@@ -1,9 +1,8 @@
 import { env } from './utils/misc.js'
 
-const DEFAULT_CHUNK_SIZE = 10 * 1024 ** 2
 const DEFAULT_MAX_CACHE_SIZE = BigInt(10 * 1024 ** 3)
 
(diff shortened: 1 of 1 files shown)


backend/src/config.ts
```diff
@@ -3,6 +3,9 @@ import { env } from './utils/misc.js'
 const DEFAULT_CHUNK_SIZE = 10 * 1024 ** 2
 const DEFAULT_MAX_CACHE_SIZE = BigInt(10 * 1024 ** 3)
 
+const DEFAULT_CACHE_MAX_SIZE = 100 * 1024 ** 2 // 100mb
```
marc-aurele-besner, line 6: Is it me or 100mb seems rather small? 🤔

clostao: We increase it, sure

clostao: I've increased it to 10gb

marc-aurele-besner (approved):
Awesome work! 🙇

PR #155, opened 2025-01-29
fix: empty susbcription info

clostao:
With multi-chain update I forgot to implement subscription retrieval


marc-aurele-besner (approved):
Awesome

PR #156, opened 2025-01-29
update: use gateway raw mode

clostao:
When http receives a `Content-Encoding: deflate` it automatically decompresses. Since the SDK manages decompression some issues were raised due to intended double decompression. 

Enabling a raw download makes the SDK the unique responsible for decompression.


marc-aurele-besner (approved):
Awesome

PR #159, opened 2025-01-30
update: improve docker setup

clostao:
Separates docker compose in three:

- production: not public
- dev using prod auth: not public 
- dev using local auth: Public env file

Additionally, it enables an environmental variable (`JWT_SECRET_ALGORITHM`) for choosing the jwt secret encryption schema. Enabling having symmetric schemas for dev environments and asymmetric for production.


marc-aurele-besner (approved):
LGTM

PR #176, opened 2025-02-03
update: removed db listeners in substitute of rabbit queues

clostao:
For enabling horizontal scaling of the backend we have to remove db listeners to ensure one particular task is not executed as many times as servers are deployed.


marc-aurele-besner (approved):
LGTM

clostao:
I've discovered a buggy behaviour in the transaction manager that affects this PR. I'm working on a fix for this issue. The problem is nonce is calculated wrongly when two batches are sent with low time difference.

clostao:
I've come up with a simple solution though I'll refactor this component (transaction manager) within this week.

marc-aurele-besner (approved):

PR #177, opened 2025-02-04
Enable public downloads

clostao:
This PR enables to allow anonymous downloads if so is determined by envar. This would allow to deploy one instance with public downloads.


marc-aurele-besner (approved):
LGTM

PR #179, opened 2025-02-05
Small updates

clostao:
Some small updates:

- Unwrap async iterable consumer in `forkAsyncIterable` method
- Add `OPTIONAL_AUTH` environment variable
- Cache: 
     -  Custom serialisation for BigInt
     -  Fix inconsistent state due to parallel caches set execution.
-   Extract download response headers to common function
- Re-structure config 
- Add limit to anonymous downloads

### DIFF
--- a/backend/src/api.ts
+++ b/backend/src/api.ts
@@ -14,19 +14,19 @@ const createServer = async () => {
 
   app.use(
     express.json({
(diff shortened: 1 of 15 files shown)


marc-aurele-besner (approved):
LGTM

Issue #178, opened 2025-02-05
Refactor transaction manager


clostao:
The current implementation (the one introduced at #176) is improvable:

* Only one batch could be sent simultaneously
* Improve robustness checking prior the submit of transactions posible errors such as account with low balance

PR #182, opened 2025-02-06
Refactor transaction manager

clostao:
As described in #178 I wanted to refactor the transaction manager because was unnecessarily complex and I implemented the following improvements:

- Simplify nonce management with `p-limit` library
- Remove accounts when a transaction fails
- Make configurable the limit of concurrent transactions
- Improved error handling (if a node fails to publish the 'publish-nodes' message is re-sent to the queue with a 10s delay).


marc-aurele-besner (approved):
LGTM

PR #183, opened 2025-02-07
Update production setup

clostao:
The instance of RabbitMQ in testnet will be in docker-compose whereas in mainnet will be on a standalone server. This change is needed for enabling those use cases.


marc-aurele-besner (approved):
LGTM

PR #187, opened 2025-02-07
feat: add web3 login

marc-aurele-besner:
## Example of SIWE custom integration with NextAuth

image
image
image
image

This is just a crude example of a sign-in with ethereum (SIWE) flow, it's crude, the final handshake with the auth endpoint is not implemented and it's missing some better edge case detection on the client side.


clostao:
The issue on the workflow was an underlying error in `testcontainers`

PR #189, opened 2025-02-10
Remove transaction results table

clostao:
I'm trying to reduce the complexity of the data model because has been getting more complex than needed in some aspects. I'll continue working on this direction. 

This first step is to remove the table `transaction_results` and replacing this information with two new columns in `nodes` table. Frontend & Hasura updated too.

PS: I updated the queries to .graphql files. Thanks for your idea @marc-aurele-besner


marc-aurele-besner (dismissed):
Awesome 👍

marc-aurele-besner (approved):

PR #191, opened 2025-02-11
Update transaction manager

clostao:
Make nonce read on-chain when accounts are re-init. This would make the transaction manager more robust to errors within submitted transactions.  

I introduced the `uniqueExecution` framework here for improving the readability of this service since managing correctly lock-ish mechanism with Promises tend to have poor readability.


marc-aurele-besner (approved):

clostao:
I've updated the branch base since I need to deploy the `remove-trxs-results` in testnet because merging into main would provoke an undesired deployment of the frontend

PR #192, opened 2025-02-11
feat: add get summary endpoint

clostao:
I added this endpoint because would be hepful for implementing some features in auto-drive sdk


marc-aurele-besner (approved):
LGTM

PR #193, opened 2025-02-12
Add developers section

clostao:
From different inputs I received feedback explaining that was difficult to find where to create ApiKeys in auto-drive frontend. For that purpose I separated Profile section into: 

- Profile: Logout, PublicId & Default Password Updater
- Developers: For the moment, ApiKey management

@EmilFattakhov: Is there any place where we should update how to create an ApiKey besides https://github.com/autonomys/auto-sdk/blob/main/packages/auto-drive/README.md?


EmilFattakhov:
@clostao I think there's only one place beside it - on the dev hub https://develop.autonomys.xyz/sdk/auto-drive#creating-an-api-key. I will get the it updated today, alongside adding more information on the functions you introduced

PR #194, opened 2025-02-12
Use new version of auto-drive

clostao:
New package used in auto-drive (yet not published)


examples/.env.test
```diff
@@ -1,2 +1,2 @@
-AUTO_DRIVE_API_KEY=
-AUTO_DRIVE_URL=http://localhost:3000
\ No newline at end of file
+AUTO_DRIVE_API_KEY=12a23f19a5344716aa97ffe8896a4563
```
marc-aurele-besner, line 1: is this a valid key, or an example?

marc-aurele-besner (approved):
LGTM

PR #196, opened 2025-02-13
Update faqs

clostao:
From the questions from users I updated the FAQs

### DIFF
--- a/frontend/src/constants/faqs.tsx
+++ b/frontend/src/constants/faqs.tsx
@@ -13,10 +13,6 @@ export const faqs = [
       'Yes, it is free but uploads and downloads have a monthly limit in bytes.',
   },
   {
-    question: 'Is all data stored on-chain?',
-    answer: 'Yes, all data is stored on-chain.',
-  },
-  {
     question: 'Are all files public?',
     answer:
       'Autonomys is a decentralized storage network, so all files are public by default. However, you can encrypt your files before uploading them to ensure they are private.',
--- a/frontend/src/constants/faqs.tsx
+++ b/frontend/src/constants/faqs.tsx
@@ -21,14 +21,14 @@ export const faqs = [
     question: 'How data availability is ensured?',
     answer: (
       <span>
-        Your file is decomposed into chunks and each chunk is processed by an
-        erasure code. This means that if a chunk is lost, the file can be
-        reconstructed from the remaining chunks. Additionally, every chunk is
-        stored in multiple nodes to ensure data availability.
+        Once your file is archived is decomposed into chunks and each chunk is
+        processed by an erasure code. This means that if a chunk is lost, the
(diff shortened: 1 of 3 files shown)


frontend/src/constants/faqs.tsx
```diff
@@ -43,4 +74,37 @@ export const faqs = [
       </span>
     ),
   },
+  {
+    question: 'When should I use Auto-Drive mainnet or taurus network?',
+    answer: (
+      <span>
+        Both networks provides the same interface through our SDK so you can
+        switch between them but mainnet storage has true permanent storage and a
+        much higher replication factor, taurus storage will be eventually shut
```
marc-aurele-besner, line 83: I'll suggest replacing `taurus storage will be eventually shut down, obviously with some reasonable notice.`
by something along `taurus is a testnet, therefore it will eventually be deprecated and replace by a newer testnet.`

I'll avoid making affirmation about notice period and such, I believe that even if we deprecate Taurus and launch a new Testnet, their will likely be taurus nodes around for a reasonable amount of times, but the tools build on top (like Auto-Drive and their infra) may very quickly deprecate it, in favour of the newer testnet.

frontend/src/constants/faqs.tsx
```diff
@@ -43,4 +74,37 @@ export const faqs = [
...
+        <br />
+        <br />
+        If your application needs permanent storage, we recommend using mainnet
+        storage else you can use taurus considering that will be a cheaper
+        option.
+      </span>
+    ),
+  },
+  {
+    question: 'Could I participate pledging storage?',
```
marc-aurele-besner, line 94:
```suggestion
    question: 'Could I participate, in pledging storage to the network?',
```

frontend/src/constants/faqs.tsx
```diff
@@ -12,15 +12,46 @@ export const faqs = [
...
       'Autonomys is a decentralized storage network, so all files are public by default. However, you can encrypt your files before uploading them to ensure they are private.',
   },
+  {
+    question: 'How data availability is ensured?',
+    answer: (
+      <span>
+        Once your file is archived is decomposed into chunks and each chunk is
+        processed by an erasure code. This means that if a chunk is lost, the
+        file can be reconstructed from the remaining chunks. Additionally, every
+        chunk is stored in multiple nodes to ensure data availability.
```
marc-aurele-besner, line 27: I'll suggest modifying this block, by:

```text
Once your file is archived, it's decomposed into chunks, and each chunk is erasure encoded. This means that if a piece of data is lost, the file can be reconstructed and its integrity can be verified. Additionally, every chunk is stored in multiple nodes to ensure data availability.
```

marc-aurele-besner (commented):
Here are a few suggestion 👍

PR #199, opened 2025-02-14
update: add default sorting

clostao:
Files were showing with an arbitrary sorting, this gave bad UX. Now it sorted descending in upload date (latest uploads first).

Additionally, I removed owner column and substituted by upload date. The prior approach for getting root_metadata couldn't be filtered so I had to change the approach using other Hasura features.

### DIFF
--- a/frontend/src/utils/time.ts
+++ b/frontend/src/utils/time.ts
@@ -1,3 +1,17 @@
 export const currentYear = () => {
   return new Date().getFullYear();
 };
+
+export const formatDate = (date: string) => {
+  const localLocale =
+    typeof navigator !== 'undefined' && navigator.language
+      ? navigator.language
+      : 'en-US';
+  return new Date(date).toLocaleDateString(localLocale, {
+    year: 'numeric',
+    month: 'short',
+    day: 'numeric',
+    hour: 'numeric',
+    minute: 'numeric',
+  });
+};
--- a/frontend/src/utils/time.ts
+++ b/frontend/src/utils/time.ts
@@ -1,3 +1,10 @@
+import dayjs from 'dayjs';
+import relativeTime from 'dayjs/plugin/relativeTime';
+import utc from 'dayjs/plugin/utc';
+
+dayjs.extend(relativeTime);
+dayjs.extend(utc);
+
 export const currentYear = () => {
   return new Date().getFullYear();
 };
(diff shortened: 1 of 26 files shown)


frontend/src/utils/time.ts
```diff
@@ -1,3 +1,17 @@
 export const currentYear = () => {
   return new Date().getFullYear();
 };
+
+export const formatDate = (date: string) => {
```
marc-aurele-besner, line 5: I'll recommend using datejs instead of Date(), you can take inspiration of utils/time.ts in Astral

frontend/src/views/GlobalFiles/query.graphql
```diff
@@ -1,19 +1,26 @@
...
+query GetGlobalFiles(
+  $limit: Int!
+  $offset: Int!
+  $orderBy: [metadata_roots_order_by!]
+) {
+  metadata_roots(
     where: {
-      root_metadata: { object_ownership: { _and: { is_admin: { _eq: true } } } }
+      inner_metadata: {
+        object_ownership: { _and: { is_admin: { _eq: true } } }
```
marc-aurele-besner, line 9:
```suggestion
        object_ownership: { is_admin: { _eq: true } }
```
This query can be simplify

frontend/src/views/GlobalFiles/query.graphql
```diff
@@ -54,10 +61,11 @@ query GetGlobalFiles($limit: Int!, $offset: Int!) {
       }
     }
   }
-  metadata_aggregate(
-    distinct_on: root_cid
+  metadata_roots_aggregate(
     where: {
-      root_metadata: { object_ownership: { _and: { is_admin: { _eq: true } } } }
+      inner_metadata: {
+        object_ownership: { _and: { is_admin: { _eq: true } } }
```
marc-aurele-besner, line 67:
```suggestion
        object_ownership: { is_admin: { _eq: true } }
```

marc-aurele-besner (approved):
Some of these queries could be simplified since they only have one condition, therefore not making usage of the _and, else than that, LGTM

frontend/src/views/GlobalFiles/query.graphql
clostao: Thanks for the catch!

frontend/src/utils/time.ts
clostao: Thanks for the suggestion, looks better having a relative reference`

marc-aurele-besner (approved):
LGTM

Issue #66, opened 2024-10-23
[Front] Reload management


clostao:
Currently when a user makes a write operation (e.g upload a file) for reloading data from the application a `window.location.reload()` is used. This should change to some alternative such as `router.refresh()`.

iamnamananand996:
Hi @clostao,

This is not that simple 🤔, and I agree with the fact that, `after each action reloading the whole page is a Bad UX`.

`window.location.reload()` --> `router.refresh()`

Above could have worked if we where use **server-side data fetching** e.g. `fetch` inside a Server Component or `getServerSideProps`.

But In our case, it we using `client-side data fetching` like `useQuery` or fetch inside `useEffect / useCallback`, so `router.refresh()` does not automatically trigger a re-fetch unless the fetching library supports it.


I am listing some way, which can achieve this and make it better.

### Approach 1 (not UX friendly) because we will need to add a refresh button 

```
const { data, loading, refetch } = useGetMyFilesQuery({
  variables: {
    oauthUserId: user?.oauthUserId ?? '',
    oauthProvider: user?.oauthProvider ?? '',
    limit: pageSize,
    offset: currentPage * pageSize,
  },
  fetchPolicy: 'no-cache', // Ensures fresh data every time
});

// Call refetch() after a file upload if possible or else add refresh button 
const handleFileUpload = async () => {
  await uploadFile(); // Your file upload logic
  refetch(); // Re-fetch the GraphQL query instead of full page reload
};
```


### Approach 2

```
import { useApolloClient } from "@apollo/client";

const client = useApolloClient();

const handleFileUpload = async () => {
  await uploadFile();
  client.refetchQueries({ include: ["GetMyFiles"] }); // Replace "GetMyFiles" with the actual query name
};
```


### Approach 3 (I would go for, will help in long run)

Whenever, we have this kind of use case, where component are dependent to each other, we should consider having **global store, architecture**, that help to reduce the complexity and maintain the data Consistency.

So we should Maintain a global store, as we are using `zustand` so `zustand global store` for the page, and do API call under the store, and share data from their to each component.

One time changes, but help to extend thing further, easily like
- Store level caching.
- optimistic updates
- No need to call the API again, when deleting something, just remove entry from the store
- Improved UX

etc.


Let me know, if you like any of the solution, I would love to do the implementation.

cc - @jfrank-summit, @marc-aurele-besner 


Meaning while, can I get the deployed url version for this ENV key.

[...]

PR #202, opened 2025-02-18
Refactor files state

clostao:
As stated in #66, I created a global state for managing the fetching and storage of files information for the four views:  Global, User, Shared & Trashed.

With this change reload management becomes an easy task since it's designed for it.


marc-aurele-besner (dismissed):
I'm fine with this, that been said, I do think most of these fetchPolicy could be switch to `cache-and-network` this way the data from cache is loaded first, then hydrated from network.

I also think some of these queries could benefit from a poolinfInterval with a skip when the component is not in view.

clostao:
Thanks for your comments. 

The `cache-and-network` policy is nice though for re-fetching is not the best option since the cache would have outdated values being those retrieved first. To sum up, the `no-cache` policy should remain for one-of refetching though having a  hook `pollInterval` and `cache-and-network` makes sense.

marc-aurele-besner (approved):
LGTM

PR #205, opened 2025-02-19
Add sorting to files table

clostao:
Added sorting to files table by name or date both descending or ascending


marc-aurele-besner (approved):
Nice

PR #207, opened 2025-02-19
fix: folder upload

clostao:
Folder uploads consists of several linked uploads where node publishing doesn't occurr until the whole upload is finished. 

After the update of the message-based uploads within a folder were being migrated, this shouldn't occur until the whole upload is uploaded and parsed into IPLD nodes. For solving this, we should just ensure the scheduling of node migration is only emitted once the root upload is finished and processed.


marc-aurele-besner (approved):
LGTM

PR #209, opened 2025-02-20
update: search placeholder

clostao:
Search bar in "My files" search within your files but the placeholder didn't reflect this


marc-aurele-besner (approved):
LGTM

PR #212, opened 2025-02-24
fix: session lasting hour

clostao:
@jfrank-summit reported that sessions in auto-drive were lasting an hour only though the refreshing token was setup  to 7-day-long. 

Refreshing access token was not managed correctly now every time a session called is made by the user a refreshing in intended (in the case access token is to expired)


marc-aurele-besner (approved):
LGTM

PR #227, opened 2025-03-05
fix: request tracing

clostao:
The usage of middleware was provoking duplicated headers sent. With this setup we avoid that behavior.

Also, I unified the monitoring object with `auto-files-gateway`


marc-aurele-besner (approved):

clostao:
Yup, thanks!

PR #264, opened 2025-03-20
Fix `net::ERR_HTTP2_PROTOCOL_ERROR` on google chrome

clostao:
The issue was that the `Content-Length` mismatched the actual size (due to decompression) of the returned object and chrome detected this mismatch as a protocol error.


marc-aurele-besner (approved):
LGTM

DaMandal0rian (approved):

PR #266, opened 2025-03-25
add AD data access layer nginx configs

DaMandal0rian:
- add backend configurations for each entity
- add rate-limiting to protect from abuse


clostao (commented):
🤲

PR #272, opened 2025-03-26
feat(ui): ✨ branded landing page for `gateway.autonomys.xyz`

iamnamananand996:
iScreen Shoter - Google Chrome - 250327020400

### DIFF
--- a/gateway/public/index.html
+++ b/gateway/public/index.html
@@ -130,7 +130,7 @@
                 viewBox="0 0 24 24"
                 fill="none"
                 xmlns="http://www.w3.org/2000/svg"
-                class="mr-1"
+                class="mr-2"
               >
                 <g id="SVGRepo_bgCarrier" stroke-width="0"></g>
                 <g
--- a/gateway/public/index.html
+++ b/gateway/public/index.html
@@ -7,7 +7,7 @@
       name="description"
       content="A unified gateway solution designed to seamlessly connect and access files or folders permanently stored on Autonomys Mainnet&Tesnet through a single, streamlined interface."
     />
-    <title>Auto Drive Getaway | Autonomys</title>
+    <title>Auto Drive Gateway | Autonomys</title>
     <link
       rel="icon"
       href="https://ai3.storage/favicon.ico"
@@ -50,8 +50,9 @@
           Auto Drive Getaway
         </h2>
         <p class="text-lg md:text-xl text-muted-foreground max-w-3xl mx-auto">
-          A unified gateway solution designed to seamlessly connect and manage
-          multiple networks through a single, streamlined interface.
+          A unified gateway solution designed to seamlessly connect and access
+          files or folders permanently stored on Autonomys Mainnet & Tesnet
+          through a single, streamlined interface.
         </p>
       </section>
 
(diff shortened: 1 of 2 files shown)


marc-aurele-besner:
It looks great! 💪

iamnamananand996:
> I don't believe it should always behave as a download, I believe some files get view in the browser (like JSON or most images).
> 
> I'm not against a download button, but if we add a download button, we should also just have a view button that redirect to the route for browser preview.
> 
> For example, I usually prefer not downloading a file before I have preview it. I don't want to end-up downloading a super large file or a executable, by mistake.
> 
> Maybe @clostao can pitch-in

Hi @marc-aurele-besner , I agree with you button `label` and `icon` look misleading.

The actual functionality is on click of the button `redirect to the route for browser preview`, I am changing it `view` with `eye` icon.

gateway/public/index.html
```diff
@@ -0,0 +1,318 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta
+      name="description"
+      content="A unified gateway solution designed to seamlessly connect and manage multiple networks through a single, streamlined interface."
```
marc-aurele-besner, line 8: I think the different description should be a bit more specific to Autonomys, `manage multiple networks` is a bit vague. We don't really use this endpoint to `manage networks` but rather to access file regardless if they are stored on mainnet or testnet.
```suggestion
      content="A unified gateway solution designed to seamlessly connect and access files permanently stored on Autonomys Mainnet&Tesnet through a single, streamlined interface."
```
Please also change it in other location

marc-aurele-besner (changes requested):
I just find some of the description, not very clear on what this web app is use for.

marc-aurele-besner:
I think we should also take the opportunity to add some sort of footer with other links, like link to the main Autonomys website, explorer.autonomys, Auto-Drive, the academy the developer hub...

clostao:
> Maybe @clostao can pitch-in

Yeah, I agree with Marc's feedback I wouldn't name the button "download". 

Aswell, would be nice to add the footer with some links

gateway/public/index.html
```diff
@@ -0,0 +1,318 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta
+      name="description"
+      content="A unified gateway solution designed to seamlessly connect and access files or folders permanently stored on Autonomys Mainnet&Tesnet through a single, streamlined interface."
+    />
+    <title>Auto Drive Getaway | Autonomys</title>
```
clostao, line 10: It's mispelled Gateway

clostao (changes requested):
Besides that, looks amazing!

iamnamananand996:
Hi @marc-aurele-besner, @clostao 
updated footer 

iScreen Shoter - Google Chrome - 250327033344


I am not sure how quickly we need to extend it more to have more pages and features. If that case comes, we will migrate it to a Next.js application.

clostao:
>  If that case comes, we will migrate it to a Next.js application.

Sure, I think that a static page is what makes the most sense right now but if it gets more complex we could consider other options such as NextJS.

clostao (commented):
🚀

clostao (approved):

marc-aurele-besner (approved):
Looking great, I think this is all we need! Thank you

PR #280, opened 2025-04-03
chore: update @autonomys/file-caching

clostao:
Updates `@autonomys/file-caching` for incorporating [this change](https://github.com/autonomys/auto-sdk/pull/352)


jfrank-summit (approved):

marc-aurele-besner (approved):
I'll really like that we keep all the @autonomys packages to the same version when we update one

PR #308, opened 2025-04-10
fix: refresh session on landing page

clostao:
The middleware uses `getToken` next-auth method for retrieving the session information of the request user. The root cause of the issue is that `getToken` method doesn't refresh the session if needed making the call to `checkAuth` to fail.

For ensuring no authentication is asked before refresh token expiration a session refresh is forced on the landing page.


clostao:
@jfrank-summit This seems to be the problem for quick token expiration

marc-aurele-besner (approved):

jfrank-summit (approved):
LGTM

PR #322, opened 2025-04-16
Update sidebar

### LINKED ISSUE #319: Update sidebar
- Add AccountID
- Update the credit renewal to relative time reference

clostao:
Closes #319


frontend/src/components/FileTables/common/CopiableText.tsx
```diff
@@ -0,0 +1,46 @@
+import { Check, Copy } from 'lucide-react';
```
iamnamananand996, line 1: Thanks for this refactor, really helpfull 💪

iamnamananand996 (approved):

clostao:
This is how it would look the sidebar w/ the last update:

image

Not fully convinced on how the Public ID is shown

iamnamananand996:
For `public ID`, ideally it should be something like this.
iScreen Shoter - Google Chrome - 250417040001

cc - @jfrank-summit, @marc-aurele-besner

frontend/src/components/AccountInformation/index.tsx
```diff
@@ -77,11 +92,8 @@ export const RemainingCreditTracker = ({
         </div>
         <div className='mt-2 flex flex-col text-center'>
           <span className='text-sm text-gray-400 dark:text-darkBlack'>
-            Credit for period
+            Credit renews on {utcToLocalRelativeTime(renewalDate.toISOString())}
```
jfrank-summit, line 95:
```suggestion
            Credit renews in {utcToLocalRelativeTime(renewalDate.toISOString())}
```

jfrank-summit (commented):
Regarding location of Public ID, moving it where @iamnamananand996 suggested makes sense, but this means more changes than I think is necessary for this PR. If we want to redesign how profile is accessed we should do that in a separate PR. Also, @amiknyc may want to weigh in.

clostao:
I removed the update for Public ID & update the text for fitting in just one line:

image

iamnamananand996 (approved):

PR #331, opened 2025-05-01
update: ignore node publication if already published

clostao:


clostao:
@marc-aurele-besner have in mind that this solution doesn't provide the user a way to force the upload, I don't see a clear use case for that feature

marc-aurele-besner (approved):
LGTM

PR #334, opened 2025-05-05
fix: workflows should run on auth update

clostao:
Services workflow should run when auth is updated

Fixed some tests


iamnamananand996 (changes requested):
Hi @clostao, this PR have 7 commit, but show 0 file changes

PR #350, opened 2025-05-13
feat: add max retries to task manager

clostao:
Add a limit to the times a task is tried to be executed


marc-aurele-besner (approved):
LGTM

marc-aurele-besner (approved):

Issue #361, opened 2025-05-19
Large Files Download


clostao:
Large files downloads from the DSN are currently inconvenient since it would require long time waits.  Some other mechanism for this kind of files would be a good option to improve the performance of this downloads.

clostao:
I'll start implementing this on the backend but it would be nice to have some opinions on how this should be showed in the front-end.

iamnamananand996:
## Option 1

Not sure, either it is too much to ask. I would like to have bellow on the `getway`.


Purpose | URL Example
-- | --
Preview | https://gateway.autonomys.xyz/file/CID
Download | https://gateway.autonomys.xyz/file/CID?download=true


and let the downloading be handled by the native browser, the same way we would when downloading any small or big file from centralised storage.

It may require some of these headers to be added to our getaway server

```javascript
res.setHeader('Content-Type', 'application/octet-stream');


// this trigger the native browser downloading
res.setHeader('Content-Disposition', 'attachment; filename="example.zip"');


res.setHeader('Content-Length', fileSize);
stream.pipe(res);
```

Then on the frontend, we will need to just add 

```html
<a href="https://gateway.autonomys.xyz/file/bafk...?download=true" download>Download File</a>
```

## Option 2

We have already discussed on `Bi weekly` https://github.com/autonomys/auto-drive/issues/360#issuecomment-2892192779 


cc - @clostao, @jfrank-summit , @marc-aurele-besner

clostao:
Option #1 is completely feasible

PR #364, opened 2025-05-21
[Backend] Add async downloads

clostao:
Backend implementation of #361

Created `AsyncDownload` entity for tracking the "Bring To Cache" operations I've added the namespace download:

With this three endpoints:

POST `async/:cid`: Creates a async download for file with given CID
POST `async/:cid/dismiss`: Hides an async download sth like acknowledge notifications
POST `:/cid`: Handles the download of files. Is the same controller as `/objects/:cid/download` that is now marked as deprecated

### DIFF
--- a/backend/__tests__/e2e/downloads/async.spec.ts
+++ b/backend/__tests__/e2e/downloads/async.spec.ts
@@ -174,6 +174,22 @@ describe('Async Downloads', () => {
     )
   })
 
+  it('should get all downloads for a user', async () => {
+    // Create multiple downloads for the same user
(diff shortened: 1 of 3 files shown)


iamnamananand996 (approved):
Hi @clostao, if we can also add the uses example for the PR, will help us understand what's happening here.

marc-aurele-besner (dismissed):

iamnamananand996 (approved):

marc-aurele-besner (dismissed):

clostao:
I've added a few endpoints more for the SDK

iamnamananand996 (approved):

marc-aurele-besner (approved):

PR #372, opened 2025-05-30
fix: misusage of Readable stream

clostao:
Problem: 

`Stream.read()` is meant for ask for more chunks not for retrieving the whole file

Solution:

Implement a reading mechanism using a "closure" variable in the "retrieveAndReassembleFile" method.

### DIFF
--- a/backend/src/useCases/objects/files.ts
+++ b/backend/src/useCases/objects/files.ts
@@ -149,25 +149,33 @@ const retrieveAndReassembleFile = async (
     return Readable.from(chunkData)
   }
 
-  const CHUNK_SIZE = 100
+  const SIMULTANEOUS_CHUNKS = 100
+  let currentIndex = 0
   return new Readable({
-    read: async function () {
-      for (let i = 0; i < metadata.chunks.length; i += CHUNK_SIZE) {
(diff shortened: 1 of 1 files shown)


marc-aurele-besner (approved):

backend/src/useCases/objects/files.ts
```diff
@@ -149,25 +149,33 @@ const retrieveAndReassembleFile = async (
...
         if (chunkedData.some((e) => e === undefined)) {
-          throw new Error('Chunk not found')
+          this.destroy(new Error('Chunk not found'))
+          return
         }
 
-        const canContinue = this.push(Buffer.concat(chunkedData.map((e) => e!)))
-        if (!canContinue) {
-          await new Promise((resolve) => this.once('drain', resolve))
-        }
```
jfrank-summit, line 168: Is there a risk in removing back pressure handling?

clostao: I don't think so since we're limiting the simultaneous batches (~6.4MB) but I'll implement it anyways just in case we'd want to increase or setup an environment variable for this one.

clostao: Updated

PR #375, opened 2025-06-03
Improve async downloads

clostao:
In backend: 

* Added some logs
* Update to (progress 0%) once the download is triggered

In frontend:

* Add bring to cache feature to global files too
* Create a dropdown dialog for actions to avoid overwhelming the interface with much buttons

Previous:

image

This changes to:

image


marc-aurele-besner (approved):

jfrank-summit:
I think this should be targeting `main`

iamnamananand996 (dismissed):

clostao:
Added support for chunked file downloads from the DSN

iamnamananand996 (approved):

PR #377, opened 2025-06-06
Improve admin panel

clostao:
Added pagination, showing pending credits, and search by public id


jfrank-summit:
> If you're testing with ai3.storage you won't see anything, in local you should.

I was testing with the preview link.

auth/src/controllers/user.ts
```diff
@@ -225,9 +225,27 @@ userController.get('/list', async (req, res) => {
...
+    // Validate pagination parameters
+    if (page !== undefined && (isNaN(page) || page < 1)) {
+      res.status(400).json({ error: 'Invalid page parameter' })
+
+      if (limit !== undefined && (isNaN(limit) || limit < 1 || limit > 100)) {
+        res.status(400).json({
+          error: 'Invalid limit parameter (must be between 1 and 100)',
+        })
+        return
+      }
```
jfrank-summit, line 242: Does it make sense to have this nested? Won't it send two responses if both `limit` and `page` are invalid and never validate `limit` if `page` is valid?

auth/src/services/authManager/express.ts
```diff
@@ -57,28 +58,64 @@ export const handleAuthIgnoreOnboarding = async (
...
     })
+    return false
+  }
+
+  const oauthUser = await AuthManager.getUserFromAccessToken(
+    provider,
+    accessToken,
+  ).catch((e) => {
+    logger.error(e)
     return null
```
jfrank-summit, line 92: Should this return `null` or `false`, the return type should be `boolean`.

auth/src/services/authManager/express.ts
```diff
@@ -57,28 +58,64 @@ export const handleAuthIgnoreOnboarding = async (
...
     })
-    return null
+    return false
   }
 
   const isCorrectApiSecret = accessToken === config.apiSecret
-  if (!isCorrectApiSecret) {
+  if (isCorrectApiSecret) {
+    return true
+  }
```
jfrank-summit, line 78: What is the point of `apiSecret` here?

auth/src/repositories/users.ts
```diff
@@ -97,6 +97,31 @@ const getAllUsers = async (): Promise<User[]> => {
...
+
+  // Get total count
+  const countResult = await db.query(
+    'SELECT COUNT(*) as total FROM users.users',
+  )
+  const total = parseInt(countResult.rows[0].total)
+
+  // Get paginated users
+  const users = await db.query(
+    'SELECT * FROM users.users ORDER BY public_id LIMIT $1 OFFSET $2',
```
jfrank-summit, line 115: Should we add an index for `public_id`?

jfrank-summit (commented):
A few comments.

auth/src/controllers/user.ts
clostao: No, it doesn't, I updated

auth/src/services/authManager/express.ts
clostao: Backend instances retrieve a user information by public id using a permissioned endpoint, for handling access to these endpoints I use a api secret.

auth/src/controllers/user.ts
```diff
@@ -225,9 +225,28 @@ userController.get('/list', async (req, res) => {
...
+
+      return
+    }
+
+    if (limit !== undefined && !isNaN(limit)) {
+      res.status(400).json({
+        error: 'Invalid limit parameter (must be a number)',
+      })
+      return
+    }
```
jfrank-summit, line 245: This logic seems inverted, I think it should be something like:
```suggestion
    if (page !== undefined && (isNaN(page) || page < 1)) {
      res.status(400).json({ error: "Invalid page parameter" });
      return;
    }
    
    if (limit !== undefined && (isNaN(limit) || limit < 1 || limit > 100)) {
      res.status(400).json({
        error: "Invalid limit parameter (must be between 1 and 100)",
      });
      return;
    }
```

PR #380, opened 2025-06-09
fix: port setup and health checks for workers

clostao:
Problem:

Container was port used by environment variable and it was mapped to the same variable. However, healthcheck was poiting to :3000

Solution:

All the services use :3000 in the container and are mapped to port determined by environment variable


jfrank-summit (commented):
LGTM

jfrank-summit (approved):

PR #390, opened 2025-06-16
Save events failed events in a separate queue

clostao:
Problem:

When there's an bug in auto-drive that makes events to not be processed correctly (e.g low balance in publishing accounts) these events are lost once all retries have been consumed.

Solution:

Create a separate queue for saving failing events so when the an issue is detected and sorted the events could be redirected again to the processing queue so no events are lost.

### DIFF
--- a/backend/src/drivers/rabbit.ts
+++ b/backend/src/drivers/rabbit.ts
@@ -1,8 +1,7 @@
 import { Channel, connect } from 'amqplib'
 import { config } from '../config.js'
 
-export type AutoDriveQueues = (typeof queues)[number]
-const queues = ['task-manager', 'download-manager'] as const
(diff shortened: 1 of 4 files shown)


jfrank-summit (commented):
Overall looks good. One question, how will the error queues be re-processed?

clostao:
> Overall looks good. One question, how will the error queues be re-processed?

There's a feature on rabbitmq that allows to move all messages from queue to another. So, whenever we got some issue producing errors, the error queue would be populated with failing tasks. Then, when we're sure that the issue is solved we simply move messages from the error queue to the processing queue using the rabbitmq interface.

jfrank-summit (approved):
LGTM

PR #408, opened 2025-06-25
fix: no data on table init

clostao:
**Problem:**

The first fetching was not performed correctly

**Solution:**

Fetch data when fetcher is updated


jfrank-summit (approved):
LGTM

Issue #417, opened 2025-06-28
Files Not Cached On Archival


clostao:
Files are meant to be cached on archival

```ts
const onObjectArchived = async (cid: string) => {
  await ObjectUseCases.populateCaches(cid) // here
  await metadataRepository.markAsArchived(cid)
  await nodesRepository.removeNodeDataByRootCid(cid)
}
```

Files recently archived are not being cached at the auto-drive level.

clostao:
I noticed that the file is getting fetched from the DSN for populating the cache with the file. This shouldn't happen since the point is populating the file from DB reducing the usage of bandwidth of DSN and faster retrievals.

I'm debugging this issue but I didn't have some logs that are now in production.

PR #418, opened 2025-06-29
Fix file cache population on object archived

clostao:
This would solve #417. 

**Problem:**

Populating the cache asynchronously leads to a race condition where the removal of nodes (performed just after the file is populated in cache) starts before the end of the cache population leading to the error shown. 

**Solution:**

Awaiting for the file to be populated for nodes data to be wiped.


jfrank-summit (approved):
LGTM!

PR #423, opened 2025-07-07
Fix submodules sync

clostao:
@jfrank-summit You reported some issue on the past syncing, for already initialised workspaces the `init-submodules` would correctly keep up to date the latest pointer to submodules though would fail for not initialised workspaces.


jfrank-summit (approved):
🚀

PR #425, opened 2025-07-08
feat: implement file preview

clostao:


frontend/src/components/FilePreview/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+      case AutoDriveNetworkId.MAINNET:
+        return AutoUtilsNetworkId.MAINNET;
+      case AutoDriveNetworkId.TAURUS:
+        return AutoUtilsNetworkId.TAURUS;
+      default:
+        return AutoUtilsNetworkId.MAINNET;
+    }
+  }, [network]);
+
+  const gatewayUrl = EXTERNAL_ROUTES.gatewayObjectDownload(metadata.dataCid);
```
jfrank-summit, line 37: Should we validate that this is a valid CID prior to making the request?

frontend/src/components/FilePreview/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+            `Failed to fetch file: ${response.status} ${response.statusText}`,
+          );
+        }
+        const blob = await response.blob();
+        setFile(blob);
+        // For text-based files, also read the content
+        if (needsContentParsing(metadata)) {
+          const text = await blob.text();
+          setTextContent(text);
+        }
```
jfrank-summit, line 73: Do we do any kind of sanitization to ensure that no malicious code is being rendered?

frontend/src/components/FilePreview/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+            );
+            const decryptedBlob = await processFileData(decryptedFileData);
+            setFile(decryptedBlob);
+            setIsDecrypted(true);
+            setLoading(false);
+            return;
+          } catch {
+            setDecryptionError('Invalid password or decryption failed');
+          }
+        }
```
jfrank-summit, line 96: We should clear the password from state after it is used. We should also think about how we want the lifecycle of decrypted data to be handled.

frontend/src/components/FilePreview/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+        setLoading(false);
+        return;
+      }
+
+      // Encrypted files always need to be fetched and decrypted
+      setIsFilePreview(true);
+      setLoading(true);
+
+      try {
+        const response = await fetch(gatewayUrl);
```
jfrank-summit, line 61: How are we handling very large file requests?

frontend/src/components/FilePreview/index.tsx
clostao: I don't think so, the cid passed (`metadata.dataCid`) to the method comes from the metadata structure

frontend/src/components/FilePreview/index.tsx
clostao: Password isn't saved on the state, is just a callback parameter: [see here](https://github.com/autonomys/auto-drive/pull/425/files#diff-255c318571563f19b9c163ed93b9823c0ac9a1f33dfa22dc7b43585b757e4c0bR40)

frontend/src/components/FilePreview/index.tsx
clostao: We're not handling in a particular way. We could implement this mechanism for example:

Having two limits, the lower (e.g 1MB) is the maximum that could fetched from the DSN and the higher (e.g 100MB) would be biggest pre-viewable file.

For files <1MB would be pre-viewed always. For files between first and second limit will be previewed only if they're cached, and files beyond the 100MB limit the file won't be previewed.

frontend/src/components/FilePreview/index.tsx
clostao: No, I didn't but it's a good idea. I just implemented [here](https://github.com/autonomys/auto-drive/pull/425/commits/f16697e02d6d76c08ea85a62006c723fdd1b1640)

jfrank-summit (approved):
LGTM

PR #433, opened 2025-07-16
feat: publishing tasks for nodes failing to be published into error queues

clostao:
Unlike the rest types of events nodes publishing task events are not retried on the same manner. Since the structure of the event is: 

```ts
interface NodePublishingEvent {
   id: 'publish-nodes'
   params: {
     nodes: string[]
   }
}
``` 

Sometimes the publishing of the nodes fails partially, meaning that the first N elements of the array go through but the following items timeout or fail for any other reason, for avoiding triggering a republish of all the nodes in the event of a partial failure we only republish the failed ones. 

This mechanism was not publishing into error queues making errors hardly recoverable. This PR solves that


jfrank-summit (approved):
LGTM

PR #436, opened 2025-07-17
Support byte range downloads

clostao:
## Overview

This PR implements comprehensive [HTTP range request support](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Range) for the Auto-Drive download system, enabling partial content downloads, streaming optimization, and resumable file transfers.

### Core Features
- **HTTP Range Request Parsing**: Parse `Range: bytes=start-end` headers
- **HTTP 206 Partial Content**: Return proper 206 responses with `Content-Range` headers

### Minor changes

- Created interface ObjectFetcher with two implementations: `DBObjectFetcher` and `FileGatewayObjectFetcher`.
- Unified mechanism for composing nodes
- Other refactors for simplifying the repo structure

### DIFF
--- a/backend/src/utils/http.ts
+++ b/backend/src/utils/http.ts
@@ -1,0 +1,15 @@
+import { Request } from 'express'
+import { ByteRange } from '@autonomys/file-caching'
+
+export const getByteRange = (req: Request): ByteRange | undefined => {
+  const byteRange = req.headers['content-range']
+  if (byteRange == null) {
+    return undefined
+  }
+  const header = 'bytes '
+
+  const [start, end] = byteRange.slice(header.length).split('-')
+  const startNumber = Number(start)
+  const endNumber = end && end !== '*' ? Number(end) : undefined
+  return [startNumber, endNumber]
+}
--- a/packages/models/package.json
+++ b/packages/models/package.json
@@ -14,7 +14,8 @@
     "url": "https://www.autonomys.net"
   },
   "dependencies": {
-    "@autonomys/auto-dag-data": "^1.3.1",
+    "@autonomys/auto-dag-data": "^1.5.10",
+    "@autonomys/file-caching": "^1.5.10",
     "zod": "^3.24.2"
   },
   "devDependencies": {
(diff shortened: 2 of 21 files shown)


backend/src/utils/http.ts
```diff
@@ -0,0 +1,15 @@
+import { Request } from 'express'
+import { ByteRange } from '@autonomys/file-caching'
+
+export const getByteRange = (req: Request): ByteRange | undefined => {
+  const byteRange = req.headers['content-range']
+  if (byteRange == null) {
+    return undefined
+  }
```
jfrank-summit, line 8: It would be useful to do some further validation (formatting, negative numbers, start > end, etc.).

backend/src/utils/http.ts
```diff
@@ -0,0 +1,15 @@
+import { Request } from 'express'
+import { ByteRange } from '@autonomys/file-caching'
+
+export const getByteRange = (req: Request): ByteRange | undefined => {
+  const byteRange = req.headers['content-range']
```
jfrank-summit, line 5: I think this should be looking for a [`range` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Range_requests) in the request. [`content-range`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Range) is used in the response.

packages/models/package.json
```diff
@@ -14,7 +14,8 @@
     "url": "https://www.autonomys.net"
   },
   "dependencies": {
-    "@autonomys/auto-dag-data": "^1.3.1",
+    "@autonomys/auto-dag-data": "^1.5.10",
+    "@autonomys/file-caching": "^1.5.10",
```
jfrank-summit, line 18: We are updating to `1.5.11` elsewhere in this PR.

jfrank-summit (commented):
Just a few minor comments.

jfrank-summit (approved):
LGTM

PR #445, opened 2025-07-29
Gateway improvements

clostao:
This PR implements several changes to the multi-network gateway:

- Adds a timeout to download response
- Adds handler for underlying failing requests
- Single API creation
- Enhanced version of Promise.all where the first not null element is returned not waiting for the rest of requests


gateway/src/http.ts
```diff
@@ -14,8 +14,14 @@ export const internalRedirect = async (
 
   const response = await fetch(_url, {
     headers,
+    signal: AbortSignal.timeout(30_000),
```
jfrank-summit, line 17: Is 30 seconds sufficient? Should this be configurable?

clostao: I agree should be configurable

PR #447, opened 2025-07-31
Download for some files halted w/o finishing ever

clostao:
Downloads from source were performed correctly but the duplication on streams for caching was provoking streams not being consumed therefore pausing the HTTP download.

More precisely, when a stream is forked in two branches the first needs the second (and viceversa) to be consumed too in order to be consumed. The problem was in that we were caching sequentially both `fsCache` and `memoryDownloadCache`  so the first would block the caching of the first won't happen until the second is consumed, effectively halting the download.

**Problem:**

```ts
// Awaiting in first set call was making 'fsCacheStream' to not be consumed blocking the download
forkStream(cacheStream).then(async ([fsCacheStream, memoryCacheStream]) => {
      await memoryDownloadCache.set(cid, memoryCacheStream) // awaiting
      await fsCache.set(cid, { data: fsCacheStream, size })
})
```

**Solution:**

```ts
// Fork the stream again for caching w/o blocking the main thread
forkStream(cacheStream).then(async ([fsCacheStream, memoryCacheStream]) => {
   memoryDownloadCache.set(cid, memoryCacheStream)
   fsCache.set(cid, { data: fsCacheStream, size })
})
```

### DIFF
--- a/backend/src/core/objects/files/nodeComposer.ts
+++ b/backend/src/core/objects/files/nodeComposer.ts
@@ -44,7 +44,7 @@ export const composeNodesDataAsFileReadable = async ({
         )
 
         if (chunkedData.some((e) => e === undefined)) {
-          const notFoundChunkIndex = chunksToDownload.findIndex(
+          const notFoundChunkIndex = chunkedData.findIndex(
             (e) => e === undefined,
           )
           this.destroy(
--- a/backend/src/infrastructure/services/download/index.ts
+++ b/backend/src/infrastructure/services/download/index.ts
@@ -80,24 +80,25 @@ export const downloadService = {
     logger.debug(
       `Populating caches for cid=${cid} with options=${JSON.stringify(options)}`,
     )
-    const [stream1, stream2] =
+
+    // If byte range is set, don't cache and return the stream directly
+    if (options?.byteRange) {
+      return stream instanceof Readable ? stream : Readable.from(stream)
+    }
+
+    // Fork the stream once for caching and return
+    const [returnStream, cacheStream] =
       stream instanceof Readable
         ? await forkStream(stream)
         : await forkAsyncIterable(stream)
 
-    // If byte range is not set, cache the file in the memory and file system caches
-    // Do this asynchronously to avoid blocking the stream return
-    if (!options?.byteRange) {
-      forkStream(stream2)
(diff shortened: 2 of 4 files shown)


backend/src/core/objects/files/nodeComposer.ts
```diff
@@ -44,7 +44,7 @@ export const composeNodesDataAsFileReadable = async ({
         )
 
         if (chunkedData.some((e) => e === undefined)) {
-          const notFoundChunkIndex = chunksToDownload.findIndex(
+          const notFoundChunkIndex = chunkedData.findIndex(
```
clostao, line 47: This was a check that we were performing incorrectly by checking if the requested cid was not undefined rather than the actual data of that CID

backend/src/infrastructure/services/download/index.ts
```diff
@@ -80,24 +80,25 @@ export const downloadService = {
...
-        .catch((error) => {
-          logger.warn(error, 'Error caching file with cid %s', cid)
-        })
-    }
-    return stream1
+    // Fork the stream again for caching w/o blocking the main thread
+    forkStream(cacheStream).then(async ([fsCacheStream, memoryCacheStream]) => {
+      memoryDownloadCache.set(cid, memoryCacheStream)
+      fsCache.set(cid, { data: fsCacheStream, size })
+    })
```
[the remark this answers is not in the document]
clostao: This is incorrect since `forkAsyncIterable` returns [Iterable, Iterable]

backend/src/infrastructure/services/download/index.ts
```diff
@@ -80,24 +80,25 @@ export const downloadService = {
...
-    // Do this asynchronously to avoid blocking the stream return
-    if (!options?.byteRange) {
-      forkStream(stream2)
-        .then(async ([stream3, stream4]) => {
-          await memoryDownloadCache.set(cid, stream3)
-          await fsCache.set(cid, { data: stream4, size })
-        })
-        .catch((error) => {
-          logger.warn(error, 'Error caching file with cid %s', cid)
-        })
```
jfrank-summit, line 98: Why remove the error handling?

jfrank-summit (approved):
LGTM 🚀

PR #448, opened 2025-08-06
Fix memory cache population

clostao:
Testing S3 compatibility layer in staging environment I detected this error that was failing to cache correctly files in memory cache leading to download errors.


jfrank-summit (approved):
LGTM

PR #468, opened 2025-09-03
Fix anonymous metadata query

clostao:
Problem:

Anonymous users have no access to object ownerships, this led to failures on metadata fetch by CID since it retrieved object ownership. Additionally. we want to get object ownerships for authenticated users so we can show 'delete' or 'share' button when the user owns the file. 

Solution:

Separate the queries for authenticated and anonymous users.


apps/frontend/src/app/[chain]/drive/metadata/[cid]/page.tsx
```diff
@@ -1,27 +1,48 @@
...
+
+    return mapObjectInformationFromQueryResultWithOwnership(data);
+  } else {
+    const { data } = await gqlClient.query<GetMetadataByHeadCidQuery>({
+      query: GetMetadataByHeadCidDocument,
+      variables: { headCid: cid },
+    });
+
+    return mapObjectInformationFromQueryResult(data);
+  }
```
Comment (jfrank-summit) [inline apps/frontend/src/app/[chain]/drive/metadata/[cid]/page.tsx:37]: Should we have a query failure path here?

PR #478, opened 2025-09-17
Fix download worker crashes by handling Readable stream errors and zero‑byte files

clostao:
This PR resolves a crash in the download worker caused by unhandled errors emitted from Node.js Readable streams and by unexpected zero‑byte files introduced during a migration. It adds centralized stream error handling and gracefully supports empty files to prevent the process from terminating.

## Root cause

Unhandled 'error' events from Readable streams caused “throw er; // Unhandled 'error' event” and crashed the worker.
Three zero‑byte files (introduced by a migration bug) triggered edge‑cases in partial/full retrieval paths and ultimately surfaced as stream errors.

## Impact
* Download worker no longer crashes on missing chunks or malformed/empty files.
* Errors are logged consistently with context instead of bubbling up as unhandled stream errors.
* Byte‑range and full file downloads behave correctly when encountering zero‑byte files.

### DIFF
--- a/apps/backend/src/infrastructure/services/dsn/objectMappingListener/index.ts
+++ b/apps/backend/src/infrastructure/services/dsn/objectMappingListener/index.ts
@@ -1,4 +1,3 @@
-import 'dotenv/config'
 import { NodesUseCases } from '../../../../core/objects/nodes.js'
 import { config } from '../../../../config.js'
 import { createLogger } from '../../../drivers/logger.js'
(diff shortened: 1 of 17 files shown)


apps/backend/src/infrastructure/services/dsn/objectMappingListener/index.ts
```diff
@@ -63,3 +65,5 @@
 export const objectMappingArchiver = {
   start,
 }
+
+objectMappingArchiver.start()
```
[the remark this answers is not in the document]
clostao: reverted change

apps/gateway/src/http.ts
```diff
@@ -31,5 +31,10 @@ export const internalRedirect = async (
   }
 
   // @ts-ignore
-  Readable.fromWeb(response.body).pipe(res);
+  const readable = Readable.fromWeb(response.body);
```
jfrank-summit, line 34: Can `response.body` ever be null here? If so, we should have a guard.

clostao: The target endpoint is the download endpoint for auto-drive that always has a body

jfrank-summit (approved):

PR #479, opened 2025-09-18
Unify task retry handling

clostao:
## Problem

For the task 'publish-nodes' we have a custom error management handler for avoiding re-publishing nodes twice. This custom handler was justified provided the publication of all nodes happen through the same task. But there's no big reason why we should process all through just one task.

## Solution

I've updated how the publication of the IPLD nodes work, instead of all in a single task there's a task for every node published (I kept the same interface so we can use manually it as an array). During the issue of the download worker we've seen that the rabbitmq instance can handle 100,000+ of messages w/o issues so configuring properly the pre-fetch should be enough. 

In addition, I added another task ('ensure-object-published') for easing operations in the case of some issue publishing.

## Impact 

Should improve the robustness of error management during node publication


apps/backend/src/infrastructure/eventRouter/processors/frontend.ts
```diff
@@ -7,13 +7,13 @@ import { createHandlerWithRetries } from '../utils.js'
...
 export const processFrontendTask = createHandlerWithRetries(
-  ({ id, params, retriesLeft }: Task) => {
+  ({ id, params }: Task) => {
     if (id === 'migrate-upload-nodes') {
       return UploadsUseCases.processMigration(params.uploadId)
     } else if (id === 'archive-objects') {
       return NodesUseCases.processNodeArchived(params.objects)
     } else if (id === 'publish-nodes') {
-      return OnchainPublisher.publishNodes(params.nodes, retriesLeft)
+      return OnchainPublisher.publishNodes(params.nodes)
```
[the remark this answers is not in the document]
clostao: Added

apps/backend/src/core/uploads/uploads.ts
```diff
@@ -310,19 +309,14 @@ const removeUploadArtifacts = async (uploadId: string): Promise<void> => {
 const scheduleNodesPublish = async (cid: string): Promise<void> => {
   const nodes = await NodesUseCases.getCidsByRootCid(cid)
 
-  const tasks: Task[] = chunkArray(
-    nodes,
-    config.params.maxUploadNodesPerBatch,
```
jfrank-summit, line 315: Should `maxUploadNodesPerBatch` be removed from `config`?

apps/backend/src/infrastructure/services/upload/onchainPublisher/index.ts
```diff
@@ -4,93 +4,52 @@ import { safeCallback } from '../../../../shared/utils/safe.js'
...
-import { EventRouter } from '../../../eventRouter/index.js'
-import { MAX_RETRIES } from '../../../eventRouter/tasks.js'
-import { config } from '../../../../config.js'
 
 const logger = createLogger('upload:onchainPublisher')
 
 export const transactionManager = createTransactionManager()
 
-const REPUBLISH_DELAY = 10_000
+const publishNodes = safeCallback(async (cids: string[]) => {
```
jfrank-summit, line 12: `safeCallback` will swallow errors which means using `publishNodes` inside `createHandlerWithRetries` won't lead to retries on node publish failures.

clostao: You're right

PR #484, opened 2025-09-22
fix: metadata endpoint

clostao:
## Problem

I misused the `handleInternalError` when it applied to `handleInternalErrorResult` this created a double nested Result object. The returning value of the endpoint was `{value: OffchainMetadata}` when the SDK is expecting just `OffchainMetadata`

## Solution 

Update the error handler `handleInternalErrorResult` for getting a single `Result` object so the returned value is the expected by the SDK

### DIFF
--- a/apps/backend/src/app/controllers/object.ts
+++ b/apps/backend/src/app/controllers/object.ts
@@ -183,7 +183,7 @@ objectController.get(
   asyncSafeHandler(async (req, res) => {
     const { cid } = req.params
 
-    const metadataResult = await handleInternalError(
+    const metadataResult = await handleInternalErrorResult(
       ObjectUseCases.getMetadata(cid),
       'Failed to get metadata',
     )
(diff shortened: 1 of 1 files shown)


clostao:
> ### Bug: Inconsistent Metadata Handling
> The `/object/:cid/metadata` endpoint now returns `200 OK` with a `null` body when metadata is not found, instead of a `404 Not Found` error. This occurs because the null check for `metadata` was removed, even though the underlying use case can return a successful result with a null value. This behavior is inconsistent with other similar endpoints in the controller that handle missing data with a `404`.

This incorrect, the use case in case of not having found the metadata entry returns an `ObjectNotFound` error:

```ts
  const entry = await metadataRepository.getMetadata(cid)
  if (!entry) {
    logger.info('No metadata found for object (cid=%s)', cid)
    return err(new ObjectNotFoundError(`Object with cid=${cid} not found`))
  }
``` 

This error is handled by the handleError method that maps this error to a HTTP 404 error

teor2345 (commented):
Looks good, thanks!

PR #486, opened 2025-09-22
Buy credits backend

clostao:


clostao:
@cursor review

apps/backend/src/infrastructure/repositories/users/intents.ts
```diff
@@ -0,0 +1,80 @@
...
+  const db = await getDatabase()
+  const result = await db.query<DBIntent>(
+    'UPDATE intents SET status = $1, user_public_id = $2, tx_hash = $3, deposit_amount = $4 WHERE id = $5 RETURNING *',
+    [
+      intent.status,
+      intent.userPublicId,
+      intent.txHash ?? null,
+      intent.depositAmount?.toString() ?? null,
+      intent.id,
+    ],
```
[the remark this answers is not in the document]
clostao: Nope, added

apps/backend/migrations/sqls/20250915125036-payments-up.sql
```diff
@@ -0,0 +1,9 @@
+
+CREATE TABLE intents (
+  id VARCHAR(255) PRIMARY KEY,
+  user_public_id VARCHAR(255) NOT NULL,
+  status VARCHAR(32) NOT NULL,
+  tx_hash VARCHAR(255),
+  deposit_amount VARCHAR(255),
```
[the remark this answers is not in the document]
clostao: Makes sense

apps/backend/src/app/controllers/intents.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    const result = await handleInternalErrorResult(
+      IntentsUseCases.getIntent(req.params.id),
+      'Failed to get intent',
+    )
+    if (result.isErr()) {
+      handleError(result.error, res)
+      return
+    }
```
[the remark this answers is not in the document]
clostao: Makes sense, added

apps/backend/migrations/sqls/20250915125036-payments-up.sql
```diff
@@ -0,0 +1,9 @@
...
+  price_per_mb float
```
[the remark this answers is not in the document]
clostao: This is being updated for using transaction byte fee + premium defaults to '400%'

apps/backend/package.json
```diff
@@ -23,6 +23,7 @@
     "@auto-drive/s3": "workspace:*",
     "@auto-files/rpc-apis": "workspace:*",
     "@autonomys/asynchronous": "^1.5.15",
+    "@autonomys/auto-consensus": "^1.5.18",
```
jfrank-summit, line 26: I would update the rest of the packages too as sometimes there will be resolution issues with multiple versions.

apps/backend/src/config.ts
```diff
@@ -65,6 +65,7 @@ export const config = {
     contractAddress: getAddress(env('EVM_CHAIN_CONTRACT_ADDRESS')),
     confirmations: Number(env('EVM_CHAIN_CONFIRMATIONS', '12')),
     checkInterval: Number(env('EVM_CHAIN_CHECK_INTERVAL', '30000')),
+    premiumPct: Number(env('EVM_CHAIN_PREMIUM_PCT', '4.00')), // 400% premium
```
jfrank-summit, line 68: The naming here is confusing. If a variable has Pct or Percent in its name, I usually expect it to be divided by 100 before use (so 4.00 → 0.04 = 4%). Since the code is using it directly as a factor, maybe `priceMultiplier` would be clearer. In that case 4.0 would mean a 300% premium (final price = 4× base), which avoids the “is this 4% or 400%?” ambiguity.

apps/backend/package.json
clostao: I'd prefer update them in the feature branch once, for avoiding intermediary conflicts merging

apps/backend/src/config.ts
clostao: Updated to `priceMultiplier` in name and calculation

jfrank-summit: Did you push? I don't see the update.

PR #492, opened 2025-09-24
fix: downloads when user not logged in

clostao:
## Description

Downloads for unlogged users were failing because `@autonomys/auto-drive@<1.5.8` didn't allowed none authenticated users.


jfrank-summit (approved):
LGTM

PR #495, opened 2025-09-25
feat: implement buy credits into frontend (Buy Credits PR #3)

clostao:
I closed the previous one since it had a bad git history, this one already start from the `buy-credits-backend` branch.

## Description 

Implements a multi step flow for buying credits using AI3

## Demo

https://www.loom.com/share/5b4e3096e83b4c5c8eb681ad13c235d7?sid=283d8898-2e09-4fd0-a507-d91b0067d30c

### DIFF
--- a/apps/frontend/src/components/molecules/AccountInformation/index.tsx
+++ b/apps/frontend/src/components/molecules/AccountInformation/index.tsx
@@ -1,4 +1,4 @@
-import bytes from 'bytes';
+import { truncateBytes } from '../../../utils/number';
 import { utcToLocalRelativeTime } from '../../../utils/time';
 import { SubscriptionGranularity } from '@auto-drive/models';
 
@@ -19,16 +19,14 @@ export const AccountInformation = ({
 
   const uploadPercentage = (uploadUsed / uploadLimit) * 100;
 
-
   return (
     <div className='space-y-2'>
       <div className='text-xs text-muted-foreground'>Upload usage</div>
       <div className='space-y-1'>
         <div className='flex justify-between text-xs'>
-          <span>{bytes(uploadPending, { decimalPlaces: 3 })} left</span>
+          <span>{truncateBytes(uploadPending)} left</span>
           <span className='text-muted-foreground'>
-            {bytes(uploadUsed, { decimalPlaces: 3 })}/
-            {bytes(uploadLimit, { decimalPlaces: 3 })}
+            {truncateBytes(uploadUsed)}/{truncateBytes(uploadLimit)}
           </span>
         </div>
         <div className='h-1.5 w-full rounded-full bg-muted'>
--- a/apps/frontend/src/utils/number.ts
+++ b/apps/frontend/src/utils/number.ts
@@ -1,3 +1,5 @@
+import bytes from 'bytes';
+
 // Format number with commas
 export const formatNumberWithCommas = (num?: number): string => {
   if (num === undefined) return 'N/A';
@@ -10,4 +12,22 @@ export const truncateNumberWithDecimals = (
 ): number => {
   const precision = 10 ** decimals;
   return Math.floor(num * precision) / precision;
+};
+
+export const truncateBytes = (num: number): string => {
+  const f = bytes(num, { decimalPlaces: 2 });
+
+  // Extract the numeric part and unit using regex
+  const match = f.match(/^(\d+(?:\.\d+)?)([KMGT]?B)$/);
+  if (!match) return f;
+
+  const [, numericPart, unit] = match;
+
+  // If it contains a decimal point, truncate to integer
+  if (numericPart.includes('.')) {
+    const integerPart = numericPart.split('.')[0];
+    return `${integerPart}${unit}`;
+  }
+
+  return f;
 };


apps/frontend/src/components/views/PurchaseCredits/steps/Step4_Success.tsx
```diff
@@ -0,0 +1,78 @@
...
+    <div className='flex flex-col gap-4'>
+      <Section title='Payment Successful!'>
+        <Card>
+          <div className='flex flex-col gap-3 p-4'>
+            <div className='rounded-md bg-green-50 p-4'>
+              <div className='grid grid-cols-2 gap-2'>
+                <InfoRow
+                  className='items-center font-bold'
+                  label='Storage Added'
+                  value={<span>1GB</span>}
```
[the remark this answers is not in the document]
clostao: Solved

apps/frontend/src/components/atoms/AskForCredits.tsx
```diff
@@ -1,16 +1,15 @@
-import { EXTERNAL_ROUTES, Button } from '@auto-drive/ui';
+import { Button, ROUTES } from '@auto-drive/ui';
+import { InternalLink } from './InternalLink';
+import { useNetwork } from '../../contexts/network';
 
-export const AskForCreditsButton = () => {
+export const BuyMoreCreditsButton = () => {
```
teor2345, line 5: It's confusing to have this file named ask for credits when it's now about buying credits.
This would also need to change in the nav bar.

apps/frontend/src/components/views/PurchaseCredits/steps/Step1_SelectPackage.tsx
```diff
@@ -0,0 +1,190 @@
...
+              </div>
+            </div>
+            <div className='flex items-start gap-3'>
+              <div className='flex h-8 w-8 items-center justify-center rounded-full bg-primary text-sm font-bold text-primary-foreground'>
+                3
+              </div>
+              <div>
+                <div className='font-medium'>Instant Credits</div>
+                <div className='text-sm text-muted-foreground'>
+                  Once transaction is confirmed (12 confirmations) you will
```
teor2345, line 180: Should this be a constant rather than a hard-coded number of confirmations?

I don't think it matter as much either way, because in normal operation we would only have a 12 block reorg after thousands of years. (However, if there is a network split or high load, it could happen much sooner.)

packages/ui/src/constants/contracts.ts
```diff
@@ -0,0 +1,81 @@
+import { Address } from "viem";
+import { NetworkId } from "./networks";
+
+export const paymentReceiverContractsByNetworkId: Record<NetworkId, Address> = {
+  [NetworkId.TAURUS]: "0x0000000000000000000000000000000000000000",
+  [NetworkId.MAINNET]: "0x0000000000000000000000000000000000000000",
```
teor2345, line 6: Does this need a TODO comment?

packages/ui/src/constants/contracts.ts
```diff
@@ -0,0 +1,81 @@
+import { Address } from "viem";
+import { NetworkId } from "./networks";
+
+export const paymentReceiverContractsByNetworkId: Record<NetworkId, Address> = {
+  [NetworkId.TAURUS]: "0x0000000000000000000000000000000000000000",
```
teor2345, line 5: Do you mean Chronos here? (And in the next file.)

packages/ui/src/constants/evm.ts
```diff
@@ -0,0 +1,59 @@
...
+      },
+    },
+  },
+  [NetworkId.LOCAL]: {
+    // Anvil default chain id
+    id: 8700,
+    name: "Chronos Testnet",
+    nativeCurrency: {
+      name: "Chronos AI3",
+      symbol: "tAI3",
```
teor2345, line 50: It's confusing to have mainnet using ETH and testnets using tAI3, is that intentional?
If it is, it might be worth leaving a comment explaining why.

apps/frontend/src/components/views/PurchaseCredits/CreditCurrentPrice.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const { shannonsPerByte, formatCreditsAsAi3 } = usePrices();
+
+  return (
+    <div className='rounded-xl bg-primary p-6 text-primary-foreground'>
+      <div className='flex flex-col items-start justify-between gap-4 md:flex-row md:items-center'>
+        <div>
+          <div className='text-sm opacity-90'>Current Price</div>
+          <div className='text-3xl font-bold'>
+            {typeof shannonsPerByte === 'number'
+              ? `${formatCreditsAsAi3(MB).toFixed(2)} AI3 per MB`
```
teor2345, line 15: I've read the code, but I'm not sure I understand the rounding strategy.

I think it should be something like:
- when the user is buying, round up the price and megabytes, and round down the credits they'll get, so they always have enough to cover any hidden fractional upload size
- for amounts the user holds, round down the credits they have, so they don't try and upload beyond their actual capacity

teor2345 (commented):
No blockers, except maybe for TAURUS still being added to new code.

packages/ui/src/constants/contracts.ts
clostao: Marked for removal, in following PR

apps/frontend/src/components/views/PurchaseCredits/CreditCurrentPrice.tsx
clostao: Makes sense your approach, I didn't implemented that sensitive. The `Number.toFixed` method has Round as its implicit rounding strategy. So, basically we have to explicitly truncate the amounts in two cases:

- Converting AI3 into credits
- Showing user's balance

I'll add this two

apps/frontend/src/components/views/PurchaseCredits/steps/Step1_SelectPackage.tsx
clostao: What would you recommend? Should we have a higher confirmation threshold?

apps/frontend/src/components/views/PurchaseCredits/CreditCurrentPrice.tsx
clostao: Updated

PR #497, opened 2025-09-26
Update subscription data model

clostao:
## Problem

Subscription really doesn't describe well the actual entity, therefore we update for accounts that they do.

## How it works?

Renames `granularity` -> `model` and `Subscription` -> `Account` at frontend, backend & database level


apps/backend/src/app/controllers/accounts.ts
jfrank-summit, line 77: There are still some references to `granularity` that should be changed to `model`.

jfrank-summit (commented):
It looks like there are still quite a few references to `granularity` that should be `model`. I only commented on one.

PR #501, opened 2025-10-01
Buy credits feature flagging (#5)

clostao:
## Problem

We want to safely rollout the buy credits feature in a progressive manner

## Solution

This PR implements a environment variable based feature flagging system that has three posibles states:

- Active For All
- Active For Employees (via allowlisted of domains)
- Disabled

## Structure

- Introduced `/features` controller that given authorization params it returns available features
- Environment-based variables with two fields 'active' and `employeeOnly`

### DIFF
--- a/apps/backend/src/config.ts
+++ b/apps/backend/src/config.ts
@@ -1,5 +1,6 @@
 import 'dotenv/config'
 import { SubscriptionGranularity } from '@auto-drive/models'
+import { FeatureFlag } from './core/featureFlags/index.js'
 import { optionalBoolEnvironmentVariable, env } from './shared/utils/misc.js'
 import { getAddress } from 'viem'
 
@@ -106,17 +107,17 @@ export const config = {
           (optionalBoolEnvironmentVariable('TASK_MANAGER_ACTIVE') ||
             optionalBoolEnvironmentVariable('ALL_SERVICES_ACTIVE')) &&
           !optionalBoolEnvironmentVariable('TASK_MANAGER_DISABLED'),
-      },
+      } as FeatureFlag,
       objectMappingArchiver: {
         active:
           (optionalBoolEnvironmentVariable('OBJECT_MAPPING_ARCHIVER_ACTIVE') ||
             optionalBoolEnvironmentVariable('ALL_SERVICES_ACTIVE')) &&
           !optionalBoolEnvironmentVariable('OBJECT_MAPPING_ARCHIVER_DISABLED'),
-      },
+      } as FeatureFlag,
       buyCredits: {
         active: optionalBoolEnvironmentVariable('BUY_CREDITS_ACTIVE'),
         staffOnly: optionalBoolEnvironmentVariable('BUY_CREDITS_STAFF_ONLY'),
-      },
+      } as FeatureFlag,
     },
     allowlistedUsernames: env('STAFF_USERNAME_ALLOWLIST', '')
       .split(',')
--- a/apps/backend/src/config.ts
+++ b/apps/backend/src/config.ts
@@ -98,7 +98,7 @@ export const config = {
     },
     forbiddenExtensions: env('FORBIDDEN_EXTENSIONS', '').split(','),
   },
-  services: {
+  featureFlags: {
     taskManager: {
       active:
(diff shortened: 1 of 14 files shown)


apps/backend/src/config.ts
```diff
@@ -97,20 +97,29 @@ export const config = {
...
+          !optionalBoolEnvironmentVariable('OBJECT_MAPPING_ARCHIVER_DISABLED'),
+      },
+      buyCredits: {
+        active: optionalBoolEnvironmentVariable('BUY_CREDITS_ACTIVE'),
+        employeeOnly: optionalBoolEnvironmentVariable(
+          'BUY_CREDITS_EMPLOYEE_ONLY',
+        ),
+      },
     },
+    employeeDomains: env('BUY_CREDITS_EMPLOYEE_DOMAINS', '').split(','),
```
teor2345, line 123: Feature design:
Should the allow list also allow specific email addresses?
For example, if we had volunteers/ambassadors/etc who we wanted to give access to test the system.

Optional nitpick:
Would it be shorter and easier to spell if we called this "staff" instead?

apps/backend/src/core/featureFlags/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  logger.debug('Checking if user is employee:', user)
+  return Boolean(
+    user &&
+      user.oauthProvider !== 'web3-wallet' &&
+      config.featureFlags.employeeDomains
+        .filter((domain) => domain)
+        .some(
+          (domain) =>
+            user.oauthUsername &&
+            user.oauthUsername.toLowerCase().endsWith(domain),
```
teor2345, line 38: This check isn't quite correct.

For example, if the list was:
`autonomys.xyz`

Then `endsWith` would also match:
```
fakeautonomys.xyz
wwwautonomys.xyz
```

I think we'll want to do a domain-component aware match, so something like:
1. Split the username on `@` and `.`, and discard the user part (before the `@`)
2. Split the domain on `.`
3. Check the components of the username and domain, in reverse order, stopping when you get to the end of the domain

If we end up supporting full email addresses, we can check if the "domain" contains an `@`, and if it does, do a case-insensitive match of the full strings.

teor2345 (commented):
I'm disappointed the AI didn't find the string matching bug 😢

apps/backend/src/core/featureFlags/index.ts
clostao: I understand the issue, my intent with this check was to support as well the addition of complete emails. So, for adding a specific domain, the config should be: `@autonomys.xyz` rather than `autonomys.xyz`.

teor2345: I think that would be best.

If we have a separate domain list, then we can add `@` to the start of every domain in the list, if it doesn't start with it already. That would reject subdomains, which is probably ok.

And email lists should match the entire string.

apps/backend/src/core/featureFlags/index.ts
```diff
@@ -0,0 +1,46 @@
...
+        ),
+  )
```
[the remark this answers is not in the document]
clostao: outdated

apps/backend/src/core/featureFlags/express.ts
```diff
@@ -0,0 +1,35 @@
...
+    const user = await handleAuth(req, res)
+    if (!user) {
+      return
+    }
+
+    return FeatureFlagsUseCases.get(user)
+  }
+
+  return FeatureFlagsUseCases.get(null)
+}
```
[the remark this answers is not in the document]
clostao: Good catch!

apps/backend/src/core/featureFlags/index.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+const isActive = (value: FeatureFlag, user: User | null) => {
+  if (value.active) {
+    return true
+  }
+
+  return value.staffOnly && isStaff(user)
+}
```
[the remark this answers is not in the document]
clostao: Not a bug, if active is true should ignore staffOnly

PR #502, opened 2025-10-03
fix: outdated testing

clostao:
## Issue

There was some outdated tests failing that bypassed the CI/CD pipeline due to a misconfiguration in the workflow.

## Solution

Fix tests & fix workflow configuration


clostao:
I fixed the build issue, AI is not right, allowing duplicates is the way to go

PR #515, opened 2025-10-17
Add top users panel (backend)

clostao:
### Problem

We want to add a list of the top users (in terms of upload usage in bytes) in the admin panel

### Solution

Added an endpoint in backend service that returns the top accounts in terms of usage (including their `organizationId`) and then, an endpoint (`/organizations/:orgId`) in auth service (admin only) that returns the organization information with their members.

### DIFF
--- a/apps/backend/migrations/sqls/20251020202315-interactions-index-up.sql
+++ b/apps/backend/migrations/sqls/20251020202315-interactions-index-up.sql
@@ -1,1 +1,1 @@
-CREATE INDEX IF NOT EXISTS idx_interactions_account_created_at ON public.interactions (account_id, created_at);
+CREATE INDEX IF NOT EXISTS idx_interactions_account_created_at ON public.interactions ("type", account_id, created_at);


apps/backend/src/infrastructure/repositories/users/accounts.ts
```diff
@@ -75,10 +75,24 @@ const getAll = async (): Promise<Account[]> => {
...
 
+export const getTopAccounts = async (
+  limit: number = 10,
+  type: InteractionType = InteractionType.Upload,
+): Promise<Account[]> => {
+  const db = await getDatabase()
+
+  const result = await db.query<DBAccount>(
+    'SELECT a.*, SUM(i.size) as total_size FROM accounts a INNER JOIN interactions i ON a.id = i.account_id WHERE i.type = $1 GROUP BY a.id ORDER BY total_size DESC LIMIT $2',
+    [type, limit],
```
jfrank-summit: We should be able to limit this to a specific period.

apps/backend/src/infrastructure/repositories/users/accounts.ts
```diff
@@ -75,10 +97,26 @@ const getAll = async (): Promise<Account[]> => {
...
+export const getTopAccountsWithinPeriod = async (
+  limit: number = 10,
+  type: InteractionType = InteractionType.Upload,
+  fromDate: Date,
+  toDate: Date,
+): Promise<AccountWithTotalSize[]> => {
+  const db = await getDatabase()
+
+  const result = await db.query<DBAccountWithTotalSize>(
+    'SELECT a.*, SUM(i.size) as total_size FROM accounts a INNER JOIN interactions i ON a.id = i.account_id WHERE i.type = $1, i.created_at > $2 and i.created_at < $3 GROUP BY a.id ORDER BY total_size DESC LIMIT $4',
```
jfrank-summit: In the `WHERE` clause you are using a comma instead of `AND`.
```suggestion
    'SELECT a.*, SUM(i.size) as total_size FROM accounts a INNER JOIN interactions i ON a.id = i.account_id WHERE i.type = $1 AND i.created_at > $2 AND i.created_at < $3 GROUP BY a.id ORDER BY total_size DESC LIMIT $4',
```

jfrank-summit (changes requested):
A couple of comments.

jfrank-summit (approved):
LGTM

PR #521, opened 2025-10-23
Retiring things

clostao:
## Retiring reasons

Since we retired Taurus network we did the same with auto-drive taurus' instance. Additionally, we decided not to deploy auto-drive on Chronos. Therefore having a multi-network gateway was not useful since there would only be one instance to route to adding some overhead so we decided to eliminate it.

### DIFF
--- a/DEPLOYMENT_GUIDE.md
+++ b/DEPLOYMENT_GUIDE.md
@@ -231,7 +231,7 @@ Services are organized into profiles:
 - **download**: Download-related services (API, Worker)
 - **rabbit**: RabbitMQ message queue
 
-### Multinetwork Gateway (deprecated)
+### Multinetwork Gateway (retired)
 
 The gateway service provides:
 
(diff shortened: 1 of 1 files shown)


DEPLOYMENT_GUIDE.md
```diff
@@ -239,7 +231,7 @@ Services are organized into profiles:
 - **download**: Download-related services (API, Worker)
 - **rabbit**: RabbitMQ message queue
 
-### Multinetwork Gateway
+### Multinetwork Gateway (deprecated)
```
jfrank-summit: As Nazar recently pointed out to me, we are actually retiring the gateway, not deprecating it. Deprecating implies it will still exist, just not be fully supported.

DEPLOYMENT_GUIDE.md
```diff
@@ -239,7 +231,7 @@ Services are organized into profiles:
 - **download**: Download-related services (API, Worker)
 - **rabbit**: RabbitMQ message queue
 
-### Multinetwork Gateway
+### Multinetwork Gateway (retired)
```
teor2345: Can this section just be deleted instead?

clostao: Jeremy asked me to keep it

PR #529, opened 2025-11-04
feat: keep alive mechanism in rabbitmq

clostao:
# Problem

We detected an issue in which the publication of a message in rabbitmq failed because the channel was close, attached logs below.

# Solution

- Create a keep alive process that will use that channel for making ensure still alive every 60s
- Wrapping the `RabbitMQ.publish` method with a retry handler with backoff so in the unlikely case that the channel closes between the last keep alive check and a message publication, the message will be re-sent after the reconnection
- Persist in memory the callbacks for each queue and resurrect them when channel is re-established

### DIFF
--- a/apps/backend/src/config.ts
+++ b/apps/backend/src/config.ts
@@ -51,6 +51,7 @@ export const config = {
   rabbitmq: {
     url: env('RABBITMQ_URL'),
     prefetch: Number(env('RABBITMQ_PREFETCH', '10')),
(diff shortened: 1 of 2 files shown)


jfrank-summit (approved):
LGTM!

jfrank-summit:
> @cursoragent review

Is this different than bugbot? It gave a report instead of inline comments. 🤔

PR #531, opened 2025-11-04
chore: remove taurus from options on backend deployment workflow

0xjeremyfrank:
Removed taurus from options on backend deployment workflow.


jfrank-summit:
@carlos beat me to it #530

PR #532, opened 2025-11-07
feat: add treasury contract

clostao:
## Description 

A new contract that will store funds generated by auto-drive credits sale.

### DIFF
--- a/nginx/backend.mainnet.public.conf
+++ b/nginx/backend.mainnet.public.conf
@@ -126,7 +126,9 @@ server {
         rewrite /folder/(.*) /downloads/$1?ignoreEncoding=false break;
     }
 
-
+    location / {
+        return 301 https://explorer.ai3.storage$request_uri;
+    }
 
     listen 443 ssl http2; # managed by Certbot
     ssl_certificate /etc/letsencrypt/live/public.auto-drive.autonomys.xyz/fullchain.pem; # managed by Certbot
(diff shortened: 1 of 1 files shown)


packages/contracts/src/AutoDriveTreasury.sol
```diff
@@ -0,0 +1,66 @@
...
+
+        uint256 balance = address(this).balance;
+        if (balance < minimumBalance) {
+            revert InsufficientBalance(balance, minimumBalance);
+        }
+
+        uint256 withdrawAmount = balance - minimumBalance;
+        if (withdrawAmount == 0) {
+            revert InsufficientBalance(balance, minimumBalance);
+        }
```
jfrank-summit: You can avoid the second check by making the first check `<=` since when `balance` = `minimumBalance` `withdrawalAmount` will = 0:
```suggestion
        if (balance <= minimumBalance) {
            revert InsufficientBalance(balance, minimumBalance);
        }

        uint256 withdrawAmount = balance - minimumBalance;
```

packages/contracts/src/AutoDriveTreasury.sol
```diff
@@ -0,0 +1,63 @@
...
+     * @notice Allows the owner to withdraw all funds leaving at least the minimum balance
+     * @param recipient The address to receive the withdrawn funds
+     */
+    function withdraw(address payable recipient) public onlyOwner nonReentrant {
+        if (recipient == address(0)) {
+            revert InvalidRecipient(recipient);
+        }
+
+        uint256 balance = address(this).balance;
+        if (balance < minimumBalance) {
```
jfrank-summit: Since you removed the withdraw = 0 check, this should be updated to `<=`
```suggestion
        if (balance <= minimumBalance) {
```

PR #573, opened 2025-12-18
chore: bump Auto SDK version to latest (1.6.5)

EmilFattakhov:
### Summary

Updates the Autonomys Auto SDK dependency set to v1.6.5


jfrank-summit:
> Hey @jfrank-summit - do I understand correctly that we need to merge the [corresponding auto-files-gateway ](https://github.com/autonomys/auto-files-gateway/pull/152) PR first?

I don't think the submodule should be updated in this PR.

PR #582, opened 2026-01-29
fix: download modal flashing when downloading compressed files

EmilFattakhov:
Was finally able to debug the issue when, for some files, the modal window flashes, but no download happens.

This PR fixes infinite refresh loop in download modal for unencrypted compressed files.

The issue occurred because of double decompression:
1. SDK sends `?ignoreEncoding=true` to indicate it will handle decompression
2. Backend's file-server library incorrectly still set `shouldDecompressBody=true`
3. Backend decompressed the data before sending
4. SDK received already-decompressed data and tried to decompress again
5. "invalid zlib data" error triggered retry loop, causing modal to flash

The fix overrides the library's behavior: when `ignoreEncoding=true` is passed,
skip server-side decompression and let the SDK handle it as intended.

Note: Encrypted files were unaffected since the backend correctly skips
decompression for encrypted content.

### DIFF
--- a/apps/frontend/src/utils/file.ts
+++ b/apps/frontend/src/utils/file.ts
@@ -1,8 +1,8 @@
 import { OffchainMetadata } from '@autonomys/auto-dag-data';
 
 export class InvalidDecryptKey extends Error {
-  constructor() {
-    super('Invalid decrypt key');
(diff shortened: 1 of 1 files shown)


jim-counter (approved):
LGTM!

PR #585, opened 2026-02-05
Fix/cursor identified performance issues

EmilFattakhov:
## Summary

This PR includes performance optimizations for the admin dashboard and improved error handling for encrypted file downloads.

### 🚀 Performance: Server-side Top Accounts Aggregation

Replaced client-side aggregation with a PostgreSQL function that performs `GROUP BY`, `SUM`, `ORDER BY`, and `LIMIT` directly in the database.

**Before:** Fetched all interaction records to the browser, then aggregated in JavaScript  
**After:** Returns only the top 10 accounts per query from the database

This prevents memory issues and potential browser crashes when querying large date ranges with millions of interactions.

**Files changed:**
- `apps/hasura/migrations/.../up.sql` - New `get_top_accounts_by_interaction` function
- `apps/hasura/metadata/...` - Hasura function configuration
- `apps/frontend/.../AdminStats.tsx` - Updated to use server-side aggregation

### 🐛 Fix: Improved Decryption Error Detection

Expanded error detection to catch crypto failures that were previously missed, ensuring users see the password retry prompt instead of a generic "Download failed" message.

**Now detects:**
- Node.js crypto errors: `"Unsupported state or unable to authenticate data"`
- Web Crypto API: `OperationError` exceptions

**Files changed:**
- `apps/frontend/src/utils/file.ts`

### 🧹 Cleanup: Remove Unused GraphQL Queries

Deleted `query.graphql` which contained 3 unused queries (`GetUploadStats`, `GetDownloadStats`, `GetAllStats`). The component uses inline `gql` template literals instead.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,170 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+Auto Drive is a decentralized content-addressed storage platform built on the Autonomys Network. It stores files as IPLD (InterPlanetary Linked Data) nodes on the Autonomys Distributed Storage Network (DSN). The system provides a web UI, REST API, S3-compatible API, and GraphQL API.
+
+## Repository Structure
+
+Yarn 4.2.2 monorepo with workspaces in `apps/`, `packages/`, and `submodules/`.
+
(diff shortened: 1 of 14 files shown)


EmilFattakhov:
> LGTM. How come you switched back to client-side top 10 aggregation?

It probably wasn't the best call; I've reverted it back to use the server-side aggregation. The rest of the PR is unchanged.

PR #587, opened 2026-02-09
Fix/frontend worker memory leaks

EmilFattakhov:
# Summary

The frontend worker process (`frontendWorker.ts`) was crashing with OOM errors. Root cause investigation found multiple contributing issues:

### 1. In-memory download cache defaulted to 10GB
The `LRUCache<string, Buffer>` holding downloaded file buffers directly in the V8 heap had a 10GB limit (`BigInt(10 * 1024 ** 3)`), far exceeding Node.js's typical heap capacity (~1.5–4GB). The process would OOM before LRU eviction ever triggered.

**Fix:** Reduced to 1GB; still configurable via `MEMORY_DOWNLOAD_CACHE_MAX_SIZE`. The filesystem cache (`fsCache`) continues to handle bulk storage separately.

### 2. No resource cleanup on any service
- **`paymentManager.start()`** created a `setInterval` and a `viem.watchContractEvent` — the interval ID was discarded and the watcher had no unsubscribe path. Multiple calls to `start()` would silently accumulate duplicate watchers and intervals.
- **`objectMappingArchiver.start()`** opened a WebSocket RPC connection that was never closeable. Reconnections could leak prior connection state.

**Fix:** Added `stop()` methods to both services.

### 3. No graceful shutdown
The worker had no SIGTERM/SIGINT handlers, so container orchestrators (Docker, k8s) could only force-kill it, leaving connections, intervals, and subscriptions dangling.

**Fix:** Added signal handlers that cleanly shut down all services and the RabbitMQ connection.

### 4. O(n×m) CID lookup in `processNodeArchived`
`Array.includes()` inside a `.filter()` loop caused quadratic CPU time on large object mapping batches from the indexer. This blocks the event loop and delays garbage collection, contributing to memory pressure. Also identified a redundant duplicate CID computation.

**Fix:** Replaced with `Set.has()` for O(1) lookups and eliminated the redundant computation.

## Commits

1. **fix: reduce default in-memory download cache from 10GB to 1GB** — `config.ts`
2. **fix: add graceful shutdown and resource cleanup to frontend worker** — `paymentManager`, `objectMappingListener`, `frontendWorker`
3. **perf: use Set for O(1) CID lookups in processNodeArchived** — `nodes.ts`

### DIFF
--- a/apps/backend/src/config.ts
+++ b/apps/backend/src/config.ts
@@ -4,7 +4,7 @@ import { AccountModel } from '@auto-drive/models'
 import { optionalBoolEnvironmentVariable, env } from './shared/utils/misc.js'
 import { getAddress } from 'viem'
 
-const DEFAULT_MAX_CACHE_SIZE = BigInt(1024 ** 3) // 1GB
+const DEFAULT_MEMORY_CACHE_MAX_SIZE = BigInt(1024 ** 3) // 1GB
 
 const DEFAULT_CACHE_MAX_SIZE = 10 * 1024 ** 3 // 10GB
 const DEFAULT_CACHE_TTL = 0 // No TTL
@@ -28,7 +28,7 @@ export const config = {
   },
   memoryDownloadCache: {
     maxCacheSize: Number(
-      env('MEMORY_DOWNLOAD_CACHE_MAX_SIZE', DEFAULT_MAX_CACHE_SIZE.toString()),
+      env('MEMORY_DOWNLOAD_CACHE_MAX_SIZE', DEFAULT_MEMORY_CACHE_MAX_SIZE.toString()),
     ),
   },
   objectMappingArchiver: {


apps/backend/src/config.ts
```diff
@@ -4,7 +4,7 @@ import { AccountModel } from '@auto-drive/models'
 import { optionalBoolEnvironmentVariable, env } from './shared/utils/misc.js'
 import { getAddress } from 'viem'
 
-const DEFAULT_MAX_CACHE_SIZE = BigInt(10 * 1024 ** 3)
+const DEFAULT_MAX_CACHE_SIZE = BigInt(1024 ** 3) // 1GB
```
jim-counter: Might be better as `DEFAULT_MEMORY_CACHE_MAX_SIZE`? Bit confusing with the next `const`.

PR #588, opened 2026-02-10
Infra/add staging environment docker configuration

EmilFattakhov:
### Add Staging Environment Infrastructure

Adds Docker configuration and environment setup for staging deployments.

### Changes:

- docker-compose.staging.yml - Complete Docker Compose configuration for staging environment with all services (backend, frontend, Hasura, RabbitMQ, etc.)
- apps/frontend/Dockerfile - Frontend container build configuration
- apps/frontend/next.config.mjs - Next.js config updates for containerized builds
- .env.staging.example - Example environment file with all required variables for staging


docker-compose.staging.yml
```diff
@@ -0,0 +1,308 @@
...
+    networks:
+      - auto-drive-network
+
+  # ---------------------------------------------------------------------------
+  # BACKEND API
+  # ---------------------------------------------------------------------------
+  backend-api:
+    build:
+      context: .
+      dockerfile: apps/backend/Dockerfile
```
vedhavyas: for production, we should ideally have Github actions that can create these docker images under ghcr. User can simply pull from them instead of building.

If we have a new docker-compose for local development, then its its ideal to have the build command in that compose file

vedhavyas (commented):
This looks good. Dropped some comments.

Finally, only one docker container should expose port which I'm assuming is the frontend.

In the production, we put a reverse proxy like traefik and expose just ports 80 and 443 instead of 8080

EmilFattakhov:
Closed as per: https://github.com/autonomys/infra/issues/556#issuecomment-4705981655

PR #591, opened 2026-02-12
Sync prod with main

EmilFattakhov:
Release to include:

https://github.com/autonomys/auto-drive/pull/580 and https://github.com/autonomys/auto-drive/pull/587 and https://github.com/autonomys/auto-drive/pull/593

PR #593, opened 2026-02-13
fix(intents): reset priceApiPromise on connection failure

EmilFattakhov:
If ApiPromise.create() rejects, the .then() handler never fires so priceApiPromise holds a rejected promise permanently. Since rejected promises are truthy, subsequent calls return the same failed promise.

Added .catch() handler to reset priceApiPromise to null on initial connection failure, allowing recovery on next getPriceApi() call.

Needs to be merged before https://github.com/autonomys/auto-drive/pull/591


EmilFattakhov:
@cursoragent review

PR #597, opened 2026-02-25
feat(backend): expose GET /intents/price outside buyCredits feature flag

EmilFattakhov:
Register the price endpoint directly on the app before the feature-flagged /intents router so it is always publicly accessible regardless of whether the buyCredits feature flag is enabled.


EmilFattakhov:
bugbot run

jim-counter (approved):
Agree with your comment - this looks how I expected it to. LGTM.

PR #599, opened 2026-02-27
Feature/admin dashboard stats

EmilFattakhov:
---
  Summary

  - Add "Total Users" stat card to the Admin Dashboard stats panel, fetched from the auth service via
  AuthService.getUserList.
  - Track CID per interaction — adds a cid column to the interactions table so each upload and download event
   records which file (by CID) was accessed. The CID is passed through from uploadProcessing and
  downloads/sync down to the repository insert.
  - Show CID in the Activity Log — the OrganizationDetails admin view now displays a linked, truncated CID
  for each interaction row, linking to the object detail page.
  - Graceful Hasura fallback — the GET_ACCOUNT_INTERACTIONS GraphQL query falls back to a legacy query
  (without cid) if the migration hasn't been applied yet, so the admin panel keeps working on environments
  that haven't run the migration.

  Deployment steps

  1. Database migration

  yarn backend db-migrate up

  This applies 20260227000000-interactions-add-cid, which:
  - Adds cid TEXT NULL to public.interactions
  - Creates index idx_interactions_cid

  2. Hasura metadata

  cd apps/hasura
  hasura metadata apply --admin-secret <HASURA_ADMIN_SECRET>
  cd -

  This exposes the new cid column in the interactions select permission.

  Note: Steps 1 and 2 must be done before or alongside the backend/frontend deploy. If only the frontend is
  deployed first, the GraphQL fallback query (GET_ACCOUNT_INTERACTIONS_LEGACY) will be used automatically
  until the migration + metadata are applied — no downtime.

  3. Deploy backend & frontend

  Standard deploy — no additional environment variables required.

  Test plan

  - Run yarn backend db-migrate up and verify the cid column and index exist on interactions
  - Apply Hasura metadata and confirm cid appears in the GraphQL schema
  - Upload a file and verify the interaction row has the correct cid value
  - Download a file and verify the interaction row records the cid
  - Open Admin → Organization Details → Activity Log and confirm the CID column renders with a link for new
  interactions and — for historical rows
  - Confirm the "Total Users" stat card appears and shows the correct count
  - On an environment without the migration applied, verify the admin panel still loads (fallback query
  fires)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,5 @@ environment.yaml
 
 # Next
 **/.next/
+# Claude Code worktrees
+.claude/worktrees/


EmilFattakhov:
bugbot run

PR #601, opened 2026-03-03
feat(credits): ai3/step 02 wire purchased credits into core business logic

EmilFattakhov:
Teach the upload/download pipeline to consume purchased credits first (FIFO across rows, soonest-expiry first) before falling back to the free/one-off allocation.

### Interactions layer
- `interactionsRepository.createInteraction` now accepts a required `source: InteractionSource` ('free_tier' | 'purchased') and stores it in the new `interactions.source` column.
- `getInteractionsByAccountIdAndTypeInTimeRange` accepts an optional `source` filter so callers can query just free-tier rows.
- `InteractionsUseCases.createInteraction` forwards the source param.

### Accounts core
- `getPendingCreditsByAccountAndType` now passes `InteractionSource.FreeTier` to the interactions query, so only free-tier uploads/downloads reduce the free limit. Purchased-credit interactions are invisible to this check.

- `registerInteraction` (called on every upload/download) now:
  1. Attempts to consume the full size from `purchased_credits` via `purchasedCreditsRepository.consumeCredits()` (FIFO, row-locked).
  2. On success → one interaction row with source='purchased'.
  3. On `InsufficientPurchasedCreditsError` → drains whatever purchased bytes remain (creates source='purchased' interaction), then records the remainder against the free allocation (source='free_tier'). This handles the "3×10 MB purchased, 29 MB upload" scenario transparently across as many purchased-credit rows as needed.

- `addCreditsToAccount` is rewritten:
  - Enforces the per-user cap (`config.credits.maxBytesPerUser`, default 100 GiB) before inserting.
  - Creates a new `purchased_credits` row (one per purchase, with an independent `expiresAt` set to `config.credits.expiryDays` from now, default 90 days) instead of incrementing the flat account limits.
  - Signature change: `credits` is now `bigint` (was `number`); `intentId` is a new required third argument for the FK reference.

### Intents core
- `getIntentCredits` updated to return `bigint` (was `number`) to match the new `addCreditsToAccount` signature.
- `onConfirmedIntent` passes `intentId` as the third argument to `addCreditsToAccount`.

### DIFF
--- a/apps/backend/__tests__/unit/useCases/accounts.spec.ts
+++ b/apps/backend/__tests__/unit/useCases/accounts.spec.ts
@@ -1,5 +1,4 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-/* eslint-disable @typescript-eslint/no-explicit-any */
 import {
   jest,
   describe,


EmilFattakhov:
Took me ages to clean it up, going to squash the commits when merging.

PR #603, opened 2026-03-04
feat(pay-with-ai3): ai3/step 04 gate buyCredits on Google auth for public launch …

EmilFattakhov:
Changes the feature flag evaluation so the two operating modes are explicit:

  - Staff mode  (BUY_CREDITS_ACTIVE=false, BUY_CREDITS_STAFF_ONLY=true) Access granted via STAFF_DOMAINS / STAFF_USERNAME_ALLOWLIST config (currently @subspace.network and @autonomys.xyz).  No change to existing behaviour.

  - Public mode (BUY_CREDITS_ACTIVE=true) Access granted only to users authenticated via Google OAuth (oauthProvider === 'google').  All other providers — GitHub, Discord, web3-wallet, etc. — are blocked.  This prevents Sybil purchases without requiring a separate account-linking flow.

Additionally adds a defense-in-depth check on POST /intents that returns 403 GOOGLE_AUTH_REQUIRED if the active flag is set and the user is not on Google, ensuring a clear error even if the middleware is somehow bypassed.

Files changed:
- core/featureFlags/index.ts: add hasGoogleAuth(); update isActive()
- app/controllers/intents.ts: add explicit Google auth guard on POST /
- __tests__/unit/useCases/featureFlags.spec.ts: new test suite (14 cases) covering active/staffOnly/disabled modes and all provider variants

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,5 @@ environment.yaml
 
 # Next
 **/.next/
+# Claude Code worktrees
+.claude/worktrees/
--- a/apps/backend/__tests__/e2e/users/credits.spec.ts
+++ b/apps/backend/__tests__/e2e/users/credits.spec.ts
@@ -1,5 +1,6 @@
 import {
   InteractionType,
+  InteractionSource,
   AccountModel,
   IntentStatus,
   UserWithOrganization,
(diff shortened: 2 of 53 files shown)


jim-counter (approved):
LGTM. Like the strong test coverage.

PR #620, opened 2026-03-19
Feat/pay with ai3/step 09 credit history

EmilFattakhov:
## Summary

- Adds a backend credit guard to folder upload finalisation that was
  previously missing any account validation
- Introduces a full credit purchase history page at `/[chain]/drive/credits`
- Surfaces a "View history →" link in the sidebar purchased-credits widget

## Changes

**Backend — folder upload credit guard** (`fix(uploads)`)

`handleFolderUploadFinalization` had no credit check at all, allowing any
user — including those with a zero balance — to complete a folder DAG
structure without account validation.

A guard-only check is added that reads the pending credit balance and throws
if it has gone negative. `registerInteraction` is intentionally NOT called
here: each child file upload is individually finalised via
`handleFileUploadFinalization`, which already deducts the file's content
bytes. The folder root's `metadata.totalSize` equals the sum of its
children's sizes, so calling `registerInteraction(metadata.totalSize)` would
double-charge the user for all folder content. The commit includes a detailed
comment explaining the credit model.

**Frontend — `getCreditBatches()` API method**

New `getCreditBatches()` method on the API service calls
`GET /credits/batches`, returning the complete purchase history (all rows,
including expired, newest-first). Reuses the existing `ExpiringCreditBatch`
wire type — the serialisation shape is identical for both endpoints.

**Frontend — `/[chain]/drive/credits` history page**

New `CreditHistoryView` component and Next.js route, wrapped in
`UserProtectedLayout`. The view:

- Lists all purchased credit batches, newest-first
- Shows a status badge per batch: **Active** / **Expiring soon** /
  **Depleted** / **Expired**
- Renders a consumption progress bar per batch (bytes used / original
  purchase size with percentage)
- Shows a "Buy more credits" CTA when the user has no active batches or
  all non-expired batches expire within 7 days
- Guarded by `hasBuyCreditsFeature` (`features.buyCredits &&
  account.model === AccountModel.OneOff`) — non-qualifying users see a
  polite unavailable message; unauthenticated users are redirected by
  `UserProtectedLayout`

**Frontend — "View history →" sidebar link**

`AccountInformation` gains an optional `creditHistoryHref` prop. When
present and `purchasedBytesRemaining > 0`, a `Next.js Link` appears below
the balance/expiry line linking to the credits history page. `SideNavBar`
derives the href from `networkId` and passes it only when
`hasBuyCreditsFeature` is true. All existing callers are unaffected (the
prop is optional and defaults to `undefined`).

### DIFF
--- a/apps/frontend/src/components/molecules/AccountInformation/index.tsx
+++ b/apps/frontend/src/components/molecules/AccountInformation/index.tsx
@@ -68,9 +68,7 @@ export const AccountInformation = ({
         </p>
       )}
 
-      {/* Purchased credits — only rendered when the user has active purchased
-          credits (i.e. hasBuyCreditsFeature AND uploadBytesRemaining > 0).
-          Invisible to all other account types and feature-flag states. */}
+      {/* Purchased credits — rendered when the user has remaining balance. */}
       {hasPurchasedCredits && (
         <div className='border-t border-border pt-2'>
           <div className='text-xs text-muted-foreground'>Purchased credits</div>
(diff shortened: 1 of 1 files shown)


jim-counter (commented):
Just wanted to confirm this process - when we upload a folder we do an initial check that there is balance to upload the whole folder and then do a deduction and balance heck after each file upload?

PR #621, opened 2026-03-20
Feat/pay with ai3/step 10 payment robustness

EmilFattakhov:
## Summary

Production-hardening for pay-with-AI3: idempotency, edge-case payment fixes, orphaned transaction recovery, env var documentation, and an admin credits panel.

- **Idempotency on payment confirmation** — `markIntentAsConfirmed` now skips re-processing if the intent is already `CONFIRMED`, `COMPLETED`, or `OVER_CAP`, preventing double credit grants from duplicate on-chain events.
- **Dust payment guard** — `onConfirmedIntent` now rejects payments where `paymentAmount / shannonsPerByte` rounds to zero, rather than silently marking the intent `COMPLETED` with 0 bytes granted.
- **Orphaned transaction recovery** — on startup, the payment manager sweeps for `PENDING` intents with a `tx_hash` (submitted while the manager was down) and resumes watching each one via `Promise.allSettled`.
- **`.env.sample` updated** — all 11 pay-with-AI3 env vars documented with descriptions and defaults.
- **Admin credits panel** — new section in the admin dashboard showing: system-wide credit economics (expiring totals), all `OVER_CAP` intents with a Reprocess button, and a full purchase history table across all users (user, status, original/remaining bytes, usage bar, expiry date). Backed by a new `GET /credits/batches/all` admin endpoint.

### DIFF
--- a/apps/backend/.env.sample
+++ b/apps/backend/.env.sample
@@ -4,5 +4,60 @@ FILES_GATEWAY_URL=http://changeme.com # change to your own (or it'll fail when f
 FILES_GATEWAY_TOKEN=changeme # change to your own (or it'll fail when fetching archived files)
 AUTH_SERVICE_API_KEY=1234567890 # change to your own if updated in auth
 RABBITMQ_URL=amqp://guest:guest@localhost:5672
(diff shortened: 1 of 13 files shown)


jim-counter (dismissed):
LGTM

jim-counter (approved):

PR #646, opened 2026-03-31
fix(frontend): gate credits API calls behind buyCredits feature flag

EmilFattakhov:
The backend returns 404 for /credits/summary and /credits/batches/expiring when BUY_CREDITS_ACTIVE is disabled. SessionEnsurer was polling /credits/summary every 30 s and ExpiryWarningBanner was polling /credits/batches/expiring every 5 min for all authenticated users, flooding the console with 404 errors and adding unnecessary network overhead.

Both queries now use `enabled: !!session?.data && !!features?.buyCredits` so they only fire when the feature flag is actually on. ExpiryWarningBanner also gains the useUserStore import needed to read the flag.


EmilFattakhov:
Closing down as not needed

Issue #649, opened 2026-04-02
Rapidly uploading the same file produces new unretrievable CIDs


EmilFattakhov:
### Summary

Uploading the same file (identical name and content) in rapid succession produces a different CID on each upload, rather than deduplicating to the same CID. Each of these new CIDs cannot be downloaded — the gateway returns 404 and the API returns 500. Uploading the same file slowly (one at a time, with a gap between uploads) correctly deduplicates and returns the existing CID with no issue.

### Steps to Reproduce

1. Take any file — observed consistently with `.mjs` build artifacts (~180KiB, `text/javascript`)
2. Upload it multiple times in rapid succession (e.g. as part of a CI/CD pipeline uploading a bundle of 100+ files)
3. Observe that each upload returns a different CID despite identical content and filename
4. Attempt to download any of the new CIDs — receive `Failed to download file / Error: Failed to fetch file: 500`
5. Gateway returns: `{"error":"Failed to retrieve data","details":"Error fetching file header: 404 Not Found"}`
6. The file detail page shows: Status `Archived`, Archive blocks count `0`

By contrast, uploading the same file again after a pause correctly deduplicates — the system recognises the existing CID, shows the original entry, and no new broken CID is created.

### Root Cause (Suspected)

CID computation is deterministic given the same content, filename, and IPLD node structure. The fact that rapid uploads of identical files produce different CIDs means something non-deterministic is entering the IPLD node construction during concurrent processing.
The most likely candidate is a race condition in `migrateFromBlockstoreToNodesTable` (`apps/backend/src/core/objects/nodes.ts`): when two uploads of the same file are in the migration pipeline simultaneously, the second upload may read partially-committed blockstore state from the first, causing the DAG to be assembled with different node linking and therefore a different root CID.
Once a spurious CID is created, it ends up in a permanently broken state:

- The nodes are published on-chain but the DSN never properly indexes them (Archive blocks count stays 0)
- The file is prematurely marked `is_archived = true` by `checkObjectsArchivalStatus`
- `onObjectArchived` deletes `encoded_node` from the database
- The gateway returns 404 because it has no piece mappings for this CID
- There is no DB fallback — the file is unrecoverable without re-upload

### Affected Code

- `apps/backend/src/core/objects/nodes.ts — migrateFromBlockstoreToNodesTable()` : suspected race condition under concurrent uploads of identical files
[...]

PR #651, opened 2026-04-02
fix: prevent duplicate CID failures on rapid concurrent uploads of id…

EmilFattakhov:
…entical files

Three related issues caused uploads of the same file content (producing identical CIDs) to fail when uploaded rapidly/concurrently:

1. metadata.ts - setMetadata ON CONFLICT clause did not reset is_archived, so a re-upload of a previously archived file kept is_archived = true. This routed downloads immediately to the FileGateway fetcher (which had no piece mapping yet), returning 404/500. Fix: add is_archived = false to the ON CONFLICT DO UPDATE clause.

2. nodes.ts - saveNode and saveNodes used plain INSERT with no conflict handling. Concurrent migrations of the same CID caused a PRIMARY KEY violation on nodes.cid, leaving one upload's node data unwritten. Fix: add ON CONFLICT (cid) DO UPDATE with COALESCE so existing blockchain fields are preserved while the encoded_node is refreshed.

3. object.ts - onObjectArchived could mark an object as archived and delete its encoded_node even when no nodes existed in the table (e.g. after a failed migration). Once archived with no nodes, the file becomes permanently inaccessible. Fix: add a guard that skips archival when getNodesByRootCid returns empty.


EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot review

EmilFattakhov:
Resolves (potentially) https://github.com/autonomys/auto-drive/issues/649

PR #652, opened 2026-04-06
fix(frontend): pass explicit gasPrice to fix MetaMask fee estimation …

EmilFattakhov:
…on Auto EVM

Auto EVM is a Substrate-based network that does not implement EIP-1559 fee history (eth_feeHistory). MetaMask falls back to its own estimation which fails on custom networks, showing "Network fee: Unavailable" and blocking confirmation.

Fix: fetch the current gas price via eth_gasPrice (which the network does support) and add a 1 GWEI buffer before passing it explicitly to writeContractAsync. Supplying gasPrice forces a legacy (type-0) transaction, bypassing EIP-1559 estimation entirely — MetaMask receives a concrete fee it can display, and the tx is reliably included without the user having to manually adjust gas.


EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM

PR #655, opened 2026-04-07
feat(frontend): surface complete object metadata on details page

EmilFattakhov:
This PR fixes some incorrect, or hardcoded values on the file display page. 

- Show upload timestamp (createdAt) in Metadata card
- Show chunk count for files, total file count for folders
- Show publishedObjectId (truncated, full on hover) when available
- Replace hardcoded 'Archive blocks count: 0' with real archivedNodes
- Expand block range to show both min and max as explorer links
- Add null/undefined guards on every field so older or incomplete records never render broken output (fall back to 'N/A')
- Remove duplicate Uploaded/Total Nodes rows from Metadata card (they belong in Upload Status only)


EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM

PR #656, opened 2026-04-08
fix(backend): enable API key auth for credit purchase intents

jim-counter:
## Summary

**Root cause fix:** API key auth didn't resolve `oauthUsername` from the database, causing staff-only feature flag checks to fail.

**Additional fix:** `featureFlagMiddleware` silently swallowed auth errors via a catch-all, masking the real issue as a 404.

## Root cause

When `BUY_CREDITS_STAFF_ONLY=true`, the `featureFlagMiddleware('buyCredits')` calls `isStaff(user)`, which checks `user.oauthUsername` against staff domains and allowlisted usernames. This works for direct OAuth auth (Google, GitHub, Discord) because those providers return `{ provider, id, username, avatarUrl }`. But the API key provider in `apps/auth/src/services/authManager/providers/apikey.ts` only returned `{ provider, id }` — no `username`.

This `OAuthUser` flows into `getUserByOAuthUser()` which sets `oauthUsername: user.username` on the resolved user object. Since `user.username` was `undefined` from the API key provider, `isStaffDomain()` and `isStaffUsername()` both returned `false`, so `buyCredits` evaluated to `false`, and the middleware returned 404.

### 1. Fix API key auth to resolve full user info (root cause)
**`apps/auth/src/services/authManager/providers/apikey.ts`**

The API key provider now looks up the full user row from the DB via `usersRepository.getUserByOAuthInformation()` to populate `username` and `avatarUrl`, matching the behaviour of all other auth providers.

### 2. Fix `featureFlagMiddleware` error handling
**`apps/backend/src/core/featureFlags/express.ts`**

The middleware delegated to `getFeatureFlags()` which wraps `handleAuth` in a try/catch that silently falls back to unauthenticated flags on any error. This meant auth failures surfaced as a misleading 404 instead of the actual error. The middleware now calls `handleAuth` directly and lets auth errors propagate. Added debug logging of `oauthProvider` when a flag evaluates to false.

### 3. Improve error messages
**`apps/backend/src/app/controllers/intents.ts`**

`GOOGLE_AUTH_REQUIRED` → `GOOGLE_ACCOUNT_REQUIRED` with message clarifying that API keys from Google-registered accounts are also accepted.

### 4. Add OpenAPI docs for intent endpoints
**`apps/backend/src/docs/reference/intents.ts`** (new) + **`apps/backend/src/docs/reference/index.ts`**

Documents `POST /intents`, `GET /intents/:id`, `POST /intents/:id/watch`, `GET /intents/price` with a third-party integration guide describing the full Pay with AI3 flow via API key.

### 5. Add tests
**`apps/backend/__tests__/unit/useCases/featureFlags.spec.ts`**

Tests covering API key auth through feature flags, including:
- `hasGoogleAuth` with API key-resolved users
- `buyCredits` active mode with API key auth
- `buyCredits` staff-only mode with API key user that has `oauthUsername` (the fixed path)
- `buyCredits` staff-only mode with undefined `oauthUsername` (the pre-fix broken path)

## Test plan

- [x] `yarn backend lint` passes
- [x] `yarn auth lint` passes
- [x] All 31 feature flag tests pass (including 7 new)
- [x] All 41 intent tests pass
- [ ] Deploy and verify `POST /intents` works with API key auth from a Google-registered staff account
- [ ] Verify `GET /api/features` returns `{"buyCredits": true}` for the same API key

### DIFF
--- a/apps/auth/src/services/authManager/providers/apikey.ts
+++ b/apps/auth/src/services/authManager/providers/apikey.ts
@@ -1,12 +1,24 @@
 import { OAuthUser } from '@auto-drive/models'
 import { ApiKeysUseCases } from '../../../useCases/apikeys.js'
+import { usersRepository } from '../../../repositories/users.js'
 
 const getUserFromApiKey = async (secret: string): Promise<OAuthUser> => {
   const apiKey = await ApiKeysUseCases.getApiKeyFromSecret(secret)
 
+  // Look up the full user row so that oauthUsername and oauthAvatarUrl are
+  // available downstream.  Without this, feature flag checks that inspect
+  // oauthUsername (e.g. staff domain / allowlist checks for buyCredits)
+  // fail because the username is undefined.
+  const dbUser = await usersRepository.getUserByOAuthInformation(
+    apiKey.oauthProvider,
+    apiKey.oauthUserId,
+  )
+
   return {
     provider: apiKey.oauthProvider,
     id: apiKey.oauthUserId,
+    username: dbUser?.oauth_username ?? undefined,
+    avatarUrl: dbUser?.oauth_avatar_url ?? undefined,
   }
 }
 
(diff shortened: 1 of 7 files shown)


jim-counter:
bugbot run

EmilFattakhov (approved):
Looks good!

PR #657, opened 2026-04-09
feat(backend): expose contract info via GET /intents/contract

EmilFattakhov:
Third-party integrators need the contract address, chain ID, and ABI to construct the on-chain payIntent transaction. This was previously only available in frontend source code.


EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM!

PR #660, opened 2026-04-14
fix: authenticated downloads fail due to credit enforcement and broken retry logic + disable download limiting

EmilFattakhov:
## Problem

The backend code comment in `downloadObjectByUser` states:

> "Download credits are not enforced right now — infrastructure exists for future use but purchased download bytes are not allocated on purchase and are not counted here."

However, `downloadObjectByUser` was still checking `pendingCredits < metadata.totalSize` and returning a `402 PaymentRequiredError` when the user's free download quota was exhausted. Since purchased download bytes are never allocated on purchase, users have no way to replenish a depleted quota — making the block permanent.

Verified via browser: authenticated requests to `public.auto-drive.autonomys.xyz/api/downloads/{cid}` returned `{"error":"Not enough download credits"}` because `pendingDownloadCredits` was deeply negative, despite the design intent being that downloads are unrestricted for authenticated users.

A second independent bug prevented decrypted (password-protected) downloads from recovering even after the credit block was lifted: the session-auth retry in `fetchFromApi` never fired for the decrypted path. The SDK's `downloadFile` wraps all HTTP failures as `new Error('Failed to download file: ' + response.statusText)`. On HTTP/2, `statusText` is always `""`, producing `"Failed to download file: "` — a message that `isAnonymousTooLargeError` in `download.ts` does not match (it looks for "payment required", "file too large", etc.), so the anonymous→session retry was silently skipped.

### Backend (`apps/backend`)
- **Commented out** the `pendingCredits` check in `downloadObjectByUser` so authenticated users are never blocked by download quota. The comment explains why enforcement is disabled and what must be wired up before re-enabling it. Download interactions are still registered via `registerInteraction` for future analytics/billing.
- **Removed `PaymentRequiredError`** from the `downloadObjectByUser` return type (it could no longer be returned after the check was disabled, flagged as a separate bugbot finding).

### Frontend (`apps/frontend`)
- **Fixed the session-auth retry** for decrypted downloads: `downloadObject` now wraps `api.downloadFile` in a try-catch and re-throws with the recognised login message when the error is exactly `"Failed to download file: "` (the HTTP/2 empty-statusText form) and `authMode` is `'anonymous'`. Errors with a populated `statusText` (HTTP/1.1 404, 500, etc.) propagate unchanged so `isAnonymousTooLargeError` in `download.ts` can still discriminate them correctly.
- **Added `moduleNameMapper`** to `jest.config.ts` so Jest resolves `tsconfig` `baseUrl` path aliases (`utils/*`, `services/*`, etc.).
- **Added 13 unit tests** for `createApiService().downloadObject` covering:
  - Encrypted raw download (`skipDecryption: true`): anonymous 402/404 error handling, session success
  - Decrypted download (`skipDecryption: false`): HTTP/2 empty-statusText bug regression, HTTP/1.1 populated-statusText errors propagate unchanged, unrelated errors propagate, session success, password forwarding, immediate guard when no session is available

## Test plan
- [ ] Download a file (encrypted) as an anonymous user — should work for small files, prompt login for large ones
- [ ] Download a file (decrypted, with password) as an anonymous user hitting a large file — should now prompt login instead of silently failing
- [ ] Download a file as an authenticated user whose free download quota is exhausted — should now succeed instead of returning 402
- [x] Run `yarn frontend test` — all 100 tests pass

### DIFF
--- a/apps/frontend/__tests__/unit/services/downloadObject.spec.ts
+++ b/apps/frontend/__tests__/unit/services/downloadObject.spec.ts
@@ -1,0 +1,327 @@
+/**
+ * Unit tests for `createApiService().downloadObject`
+ *
+ * These tests cover all four download scenarios:
+ *   1. Encrypted raw download (skipDecryption=true) — anonymous 402 path
+ *   2. Encrypted raw download (skipDecryption=true) — session success path
+ *   3. Decrypted download (skipDecryption=false) — anonymous, including the
+ *      HTTP/2 empty-statusText bug regression
+ *   4. Decrypted download (skipDecryption=false) — session
+ *
+ * External dependencies are fully mocked so no network or Next.js runtime is
+ * required.
+ */
+
+// ---------------------------------------------------------------------------
+// Module mocks — must be declared before any imports so Jest hoists them
+// ---------------------------------------------------------------------------
+
+jest.mock('@autonomys/auto-drive', () => ({
+  createAutoDriveApi: jest.fn(),
+}));
+
+jest.mock('utils/auth', () => ({
+  getAuthSession: jest.fn(),
+}));
+
+// uploadFileContent is imported by api.ts but never used in downloadObject.
+jest.mock('utils/file', () => ({
+  uploadFileContent: jest.fn(),
+}));
+
+// asyncFromStream is dynamically imported inside the skipDecryption=true path.
+jest.mock('@autonomys/asynchronous', () => ({
+  asyncFromStream: jest.fn(),
+}));
+
+// ---------------------------------------------------------------------------
+// Imports (after mocks)
+// ---------------------------------------------------------------------------
+
+import { createApiService } from '../../../src/services/api';
+import { createAutoDriveApi } from '@autonomys/auto-drive';
+import { getAuthSession } from 'utils/auth';
+import { asyncFromStream } from '@autonomys/asynchronous';
+
+// ---------------------------------------------------------------------------
+// Helpers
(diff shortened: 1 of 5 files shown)


EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

apps/frontend/__tests__/unit/services/downloadObject.spec.ts
```diff
@@ -0,0 +1,336 @@
...
+describe('downloadObject — encrypted raw download (skipDecryption: true)', () => {
+  beforeEach(() => jest.clearAllMocks());
+
+  it('anonymous — backend 402 throws the recognised "large file" login message', async () => {
+    (getAuthSession as jest.Mock).mockResolvedValue(null);
+
+    mockAutoDriveApi({
+      sendDownloadRequest: jest.fn().mockResolvedValue({
+        ok: false,
+        status: 402,
```
jim-counter, line 86: 402 is not entirely accurate. 401 or 413 would be a better fit but we can leave this for now as we intend to revisit paying for downloads when the time comes.

jim-counter (approved):
LGTM. This explains why the 300MB file would not download for me.

apps/frontend/__tests__/unit/services/downloadObject.spec.ts
EmilFattakhov: Right, agree

PR #662, opened 2026-04-16
fix(frontend): check response.status instead of statusText for downlo…

EmilFattakhov:
…ad errors

The skipDecryption=false path relied on the SDK's downloadFile, which throws with response.statusText — always empty on HTTP/2. This made 404, 500, and other errors indistinguishable from 402, causing all anonymous download failures to be rewritten as the "please login" message and triggering unnecessary session-auth retries.

Replace the SDK's downloadFile call with direct sendDownloadRequest + manual decryption/decompression, using numeric response.status checks that work identically on HTTP/1.1 and HTTP/2.

Made-with: Cursor

### DIFF
--- a/apps/frontend/__tests__/unit/services/downloadObject.spec.ts
+++ b/apps/frontend/__tests__/unit/services/downloadObject.spec.ts
@@ -1,15 +1,13 @@
 /**
  * Unit tests for `createApiService().downloadObject`
  *
- * These tests cover all four download scenarios:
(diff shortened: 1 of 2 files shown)


EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
This PR is addresses the issue discovered by bugbot here: https://github.com/autonomys/auto-drive/pull/661#discussion_r3093685323

jim-counter (approved):
LGTM

PR #663, opened 2026-04-21
fix(backend): batch publish-nodes tasks to prevent RabbitMQ queue bloat and worker OOM

EmilFattakhov:
## Problem

`scheduleNodesPublish()` published **one RabbitMQ message per IPLD node CID**. A single multi-chunk upload produces hundreds of nodes; a bulk upload of thousands of files produced 38k+ messages in the `task-manager` queue. On worker restart, RabbitMQ delivered 50 messages simultaneously (prefetch=50), each triggering a `publishNodes` call that loads the Polkadot API and chain metadata. Combined startup memory exceeded the Node.js heap limit (~1.7 GB), causing the worker to crash-loop every ~18 seconds with `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`.

## Fix

Batch all CIDs for a given object into groups of up to 50 per message instead of one message per CID (`PUBLISH_BATCH_SIZE = 50`):

- **Before**: 200-node file → 200 messages
- **After**: 200-node file → 4 messages

No schema changes needed — `publish-nodes` already accepts `nodes: string[]` and `OnchainPublisher.publishNodes` already handles arrays with idempotent retry semantics (already-published CIDs are skipped on retry via `getNodesBlockchainDataBatch`).

## Scope

**This PR only prevents the problem from recurring on future uploads.** The existing ~38k single-CID messages already in the queue are unaffected — they must be drained separately (purge via RabbitMQ management UI, or lower `RABBITMQ_PREFETCH` to 5–10 and let them drain as no-ops).

## Files changed

| File | Change |
|---|---|
| `apps/backend/src/core/uploads/uploads.ts` | Batch CIDs in `scheduleNodesPublish` |
| `apps/backend/__tests__/unit/useCases/uploads.spec.ts` | 6 new unit tests |

## Tests

6 unit tests added for `scheduleNodesPublish`:
- Nodes < 50 → 1 message with all CIDs
- Nodes == 50 → 1 message
- Nodes == 120 → 3 messages (50 + 50 + 20)
- Empty nodes → 0 messages
- CID ordering preserved across batches
- Correct root CID forwarded to `getCidsByRootCid`

### DIFF
--- a/apps/backend/__tests__/e2e/uploads/folder.spec.ts
+++ b/apps/backend/__tests__/e2e/uploads/folder.spec.ts
@@ -289,33 +289,14 @@ describe('Folder Upload', () => {
         tags: ['insecure'],
       })
 
-      const cids = [folderCID, subfileCID, subfolderCid]
+      // All nodes for the root CID are batched into a single publish-nodes message
+      const publishNodesCalls = rabbitMock.mock.calls.filter(
+        (call) => call[0] === 'task-manager' && (call[1] as { id: string }).id === 'publish-nodes',
+      )
+      expect(publishNodesCalls).toHaveLength(1)
 
-      cids.forEach((cid) => {
-        expect(rabbitMock).toHaveBeenCalledWith('task-manager', {
-          id: 'publish-nodes',
-          params: {
-            nodes: [cid],
-          },
-          retriesLeft: expect.any(Number),
-        })
-      })
-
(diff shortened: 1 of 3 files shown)


EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM

PR #666, opened 2026-04-24
feat(api-keys): add name, optional expiry

EmilFattakhov:
## DB migration (`apps/auth/migrations/20260424120000-api-keys-name-expiry`)

Pure-additive, no backfill, no data mutation:

- `ALTER TABLE users.api_keys ADD COLUMN name TEXT NULL` — optional label.
- `ALTER TABLE users.api_keys ADD COLUMN expires_at TIMESTAMP NULL` — `NULL` = never expires.
- Partial index `idx_api_keys_owner (oauth_provider, oauth_user_id) WHERE deleted_at IS NULL` for the listing query.
- The `secret` column is untouched.

Pre-existing keys come out with `name = NULL` and `expires_at = NULL`. No placeholder strings.

## Model changes (`@auto-drive/models`)

- `ApiKey` gains `name: string | null`, `expiresAt: Date | null`, `createdAt: Date | null`. `secret: string` stays where it was.
- `ApiKeyWithoutSecret = Omit<ApiKey, 'secret'> & { prefix: string }` — the safe listing shape. `prefix` is computed at response time as the first 8 chars of the secret; it is never stored.
- `CreateApiKeyInput = { name?: string | null; expiresAt?: string | null }` — both fields optional.
- No separate `CreatedApiKey` type: the create endpoint returns `ApiKey` directly, which already contains the plaintext.

## Auth service

- `apiKeysRepository` reads/writes `secret` in plaintext, plus the new `name` / `expires_at` columns. Lookup is `WHERE secret = $1`. `createApiKey` takes `{ id, name, secret, oauthProvider, oauthUserId, expiresAt }`.
- `ApiKeysUseCases.createApiKey`:
  - validates `name` — null/undefined/whitespace-only → stored as `NULL`; max length 64 (error otherwise).
  - validates `expiresAt` — must be in the future if set; `null`/omitted → never expires.
  - generates a random secret (uuid v4, no dashes) and inserts.
- `getApiKeyFromSecret` rejects deleted keys with `"Api key has been deleted"` and expired keys with `"Api key has expired"`.
- `getApiKeysByUser` strips the secret and computes `prefix` on the way out.
- Controller: `POST /users/@me/apiKeys/create` takes a JSON body `{ name?, expiresAt? }` (or no body at all) and returns the full `ApiKey` including `secret`. `DELETE /users/@me/apiKeys/:id` unchanged.

## Frontend

- **Creation modal** (`ApiKeyCreationModal`): name input (labelled "Name (optional)") + expiry `<select>` with presets (Never / 30d / 60d / 90d / 1y / Custom date). Submitting with an empty name sends `null` to the API. On success, shows the secret once with a Copy button. Modal won't dismiss on backdrop click until the user has hit Copy.
- **API keys table** (`ApiKeysTable`): columns Name · Key (prefix + `…`) · Created · Expires · Actions. Nameless keys render as `—`. Past-expiry rows show the date in red with an `(expired)` tag. Per-row **Delete** button; no Rotate.

## Test coverage — `apps/auth/__tests__/apikeys.spec.ts`

- creating a key with a name returns the expected shape (`{ id, secret, name, oauthProvider, oauthUserId, expiresAt: null }`)
- creating a key with no `name` field at all succeeds with `name: null`
- whitespace-only names are coerced to `null`
- names longer than 64 chars are rejected at create time
- past-dated `expiresAt` is rejected at create time
- listing exposes `prefix` but never leaks `secret`
- auth succeeds with a valid key and returns the right provider/user
- keys past their `expiresAt` fail auth with `"Api key has expired"`
- soft-deleted keys fail auth with `"Api key has been deleted"`
- unknown secret fails auth with `"Api key not found"`

## Tradeoffs

Earlier drafts of this PR also hashed secrets at rest and added a Rotate flow. Both were dropped:

- **Hashing**: storing only the hash is a real security win (DB dumps / read-only leaks no longer leak working credentials), but it forces a non-reversible up migration and spreads complexity through the repo, use cases, tests, and UI. Given the threat model hasn't changed in this PR, the payoff didn't justify the migration risk right now.
- **Rotation**: without hashing, a rotate endpoint is just "generate a new secret, overwrite the old one." Users can already get that outcome by deleting the key and creating a new one, so the extra endpoint + modal didn't earn its keep.

Both remain easy follow-ups: add `secret_hash` alongside `secret` in a future expand-only migration, backfill, switch the auth read path, then drop `secret` in a contract step. Rotation becomes ~10 lines of endpoint on top.

## Backwards compatibility

- Pre-existing keys keep authenticating — the `secret` column was never touched.
- Pre-existing rows have `name = NULL` and `expires_at = NULL`; the UI renders nameless keys as `—`.
- `GET /users/@me/apiKeys` gains `name`, `prefix`, `expiresAt`, `createdAt`; no fields removed. Existing consumers that only read `id` / `prefix` / etc. are unaffected.

## Out of scope / follow-ups

- Hashing at rest + rotation (see Tradeoffs).
- No rename endpoint — users who want to change a name drop and recreate.
- No background job to purge long-expired keys; they just fail auth.
- No rate limiting on auth attempts — same as today.

### DIFF
--- a/apps/auth/__tests__/apikeys.spec.ts
+++ b/apps/auth/__tests__/apikeys.spec.ts
@@ -1,15 +1,15 @@
-import { ApiKey, User } from '@auto-drive/models'
-import { UsersUseCases, ApiKeysUseCases } from '../src/useCases/index.js'
+import { APIKey, User } from '@auto-drive/models'
+import { UsersUseCases, APIKeysUseCases } from '../src/useCases/index.js'
 import { PreconditionError } from './utils/error.js'
 import { closeDatabase, getDatabase } from '../src/drivers/pg.js'
 import { apiKeysRepository } from '../src/repositories/index.js'
-import { ApiKeyAuth } from '../src/services/authManager/providers/apikey.js'
+import { APIKeyAuth } from '../src/services/authManager/providers/apikey.js'
 import { dbMigration } from './utils/dbMigrate.js'
 import { MOCK_UNONBOARDED_USER } from './utils/mocks.js'
 
-describe('ApiKeyUseCases', () => {
+describe('APIKeyUseCases', () => {
   let user: User
-  let apiKey: ApiKey
+  let apiKey: APIKey
 
   beforeAll(async () => {
     await getDatabase()
@@ -21,8 +21,8 @@ describe('ApiKeyUseCases', () => {
     user = result
   })
 
-  it('should create an api key with a name and no expiry', async () => {
-    apiKey = await ApiKeysUseCases.createApiKey(user, { name: 'My key' })
+  it('should create an API key with a name and no expiry', async () => {
+    apiKey = await APIKeysUseCases.createAPIKey(user, { name: 'My key' })
     expect(apiKey).toMatchObject({
       id: expect.any(String),
       secret: expect.any(String),
@@ -34,27 +34,27 @@ describe('ApiKeyUseCases', () => {
     expect(apiKey.secret.length).toBeGreaterThan(0)
   })
 
-  it('should create an unnamed api key when no name is supplied', async () => {
-    const nameless = await ApiKeysUseCases.createApiKey(user, {})
+  it('should create an unnamed API key when no name is supplied', async () => {
+    const nameless = await APIKeysUseCases.createAPIKey(user, {})
     expect(nameless.name).toBeNull()
-    await ApiKeysUseCases.deleteApiKey(user, nameless.id)
+    await APIKeysUseCases.deleteAPIKey(user, nameless.id)
   })
 
   it('should coerce whitespace-only names to null', async () => {
-    const blank = await ApiKeysUseCases.createApiKey(user, { name: '   ' })
+    const blank = await APIKeysUseCases.createAPIKey(user, { name: '   ' })
     expect(blank.name).toBeNull()
-    await ApiKeysUseCases.deleteApiKey(user, blank.id)
+    await APIKeysUseCases.deleteAPIKey(user, blank.id)
   })
 
   it('should reject names longer than 64 characters', async () => {
     await expect(
-      ApiKeysUseCases.createApiKey(user, { name: 'x'.repeat(65) }),
+      APIKeysUseCases.createAPIKey(user, { name: 'x'.repeat(65) }),
     ).rejects.toThrow('API key name must be 64 characters or fewer')
   })
 
   it('should reject an expiry in the past', async () => {
     await expect(
-      ApiKeysUseCases.createApiKey(user, {
+      APIKeysUseCases.createAPIKey(user, {
         name: 'Backdated',
         expiresAt: new Date(Date.now() - 1000),
       }),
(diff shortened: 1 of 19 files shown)


EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

apps/auth/__tests__/apikeys.spec.ts
```diff
@@ -39,6 +82,18 @@ describe('ApiKeyUseCases', () => {
...
+  it('should reject expired keys at auth time', async () => {
+    const soon = new Date(Date.now() + 2_000)
+    const shortLived = await ApiKeysUseCases.createApiKey(user, {
+      name: 'Short-lived',
+      expiresAt: soon,
+    })
+    await new Promise((r) => setTimeout(r, 2_100))
+    await expect(
+      ApiKeyAuth.getUserFromApiKey(shortLived.secret),
+    ).rejects.toThrow('Api key has expired')
```
jim-counter: Small thing but it would be good to replace all `Api` with `API` for consistency.

I see the comment below about comparison breaking but a better fix is to update in the comparison or make it case-insensitive if there are concerns about breaking code outside of this change.

docker-compose.devProdAuth.yml
```diff
@@ -4,7 +4,7 @@ volumes:
 
 services:
   postgres:
-    image: postgres:latest
+    image: postgres:17
```
jim-counter: Just checking this was intentional. I know that this means the stack will actually come up for devs rather than fail thanks to the way Postgres 18 organises the data folder.

jim-counter (dismissed):
Code looks good to me. One niggle about case on API - will leave to your discretion on whether to address. And one quick check on the Postgres version change.

docker-compose.devProdAuth.yml
EmilFattakhov: Not intentional, dev artifact, reverting 👍

apps/auth/__tests__/apikeys.spec.ts
EmilFattakhov: I agree, that would have been a better approach, replacing all "Api" references to "API"

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
Feedback addressed Jim, ready for another review 👍

EmilFattakhov:
bugbot run

EmilFattakhov:
All right, I think that should be it @jim-counter!

jim-counter (approved):
LGTM. Thanks for the changes.

PR #667, opened 2026-04-27
Feat/buy credits all account models

EmilFattakhov:
### Enable Pay with AI3 for all account models

**What**
Removes the AccountModel.OneOff gate from three frontend components so Monthly subscribers can also buy credits when the buyCredits feature flag is enabled.

**Why**
The backend already supports purchased credits for any account model — the restriction was UI-only.

**Changes**
SideNavBar, UploadingFolderModal, CreditHistoryView — dropped account.model === AccountModel.OneOff check; buyCredits feature flag is now the sole gate
E2E tests — added Google-auth gate tests for POST /intents across all OAuth providers and account models, plus a full purchase-consume flow test for Monthly accounts
docs/payments.md — corrected misleading error description
No backend changes needed
The buyCredits flag already enforces Google-auth and rollout state in core/featureFlags.


EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

apps/frontend/src/components/views/CreditHistory/index.tsx
```diff
@@ -130,8 +129,10 @@ export const CreditHistoryView = () => {
   const { api, network } = useNetwork();
   const { account, features } = useUserStore();
 
-  const hasBuyCreditsFeature =
-    !!features.buyCredits && account?.model === AccountModel.OneOff;
+  // Guard on `account` so we never fire API calls for logged-out users (the
+  // store persists `features` to localStorage and `clearUser` may not have
+  // run yet).
+  const hasBuyCreditsFeature = !!features.buyCredits && account !== null;
```
jim-counter, line 135: Why do we `!!` `features.buyCredits` in this one place and not the others?

EmilFattakhov:
bugbot run

apps/frontend/src/components/views/CreditHistory/index.tsx
EmilFattakhov: I was going to respond in my voice, but more elaborate response from claude suits better :) 

The !! coerces features.buyCredits to a strict boolean. features is typed as Record<string, boolean>, so features.buyCredits is boolean | undefined (the key may not exist). Without !!, the expression could be undefined && account !== null, which evaluates to undefined — not false.

Since hasBuyCreditsFeature is passed to enabled: on the useQuery call (line 142), React Query expects a boolean. Passing undefined there has different semantics (it means "don't manage enabled/disabled at all"), so the !! ensures it's always true or false.

The other two components (SideNavBar, UploadingFolderModal) skip the !! because they only use hasBuyCreditsFeature in JSX conditionals where truthiness is sufficient.

jim-counter: Makes sense!

EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM!

PR #669, opened 2026-04-29
feat(frontend): admin-only Cached/Not cached tag on file detail page

EmilFattakhov:
## Summary

Adds an admin-only cache status indicator on the file detail page so admins can see at a glance whether a file is currently in the download cache.

- Show a green "Cached" or gray "Not cached" pill next to the existing file tags (Insecure / On review / Banned), visible only to admins via `RoleProtected`
- Lift `useFileInCache` into the parent `ObjectDetails` component so both the tag and the "Bring to Cache" button share a single API call instead of making two
- Widen `useFileInCache` return type from `boolean` to `boolean | null` (`null` = loading), which prevents the "Bring to Cache" button from flashing on initial render
- Clean up dead code: removed an unreachable outer `try/catch` in the hook (the `.catch(() => false)` on the promise already handles errors)

### What didn't change

- The "Bring to Cache" button behavior is unchanged — it already renders for any authenticated user when `isCached === false`, and the backend endpoint (`POST /downloads/async/:cid`) has no ownership restriction
- No backend changes

### DIFF
--- a/apps/frontend/src/components/organisms/ObjectDetails/ObjectDetailsActions.tsx
+++ b/apps/frontend/src/components/organisms/ObjectDetails/ObjectDetailsActions.tsx
@@ -1,4 +1,4 @@
-import { DownloadStatus, ObjectInformation, ObjectTag } from '@auto-drive/models';
+import { ObjectInformation, ObjectTag } from '@auto-drive/models';
 import { Button } from '@auto-drive/ui';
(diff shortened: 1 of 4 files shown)


EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM

PR #671, opened 2026-05-01
fix(frontend): only disable close guard after successful clipboard copy

EmilFattakhov:
setHasBeenCopied(true) was placed after the try/catch in copyApiKey, so it ran even when navigator.clipboard.writeText threw. This disabled the dialog's close-prevention guard, letting the user dismiss the modal before successfully copying the secret — which is never shown again.


EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM

PR #676, opened 2026-05-04
fix: add timeouts to gateway fetches, safe archival, fallback fetcher…

EmilFattakhov:
**Backend resilience:**
- Add `withTimeout` utility; wrap `FileGateway.getFile()`/`getNode()` calls (default 60s, covers initial metadata fetch)
- Add inactivity-based timeout for async download streams (default 5min) — resets on every data chunk so large files stream indefinitely, but hung connections with no progress get killed and marked as failed
- Async downloads catch errors at both the stream creation and streaming phases, marking downloads as `failed` with descriptive messages instead of hanging at `pending` forever
- Stream event handlers use fire-and-forget DB updates with `.catch()` to prevent unhandled promise rejections
- Sync downloads return HTTP 504 with a clear message on `TimeoutError` instead of crashing the process
- Add `taskTimeoutMs` option to `createHandlerWithRetries` (opt-in, not enabled for download tasks)
- CORS: mount permissive `cors()` fallback when `CORS_ALLOWED_ORIGINS` is unset

**Frontend UX:**
- When a logged-in user clicks Download on a non-cached file, the modal now checks cache status first and routes through the async download path — shows a "Preparing" UI, polls every 10s, and auto-triggers the download when the file is ready
- User can close the dialog and continue browsing; the download starts automatically via toast notification
- Falls back to sync download for anonymous users or if async creation fails
- Polling interval cleaned up on component unmount and CID change
- File preview now has a 15s timeout — non-cached files that can't be served quickly silently hide the preview instead of showing "Failed to fetch"

New env vars (with defaults):
- `FILES_GATEWAY_FETCH_TIMEOUT_MS=60000`
- `DOWNLOAD_INACTIVITY_TIMEOUT_MS=300000`


EmilFattakhov:
bugbot run

apps/backend/src/core/objects/files/index.ts
```diff
@@ -8,11 +8,55 @@ import {
...
+    async fetchFile(cid: string): Promise<Readable> {
+      try {
+        return await FileGatewayObjectFetcher.fetchFile(cid)
+      } catch (error) {
+        logger.warn(
+          'FileGateway fetch failed for cid=%s, falling back to DB fetcher: %s',
+          cid,
+          error instanceof Error ? error.message : String(error),
+        )
+        return DBObjectFetcher.fetchFile(cid)
```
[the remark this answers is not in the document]
EmilFattakhov: No longer relevant, skipped

apps/backend/src/core/downloads/async.ts
```diff
@@ -143,40 +144,119 @@ const asyncDownload = async (
...
+      downloadId,
+      download.cid,
+      message,
+    )
+    await AsyncDownloadsUseCases.setError(downloadId, message)
+    return err(new InternalError('Failed to start download'))
+  }
 
   let downloadedBytes = 0n
+  const inactivityMs = config.params.downloadInactivityTimeoutMs
```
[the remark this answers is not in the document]
EmilFattakhov: The issue does not exist in the current codebase. `downloadInactivityTimeoutMs` **is** defined in `config.params`:

```123:125:apps/backend/src/config.ts
    downloadInactivityTimeoutMs: Number(
      env('DOWNLOAD_INACTIVITY_TIMEOUT_MS', '300000'),
    ),
```

And it's consumed correctly in `async.ts`:

```169:169:apps/backend/src/core/downloads/async.ts
  const inactivityMs = config.params.downloadInactivityTimeoutMs
```

The property name matches exactly, and it defaults to `300000` (5 minutes). The inactivity timer in `async.ts` will use that 300-second delay as intended — it won't fire immediately after the first chunk. There's nothing to fix here.

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

apps/frontend/src/components/molecules/FilePreview/index.tsx
```diff
@@ -83,13 +84,23 @@ export const FilePreview = ({ metadata }: { metadata: OffchainMetadata }) => {
...
+      );
+
       setIsFilePreview(true);
       setLoading(true);
 
       try {
         const response = await fetch(gatewayUrl, {
           signal: controller.signal,
         });
+        clearTimeout(timeoutId);
```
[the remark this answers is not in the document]
EmilFattakhov: The issue as described is not present in the current code, doesn't require addressing.

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

apps/backend/src/app/apis/download.ts
```diff
@@ -47,7 +47,8 @@ const createServer = async () => {
       }),
     )
   } else {
-    logger.warn('CORS is not configured - no allowed origins specified')
+    logger.warn('CORS is not configured - no allowed origins specified, allowing all origins')
+    app.use(cors())
```
jim-counter: This gives me an uneasy feeling. How do we improve things by making this change?

apps/frontend/src/components/organisms/UserAsyncDownloads/state.tsx
```diff
@@ -1,17 +1,28 @@
 import { AsyncDownload } from '@auto-drive/models';
 import { create } from 'zustand';
 
+export interface PendingAutoDownload {
+  cid: string;
+  password?: string;
```
jim-counter, line 6: Is this plaintext?

jim-counter (commented):
Couple of questions.

apps/backend/src/app/apis/download.ts
EmilFattakhov: Good catch, that was one of the bugbot review suggestions that I assumed to be true, my bad. This is definitely a bad idea.

apps/frontend/src/components/organisms/UserAsyncDownloads/state.tsx
EmilFattakhov: Yes. The practical risk is low here — the password already lives in the useEncryptionStore Zustand store and in the ObjectDownloadModal component state during the download flow anyway.

That said, upon additional review a better approach was identified, going ahead with it.

EmilFattakhov: Okay, after a few iterations two agents agreed that this approach should remain as-is, who am I tell them otherwise 😄 

> if we don't snapshot it here when they close the modal, we lose it and the background auto-download will fail to decrypt. Worth noting that useEncryptionStore already persists the default password to localStorage in plaintext, so this in-memory-only copy is strictly less exposed.

jim-counter: Sounds like we're not extending any attack surface. Makes sense.

jim-counter (approved):
LGTM

Issue #674, opened 2026-05-04
Objects permanently stuck in "Archiving" status due to object-mapping-indexer missing block events


EmilFattakhov:
## Summary

Approximately **8,991 objects** containing **14,011 IPLD nodes** are permanently stuck in **"Archiving"** status on mainnet. These objects have all their nodes successfully published on-chain (they have valid `block_published_on` and `tx_published_on`), but are missing `piece_index` and `piece_offset` values — meaning the object-mapping-indexer never recorded the DSN piece assignments for them. The existing recovery mechanism cannot fix these objects because it only replays mappings **forward** from the current high-water mark.

## How the Issue Was Found

While investigating the status of a specific file (`bafkr6igscduprydpytouaexptllq2b24v3ps2lbmnlwg5byjo54maqb2qe`, "Taeko Ohnuki.zip"), the following was observed:

1. **Metadata**: The file exists in `public.metadata` with `is_archived = false`, created 2026-04-17.
2. **Nodes**: 1,688 total nodes, all 1,688 published on-chain, but only 1,648 have `piece_index` set — **40 FileChunk nodes are missing archival data**.
3. **All 40 stuck nodes** were published at the same block: **7,458,231**.

### Confirming the Root Cause

**Step 1: Verified the indexer has no data for the affected block**

Querying the object-mapping-indexer HTTP API:

```
GET https://indexer.mainnet.autonomys.xyz/objects/by-block/7458231
Response: []    ← empty, no mappings indexed for this block
```

Compare to a block where sibling nodes were successfully archived:

```
GET https://indexer.mainnet.autonomys.xyz/objects/by-block/7458212
Response: [["0f52f848...", 258918, 202192], ["39c21fb9...", 258918, 267386], ...]   ← data present
```

**Step 2: Verified the indexer has no record of the individual hashes**

Extracted blake3 hashes from the stuck CIDs (via `CID.parse()` → `multihash.digest`) and queried:

```
GET https://indexer.mainnet.autonomys.xyz/objects/<hash>
Response: Internal Server Error    ← hash not found in indexer DB (throws, not 404)
```

Archived sibling hashes return data successfully:

```
GET https://indexer.mainnet.autonomys.xyz/objects/39c21fb9a76eaaf3888eccd763f2aeeada4a1368f3d4bfc32e084e3f23c119a5
Response: {"blockNumber":7458212,"v0":{"objects":[["39c21fb9...",258918,267386]]}}
```

[...]

PR #677, opened 2026-05-05
fix: add archival reconciliation to resolve stuck objects (#674)

EmilFattakhov:
Objects with all nodes published on-chain but missing piece_index/piece_offset from the indexer remain permanently stuck in 'Archiving' status (~14K nodes, ~9K objects on mainnet). The root cause is the object-mapping-indexer missing chain events during WebSocket disconnections, with no backfill mechanism.

This adds a backend-side reconciliation system that:

- Periodically queries nodes stuck in archiving (published but no piece data)
- Batch-resolves their blake3 hashes against the indexer HTTP API
- Applies archival data and triggers object status checks
- Prioritizes newest objects first (ORDER BY metadata.created_at DESC)
- Processes in batches of 500, self-rescheduling via RabbitMQ when backlog exists, with a periodic 5-min interval as safety net

New files:
- core/objects/reconciliation.ts — core reconciliation logic
- services/dsn/objectMappingIndexerClient/index.ts — indexer HTTP client with batch fallback and circuit breaker
- services/reconciliationJob.ts — setInterval service (creditExpiryJob pattern)

Modified files:
- repositories/objects/nodes.ts — getUnreconciledNodes(), getUnreconciledNodesCount()
- eventRouter/tasks.ts — reconcile-archival task type
- eventRouter/processors/frontend.ts — task handler
- config.ts — reconciliation.intervalMs (RECONCILIATION_INTERVAL_MS env var)
- servers/frontendWorker.ts — job startup and shutdown

Addresses https://github.com/autonomys/auto-drive/issues/674

### DIFF
--- a/apps/backend/src/app/servers/frontendWorker.ts
+++ b/apps/backend/src/app/servers/frontendWorker.ts
@@ -70,10 +70,17 @@
   const shutdown = async () => {
     logger.info('Shutting down frontend worker...')
     objectMappingArchiver.stop()
-    const { reconciliationJob } = await import(
-      '../../infrastructure/services/reconciliationJob.js'
-    )
-    reconciliationJob.stop()
+    // reconciliationJob.stop() is safe even if never started (no-op),
+    // but we only import it when it was actually activated.
+    if (
+      config.featureFlags.flags.objectMappingArchiver.active &&
+      config.featureFlags.flags.taskManager.active
+    ) {
+      const { reconciliationJob } = await import(
+        '../../infrastructure/services/reconciliationJob.js'
+      )
+      reconciliationJob.stop()
+    }
     paymentManager.stop()
     const { creditExpiryJob } = await import(
       '../../infrastructure/services/creditExpiryJob.js'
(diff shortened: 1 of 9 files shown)


EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
Initially, here is the implementation I wanted to go with: adding gap detection and backfill logic to the object-mapping-indexer itself — tracking the last processed block number and replaying missed chain events on reconnect. This approach was later rejected because the indexer's WebSocket subscription API (`subspace_subscribeObjectMappings`) has no historical query capability, meaning backfill would require new RPC methods and iterating through potentially millions of blocks. A second alternative was querying the Subspace chain directly for object mappings at each affected block, which would be the most complete solution but also the slowest and most complex to implement. Instead, a backend-side periodic reconciliation was chosen: the backend already knows exactly which nodes are stuck (published on-chain but missing piece data), and can batch-resolve them against the indexer's existing HTTP API in a single targeted query. This approach requires no changes to the indexer or chain layer, ships fastest, and naturally handles both the current 14K-node backlog and any future disconnection gaps.

jim-counter (commented):
Has this been tested against a database backup?

jim-counter (approved):
LGTM. Thanks for the additional testing!

PR #690, opened 2026-05-14
fix: defensive ZLIB header validation to prevent crash on compression…

### LINKED ISSUE #689: File preview on the explorer fails with "Failed to fetch" for certain CIDs
### Summary

File preview on the explorer fails with "Failed to fetch" for certain CIDs (e.g., `bafkr6if72fohtg75d7ovwc5h7izoz7b6twdhwwdjtaqfnnwzxxkgg6ewoq`). The backend crashes with `Z_DATA_ERROR: incorrect header check` when attempting to decompress data that is not actually ZLIB-compressed, despite the metadata claiming `compression.algorithm: "ZLIB"`.

### Steps to Reproduce

1. Visit https://explorer.ai3.storage/mainnet/drive/metadata/bafkr6if72fohtg75d7ovwc5h7izoz7b6twdhwwdjtaqfnnwzxxkgg6ewoq
2. Observe the Preview section shows "Failed to fetch"

### Root Cause

The file's metadata declares `uploadOptions.compression.algorithm: "ZLIB"`, but the actual stored data is **uncompressed raw JSON**. This mismatch causes a crash in the download path:

1. The explorer's `FilePreview` component calls `fetch("https://gateway.autonomys.xyz/file/<cid>")`. Because this is a programmatic `fetch()` (not a document navigation), the browser sends `Sec-Fetch-Dest: empty` and `Sec-Fetch-Mode: cors`.

2. `handleDownloadResponseHeaders` (in `@autonomys/file-server`) checks `isDocumentNavigation(req)` which returns `false` for programmatic fetches.

3. Since the metadata says the file is compressed and it's not a document navigation, `shouldDecompressBody` is set to `true` — meaning the server should decompress before sending.

4. The download controller pipes the data through `createInflate()` (Node.js zlib):
   ```
   pipeline(sourceStream, createInflate(), res, ...)
   ```

5. **The stored data is plain JSON, not ZLIB-compressed.** `createInflate()` immediately fails with:
   ```
   Z_DATA_ERROR: incorrect header check (errno: -3)
   ```

6. By this point, response headers have already been sent (`res.headersSent` is true), so the error handler cannot write a 500 response. The connection is dropped silently.

7. The browser receives an incomplete/empty response → `TypeError: Failed to fetch`.

### Why it works via direct URL / curl (without Sec-Fetch headers)

When the same CID is accessed via direct browser navigation or curl (no `Sec-Fetch-Dest`/`Sec-Fetch-Mode` headers), `isDocumentNavigation()` returns `true`. In that path, the server sets `Content-Encoding: deflate` and sends the raw stored data as-is (no server-side decompression). The browser/Cloudflare then attempts to deflate-decode the body. Since the data is plain JSON (not actually compressed), this happens to "work" in some clients because curl/Cloudflare handle the encoding mismatch gracefully.

### Evidence from logs

```json
{
  "level": "ERROR",
  "name": "http:controllers:download",
  "msg": "Error streaming/decompressing data",
  "err": {
    "msg": "incorrect header check",
    "code": "Z_DATA_ERROR",
    "errno": -3
  }
}
```

### Affected file metadata

```json
{
  "name": "auto-memory-openclaw-agent-20260514T080236.json",
  "type": "file",
  "mimeType": "application/json",
  "totalSize": "632",
  "totalChunks": 1,
  "uploadOptions": {
    "compression": { "algorithm": "ZLIB" }  // ← claims ZLIB, but data is NOT compressed
  }
}
```

### Possible Fixes

1. **Upload validation (preventive):** Validate that data is actually ZLIB-compressed when `compression: ZLIB` is declared in upload metadata. Reject uploads where the metadata doesn't match reality.

2. **Graceful decompression fallback (defensive):** In the download controller, catch `Z_DATA_ERROR` from `createInflate()` and fall back to serving the raw data without decompression. Something like:
   ```typescript
   const inflate = createInflate()
   inflate.on('error', (err) => {
     if (err.code === 'Z_DATA_ERROR') {
       // Data isn't actually compressed — restart pipeline without inflate
       sourceStream.pipe(res)
     }
   })
   pipeline(sourceStream, inflate, res, ...)
   ```

3. **Fix the error response path:** Currently, when `res.headersSent` is true and decompression fails, the connection drops silently. At minimum, `res.destroy()` should be called to cleanly close the connection, and the error should be logged with the CID for traceability.

4. **Data repair:** Identify all files in the database with `compression: ZLIB` metadata whose stored data is not actually compressed, and either re-compress the data or correct the metadata.

### Environment

- Backend image: `ghcr.io/autonomys/auto-drive-backend:latest`
- `@autonomys/file-server` (handles `handleDownloadResponseHeaders`)
- Cloudflare → Nginx → Express (download-api container, port 3000)

---

EmilFattakhov:
… metadata mismatch

Files with compression: ZLIB in metadata but uncompressed stored data caused Z_DATA_ERROR crashes in the download controller. Now validates the ZLIB header (RFC 1950) before committing to the decompression pipeline and falls back to serving raw data on mismatch.

Also fixes silent error swallowing in both the download controller and the gateway file controller — errors are now always logged with the CID, and res.destroy() is called for clean connection teardown when headers are already sent.

Closes #689

### DIFF
--- a/apps/backend/src/app/controllers/download.ts
+++ b/apps/backend/src/app/controllers/download.ts
@@ -1,5 +1,5 @@
 import { Router } from 'express'
-import { pipeline } from 'stream'
+import { Readable, pipeline } from 'stream'
 import { createInflate } from 'zlib'
 import { asyncSafeHandler } from '../../shared/utils/express.js'
 import {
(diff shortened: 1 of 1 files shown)


EmilFattakhov:
bugbot run

EmilFattakhov:
Goes alongside https://github.com/autonomys/auto-files-gateway/pull/164

jim-counter (approved):
LGTM

PR #702, opened 2026-05-20
Fix/reconciliation transport and circuit breaker

### LINKED ISSUE #700: Reconciliation job cannot resolve nodes — HTTP transport mismatch with indexer
### Summary

The reconciliation job (introduced in #677) has never successfully resolved any nodes in production. It has been running every 5 minutes for 23+ hours with **0 nodes resolved across 275+ consecutive runs**. There are **14,787 unreconciled nodes** (across 9,005 objects) that have been published on-chain but are missing `piece_index`/`piece_offset` archival data.

The root cause is a **transport protocol mismatch**: the `objectMappingIndexerClient` sends JSON-RPC over HTTP POST, but the indexer server at `indexer.mainnet.autonomys.xyz` has no HTTP POST handler — it only supports **WebSocket JSON-RPC** and **REST `GET /objects/:hash`**.

#### Issue 1: HTTP POST → 404 (the transport bug)

`apps/backend/src/infrastructure/services/dsn/objectMappingIndexerClient/index.ts` creates an HTTP client:

```typescript
const httpClient = ObjectMappingIndexerRPCApi.createHttpClient(
  getHttpUrl(config.objectMappingArchiver.url),  // converts wss:// → https://
)
```

`createHttpClient` (from `@autonomys/rpc`) sends `POST /` with a JSON-RPC body. But the indexer's Express app only registers:

- `GET /objects/:hash` — REST lookup by hash
- `GET /objects/by-block/:blockNumber` — REST lookup by block
- `GET /health` — health check

**No POST route exists**, so every request returns:
```
HTTP 404: Cannot POST /
```

Verified with live testing:

| Transport | Result |
|---|---|
| `POST /` (what the code does) | `404 Cannot POST /` |
| `GET /objects/:hash` (REST) | `200 OK` with `{blockNumber, v0: {objects: [[hash, pieceIndex, pieceOffset]]}}` |
| WebSocket `get_object_mappings` | Works (returns `[[hash, pieceIndex, pieceOffset]]`) |

#### Issue 2: Indexer has gaps — missing data for ~14.8k nodes

Even after fixing the transport, the indexer genuinely does not have object mapping data for most of these nodes. Testing individual hashes via the working REST and WebSocket paths returns "Object mapping not found" or 500 errors.

The indexer (`object-mapping-indexer`) populates its database by subscribing to live `subspace_subscribeObjectMappings` events from a Substrate node. If the indexer was down or disconnected, it misses events permanently — **there is no backfill mechanism**.

Block-level analysis:
- **2,677 blocks** have ALL their nodes unreconciled (14,168 nodes) — indexer completely missed these
- **16 blocks** have a mix of reconciled and unreconciled nodes (147 nodes) — partial misses
- Blocks span from block 1,706,609 (Mar 2025) to 7,845,279 (May 2026)

The indexer's `GET /objects/by-block/:blockNumber` returns empty `[]` for most affected blocks, confirming the data was never indexed.

### Production Impact

| Metric | Value |
|---|---|
| Total nodes | 252,114 |
| Reconciled | 237,327 (94.1%) |
| **Unreconciled (published, no archival data)** | **14,315** |
| Unreconciled (not published) | 472 |
| Affected objects | 9,005 |
| Unarchived objects (metadata.is_archived=false) | 12,232 |
| Distinct blocks affected | 2,693 |
| Consecutive zero-progress runs | 275+ |
| RabbitMQ error queues | 8 messages in `frontend-errors`, 8 in `download-errors` |

#### Part 1: Fix the transport in `objectMappingIndexerClient` (backend change)

Replace the `createHttpClient` JSON-RPC approach with direct REST calls to `GET /objects/:hash`:

```typescript
const getObjectMappings = async (
  hashes: string[],
): Promise<Array<[string, number, number]>> => {
  const results: Array<[string, number, number]> = []
  let consecutiveFailures = 0

  for (const hash of hashes) {
    try {
      const response = await fetch(`${baseUrl}/objects/${hash}`)
      if (!response.ok) {
        consecutiveFailures++
        if (consecutiveFailures >= MAX_CONSECUTIVE_FAILURES) {
          logger.warn('Aborting after %d consecutive failures', consecutiveFailures)
          break
        }
        continue
      }
      const data = await response.json()
      // data.v0.objects is [[hash, pieceIndex, pieceOffset]]
      if (data?.v0?.objects) {
        results.push(...data.v0.objects)
      }
      consecutiveFailures = 0
    } catch (error) {
      consecutiveFailures++
      if (consecutiveFailures >= MAX_CONSECUTIVE_FAILURES) {
        logger.warn('Aborting lookups: %s', error instanceof Error ? error.message : String(error))
        break
      }
    }
  }
  return results
}
```

The `baseUrl` should be derived from `config.objectMappingArchiver.url` (converting `wss://` → `https://`), or ideally a separate config variable for the indexer HTTP URL.

#### Part 2: Indexer-side backfill (files-gateway change)

The indexer needs a mechanism to re-index historical object mappings it missed. Options:

1. **Block range re-indexer**: Add a CLI command or API endpoint to the `object-mapping-indexer` that takes a block range and re-queries the Substrate node for `subspace_subscribeObjectMappings` events from those blocks. The affected blocks are known (2,693 distinct blocks from the auto-drive database).

2. **Hash-based lookup from chain**: The Substrate node exposes `subspace_subscribeFilteredObjectMappings` — if this can filter by specific hashes, it could be used to selectively recover the ~14.8k missing mappings.

3. **Full re-index**: As a nuclear option, the indexer could be re-synced from genesis, though this would be time-consuming.

#### Part 3: Also fix the indexer's `GET /objects/:hash` 500 error

The indexer's `useCases/objectMapping.ts` throws an unhandled error when a hash is not found:

```typescript
const getObject = async (hash: string) => {
  const objectMapping = await objectMappingRepository.getByHash(hash)
  if (!objectMapping) {
    throw new Error('Object mapping not found')  // → causes 500 instead of 404
  }
  // ...
}
```

The controller should return a 404 response instead of letting this propagate as a 500.

### Reproduction

```bash
# From the backend container:
# HTTP POST (what reconciliation does) — fails with 404
curl -s -X POST 'https://indexer.mainnet.autonomys.xyz' \
  -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","method":"get_object_mappings","params":{"hashes":["a33f5704a16205bec6fc7c117661f5726faa907371bb68ae742b954239d6a6f5"]},"id":1}'
# Returns: Cannot POST /

# REST GET (the working endpoint) — succeeds for known hashes
curl -s 'https://indexer.mainnet.autonomys.xyz/objects/a33f5704a16205bec6fc7c117661f5726faa907371bb68ae742b954239d6a6f5'
# Returns: {"blockNumber":7907196,"v0":{"objects":[["a33f...",271350,607624]]}}
```

### Files Involved

- `apps/backend/src/infrastructure/services/dsn/objectMappingIndexerClient/index.ts` — transport fix needed
- `apps/backend/src/core/objects/reconciliation.ts` — reconciliation logic (OK, transport is the issue)
- `submodules/files-gateway/services/object-mapping-indexer/src/http/controllers/objects.ts` — 500→404 fix
- `submodules/files-gateway/services/object-mapping-indexer/src/useCases/objectMapping.ts` — error handling fix
- `submodules/files-gateway/services/object-mapping-indexer/src/rpc/server.ts` — only has WS subscription handlers, no `get_object_mappings`

EmilFattakhov:
## fix: reconciliation transport URL and circuit breaker (#700)

The reconciliation job (from #677) has never resolved any nodes in production — 275+ consecutive runs with 0 results. Two bugs in `objectMappingIndexerClient`, plus misleading diagnostics.

### Bug 1: HTTP POST sent to wrong path

`createHttpClient` sends `POST ${baseUrl}`, but the RPC server only accepts POST on `/ws`. The URL conversion (`wss://` → `https://`) didn't append `/ws`, so every request hit Express's 404.

Verified: `POST https://indexer.mainnet.autonomys.xyz/` → `Cannot POST /`
Verified: `POST https://indexer.mainnet.autonomys.xyz/ws` → returns mapping data

The indexer DB has complete data for all ~15K stuck nodes (confirmed via direct SQL and WebSocket RPC). Nothing was lost — the reconciliation just couldn't reach the data.

### Bug 2: Circuit breaker trips on expected misses

The indexer's `getObjectMappings` throws `'Object mapping not found'` for missing hashes. The fallback's circuit breaker counted these as failures, aborting after 10 consecutive misses. With the transport fix alone, any batch containing a missing hash would still fail.

### Changes

**`objectMappingIndexerClient/index.ts`**
- `getHttpUrl`: normalize URL to always end with `/ws`, handling trailing slashes and all input formats (`wss://`, `http://`, with or without `/ws`)
- Circuit breaker: don't count `'Object mapping not found'` RPC errors as transport failures

**`reconciliation.ts`**
- Fix misleading error message that claimed "indexer also missing their data" — the indexer has all the data, the transport was broken

**`.env.sample`**
- Update sample URL to include `/ws` path and remove trailing slash that would produce `//ws`

### Note

The production env var (`OBJECT_MAPPING_ARCHIVER_URL`) has already been updated to include `/ws`, which unblocked the reconciliation immediately. This code change is a defensive hardening so the transport works regardless of how the URL is configured — with or without `/ws`, with or without trailing slashes, using `wss://` or `http://`.

### Expected result

All ~15K stuck nodes resolve on the next reconciliation cycle (already in progress via the env var fix). This PR prevents the issue from recurring if the env var is ever changed back or set up without `/ws`.

Closes #700

### DIFF
--- a/apps/backend/.env.sample
+++ b/apps/backend/.env.sample
@@ -1,5 +1,5 @@
 DATABASE_URL=postgresql://postgres:postgres@localhost:5432/postgres
-OBJECT_MAPPING_ARCHIVER_URL=wss://indexer.mainnet.autonomys.xyz/ # update to your if you want
+OBJECT_MAPPING_ARCHIVER_URL=wss://indexer.mainnet.autonomys.xyz/ws # update to your own; /ws path is required for HTTP RPC transport
 FILES_GATEWAY_URL=http://changeme.com # change to your own (or it'll fail when fetching archived files)
 FILES_GATEWAY_TOKEN=changeme # change to your own (or it'll fail when fetching archived files)
 AUTH_SERVICE_API_KEY=1234567890 # change to your own if updated in auth
--- a/apps/backend/src/core/objects/reconciliation.ts
+++ b/apps/backend/src/core/objects/reconciliation.ts
@@ -13,8 +13,8 @@ const logger = createLogger('useCases:objects:reconciliation')
 
 const BATCH_SIZE = 500
 
-// Track consecutive zero-progress runs to detect permanently unresolvable
-// nodes (the indexer also missed them and can't provide their data).
+// Track consecutive zero-progress runs to detect potential connectivity
+// issues with the indexer or genuinely unresolvable nodes.
 let consecutiveZeroProgressRuns = 0
 const ZERO_PROGRESS_WARN_THRESHOLD = 3
 
@@ -165,9 +165,9 @@ const runReconciliationBatch = async (): Promise<void> => {
     consecutiveZeroProgressRuns++
     if (consecutiveZeroProgressRuns >= ZERO_PROGRESS_WARN_THRESHOLD) {
       logger.error(
-        'Reconciliation made no progress for %d consecutive runs. ' +
-          '%d unreconciled nodes may be permanently unresolvable ' +
-          '(indexer also missing their data). Consider indexer-side backfill.',
+        'Reconciliation made no progress for %d consecutive runs (%d unreconciled nodes). ' +
+          'Possible causes: indexer connectivity issue, transport misconfiguration, ' +
+          'or hashes genuinely missing from the indexer DB.',
         consecutiveZeroProgressRuns,
         unreconciledNodes.length,
       )
--- a/apps/backend/src/infrastructure/services/dsn/objectMappingIndexerClient/index.ts
+++ b/apps/backend/src/infrastructure/services/dsn/objectMappingIndexerClient/index.ts
@@ -5,10 +5,14 @@ import { createLogger } from '../../../drivers/logger.js'
 const logger = createLogger('dsn:objectMappingIndexerClient')
 
 const getHttpUrl = (url: string) => {
-  if (url.startsWith('http://') || url.startsWith('https://')) {
-    return url
-  }
-  return url.replace(/^ws(s)?:\/\//, 'http$1://')
+  const httpUrl =
+    url.startsWith('http://') || url.startsWith('https://')
+      ? url
+      : url.replace(/^ws(s)?:\/\//, 'http$1://')
+  // The RPC server only accepts POST on the /ws path.
+  // Ensure the URL ends with /ws regardless of input format.
+  const normalized = httpUrl.replace(/\/+$/, '')
+  return normalized.endsWith('/ws') ? normalized : normalized + '/ws'
 }
 
 const httpClient = ObjectMappingIndexerRPCApi.createHttpClient(
@@ -52,7 +56,17 @@ const getObjectMappings = async (
         }
         consecutiveFailures = 0
       } catch (error) {
-        consecutiveFailures++
+        // "Object mapping not found" is an expected miss — the indexer
+        // simply doesn't have this hash. Only count genuine transport
+        // or server errors toward the circuit breaker.
+        const isNotFound =
+          error instanceof Error &&
+          error.message.includes('Object mapping not found')
+        if (isNotFound) {
+          consecutiveFailures = 0
+        } else {
+          consecutiveFailures++
+        }
         if (consecutiveFailures >= MAX_CONSECUTIVE_FAILURES) {
           logger.warn(
             'Aborting individual lookups after %d consecutive failures (%d/%d resolved so far). Last error: %s',


EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM!

Interested to see the results of the config change already rolled out too.

PR #703, opened 2026-05-20
fix: Decrypt File button invisible in light theme

jim-counter:
## Summary
- The "Decrypt File" button in the encrypted file preview was invisible in light theme (white text on transparent background) — fixed by passing `isAutoDrive` prop to `AutoFilePreview`
- Changed `--auto-drive-accent` CSS variable from purple to blue to match the primary theme color
- The decryption password modal's "Confirm" button was also invisible — the design system uses `auto-explorer-buttonLightFrom`/`buttonDarkFrom` classes which had no color defined in the Tailwind config. Added mappings to the primary blue.
- Changed the "Bring to Cache" button from `lightAccent` to `primary` variant for visual consistency with the Download button

## Test plan
- [x] View an encrypted file in light theme — verify the "Decrypt File" button is visible and blue
- [x] Click "Decrypt File" and verify the password modal's "Confirm" button is visible and blue
- [x] Verify both buttons look correct in dark theme
- [x] Verify the "Bring to Cache" button matches the "Download" button style
- [x] Verify no other UI elements are affected by the color changes

### DIFF
--- a/apps/frontend/src/app/globals.css
+++ b/apps/frontend/src/app/globals.css
@@ -20,7 +20,7 @@ All colors MUST be HSL.
     --auto-drive-primary: 234 50% 48%;
     --auto-drive-primary-foreground: 0 0% 100%;
     --auto-drive-secondary: 243 75% 59%;
-    --auto-drive-accent: 264 83% 70%;
+    --auto-drive-accent: 234 50% 48%;
 
     /* Background gradients */
     --gradient-main: linear-gradient(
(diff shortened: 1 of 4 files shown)


EmilFattakhov (changes requested):
Screenshot 2026-05-20 at 6 07 04 PM

According to preview the button is still invisible. 

While we're at it, the download and bring to cache buttons also have slightly different shades

Screenshot 2026-05-20 at 6 07 33 PM

jim-counter:
Ah, I see now - didn't even realise there was a button there! The one I was fixing is this one:

image

Both corrected and tested.

EmilFattakhov (approved):
Looks good now!

PR #704, opened 2026-05-20
Fix/publishing recovery job

### LINKED ISSUE #555: Fix Race Condition in OnchainPublisher Account Manager
## Summary

The `AccountManager` in the onchain publisher has a race condition that causes `publish-nodes` tasks to fail with `TypeError: Cannot read properties of undefined (reading 'address')` when all accounts are removed from the pool during high-concurrency scenarios.

## Problem

When multiple transactions fail simultaneously, the account pool can become empty, causing subsequent transaction registrations to fail.

### Error Observed

```
TypeError: Cannot read properties of undefined (reading 'address')
    at accounts.js:41:46
```

### Root Cause

In `apps/backend/src/infrastructure/services/upload/onchainPublisher/accounts.ts`:

```typescript
const getNextAccount = () => {
  const account = accounts[trxCounter % accounts.length]  // undefined when accounts.length === 0
  trxCounter++
  return account
}

const removeAccount = (address: string) => {
  const index = accounts.findIndex((account) => account.address === address)
  if (index !== -1) {
    accounts.splice(index, 1)
  }

  if (accounts.length === 0) {
    return initAccounts()  // async, but not awaited by callers
  }
}
```

**Issues:**
1. `getNextAccount()` returns `undefined` when `accounts.length === 0` (division by zero edge case)
2. `removeAccount()` calls `initAccounts()` asynchronously but doesn't block new registrations while accounts are being reloaded
3. Multiple concurrent `removeAccount()` calls can race to empty the array

### Trigger Conditions

- High message throughput (`RABBITMQ_PREFETCH=100`)
- High transaction concurrency (`MAX_CONCURRENT_UPLOADS=40`)
- Substrate RPC returns nonce collision errors (`1014: Priority is too low`)
- Failed transactions trigger `removeAccount()`, eventually depleting the pool

## Proposed Solution

```typescript
export const createAccountManager = async (api: ApiPromise) => {
  let accounts = getAccounts()
  let isReinitializing = false

  const uniqueExec = pLimit(1)
  const nonceByAccount: Record<string, number> = {}
  let trxCounter = 0

  const initAccounts = () =>
    uniqueExec(async () => {
      isReinitializing = true
      try {
        accounts = getAccounts()
        const promises = accounts.map(async (account) => {
          const nonce = await getOnChainNonce(api, account.address)
          nonceByAccount[account.address] = nonce
        })
        await Promise.all(promises)
      } finally {
        isReinitializing = false
      }
    })

  const getNextAccount = () => {
    if (accounts.length === 0) {
      throw new Error('No accounts available')
    }
    const account = accounts[trxCounter % accounts.length]
    trxCounter++
    return account
  }

  const registerTransaction = () =>
    uniqueExec(async () => {
      // Wait for reinitialization if in progress
      if (accounts.length === 0 || isReinitializing) {
        await initAccounts()
      }
      
      const account = getNextAccount()
      const nonce = nonceByAccount[account.address]
      nonceByAccount[account.address] = nonce + 1
      return { account, nonce }
    })

  const removeAccount = async (address: string) => {
    return uniqueExec(async () => {
      const index = accounts.findIndex((account) => account.address === address)
      if (index !== -1) {
        accounts.splice(index, 1)
      }

      if (accounts.length === 0) {
        await initAccounts()
      }
    })
  }

  await initAccounts()

  return { registerTransaction, removeAccount }
}
```

### Key Changes

1. **Guard against empty array**: `getNextAccount()` throws explicitly when no accounts available
2. **Track reinitialization state**: `isReinitializing` flag prevents new transactions during account reload
3. **Make `removeAccount` async**: Ensure callers can await the reinitialization
4. **Serialize all operations**: Use `uniqueExec` consistently to prevent race conditions

## Alternative Considerations

1. **Circuit breaker pattern**: Stop accepting new tasks when error rate exceeds threshold
2. **Retry with backoff**: Add exponential backoff before reinitializing accounts
3. **Separate nonce tracking**: Query fresh nonce from chain instead of local tracking (slower but more reliable)
4. **Account balance monitoring**: Alert when any publishing account balance falls below threshold
5. **Never fully deplete pool**: Keep at least 1 account active even on failures, or pause processing

## Testing

- [ ] Unit test: `registerTransaction` when accounts array is empty
- [ ] Unit test: concurrent `removeAccount` calls
- [ ] Unit test: `registerTransaction` during `initAccounts` execution
- [ ] Integration test: high-concurrency publish-nodes with simulated RPC failures

### LINKED ISSUE #701: Objects stuck in "Publishing" state after publish-nodes batch failures
### Summary

Objects that require multiple batches of `publish-nodes` tasks can become permanently stuck in the "Publishing" state. When a batch fails and exhausts its retries, the task is dead-lettered to `frontend-errors` — a queue with **no consumer**. The object is left partially published with no automated recovery path.

There are currently **18 objects stuck in "Publishing"** in production. Furthermore, even if these objects are re-published, most will immediately get stuck at the **next stage ("Archiving")** due to the indexer transport bug described in #XXX (reconciliation job issue). The full lifecycle is blocked at two independent points for these objects.

### How publishing works

Large files are split into IPLD nodes. Publishing submits each node on-chain via `system.remark` in batches of 50 (`PUBLISH_BATCH_SIZE`). The pipeline is:

```
Upload → Migration → Tagging → Publishing (batched) → Archiving → Archived
                                    ↑                      ↑
                                 stuck here            also stuck here
                              (failed batches,       (indexer transport bug
                               retries exhausted)     from #700, data gaps)
```

1. `tag-upload` calls `scheduleNodesPublish(cid)`, which splits all node CIDs into batches of 50 and enqueues a separate `publish-nodes` task for each batch
2. Each `publish-nodes` task submits its 50 nodes to the Substrate chain via `OnchainPublisher.publishNodes()`
3. On success, each node's `block_published_on` and `tx_published_on` are updated in the DB
4. On failure, the task retries up to 3 times (`TASK_MANAGER_MAX_RETRIES`), then is published to `frontend-errors`

### The bug

Each batch is an **independent** RabbitMQ task. If batch 15 of 25 fails, batches 1–14 and 16–25 can still succeed. The object ends up with most nodes published but one batch's worth permanently missing. Since `frontend-errors` has no consumer, the failed task is silently dropped.

The "Publishing" status is computed (not stored) in `packages/models/src/objects/object.ts`:

```typescript
if (uploadState.uploadedNodes === uploadState.totalNodes) {
  return ObjectStatus.Archiving
}
return ObjectStatus.Publishing // 0 < uploadedNodes < totalNodes → stuck here forever
```

There is no timestamp, alert, or record of when publishing stalled — the object simply stays "Publishing" indefinitely.

### Production impact

**18 objects** are currently stuck in "Publishing" on mainnet:

| File | Total nodes | Published | Unpublished | Last published block |
|------|------------|-----------|-------------|---------------------|
| `IMG_5404.MOV` | 1,214 | 1,164 | 50 | 7,845,279 |
| `IMG_7241 2.mov` | 879 | 779 | 100 | 7,836,810 |
| `KNYT Rush front.enc` | 337 | 250 | 87 | 7,817,870 |
| `KNYT Rush back.enc` | 117 | 67 | 50 | 7,817,869 |
| `original_video.mp4` | 616 | 252 | 364 | 4,764,184 |
| `13705569_3840_2160_25fps.mp4` | 616 | 252 | 364 | 4,764,134 |
| *(+ 12 smaller objects with 1–14 unpublished nodes)* | | | | |

Objects span from block ~3.6M to ~7.8M — some have been stuck for months.

Note: the unpublished counts are exact multiples of 50 (the batch size) in most cases, confirming that whole batches failed.

### Relationship to https://github.com/autonomys/auto-drive/issues/700 (reconciliation/indexer issue)

Even after re-publishing the stuck nodes, these objects will **not** reach "Archived". They will move from "Publishing" to "Archiving" and then get stuck again, because:

- The reconciliation job that resolves archival data (`piece_index`/`piece_offset`) is non-functional due to the indexer transport mismatch (HTTP POST → 404)
- Even with the transport fixed, the indexer has data gaps — it missed ~2,677 blocks worth of object mapping events

For example, `IMG_5404.MOV` (1,214 nodes):
- **50 nodes** stuck at Publishing (this issue)
- **138 nodes** already published but stuck at Archiving (indexer can't provide `piece_index`)
- **1,026 nodes** fully archived

Both issues need to be resolved for these objects to complete their lifecycle.

### Existing recovery mechanism (unused)

The codebase already has an `ensure-object-published` task and `NodesUseCases.ensureObjectPublished()` that fetches all nodes for a CID and re-runs `OnchainPublisher.publishNodes()`. The handler is wired up in `frontendProcessor.ts`. However, **nothing in production enqueues this task** — it's a manual recovery hook with no automated trigger.

### Proposed solution: Periodic publishing recovery job

Add a `publishingRecoveryJob.ts` following the same pattern as the existing `reconciliationJob.ts`:

1. **Detect stuck objects** — Query for `root_cid`s where `COUNT(block_published_on) > 0` (some published) and `COUNT(block_published_on) < COUNT(*)` (some not), indicating partial publishing
2. **Apply a staleness threshold** — Only act on objects where `MAX(block_published_on)` is older than a configurable threshold (e.g., 1,000 blocks / ~1.5 hours). This avoids interfering with objects that are still being actively published in normal pipeline flow
3. **Enqueue `ensure-object-published`** tasks for each stuck CID, reusing the existing handler
4. **Limit throughput** — Process a bounded number of objects per cycle (e.g., 5) to avoid flooding the task queue
5. **Log recovery** — For observability

```typescript
// infrastructure/services/publishingRecoveryJob.ts
// Same setInterval + enqueue pattern as reconciliationJob.ts

const detectStuckPublishingObjects = async (): Promise<string[]> => {
  // SELECT root_cid FROM nodes
  // GROUP BY root_cid
  // HAVING COUNT(block_published_on) > 0
  //    AND COUNT(block_published_on) < COUNT(*)
  //    AND MAX(block_published_on) < (SELECT MAX(block_published_on) FROM nodes) - :threshold
  // LIMIT :maxPerCycle
}

const enqueueRecovery = async () => {
  const stuckCids = await detectStuckPublishingObjects()
  for (const cid of stuckCids) {
    EventRouter.publish(
      createTask({ id: 'ensure-object-published', params: { cid } })
    )
  }
}
```

### Additional considerations

- **`frontend-errors` queue has no consumer.** Failed tasks are silently dropped. This queue should at minimum be consumed for logging/metrics, and ideally tasks should be inspectable or retryable.
- **No observability for stuck objects.** There's no metric or alert for objects that have been in "Publishing" beyond a threshold. Adding one would catch this class of issue proactively.
- **`ensure-object-published` re-publishes all nodes**, including already-published ones. The `OnchainPublisher.publishNodes()` implementation does handle duplicates (via `handleRepeatedNodes` — it copies `block_published_on` from existing entries), so this is safe but slightly wasteful. A more targeted approach would only re-enqueue the unpublished node CIDs.

### Files involved

- `apps/backend/src/core/uploads/uploads.ts` — `scheduleNodesPublish()` batching logic (line 311)
- `apps/backend/src/infrastructure/eventRouter/utils.ts` — `createHandlerWithRetries()` retry + dead-letter logic
- `apps/backend/src/infrastructure/eventRouter/processors/frontend.ts` — task handler dispatch, `frontend-errors` queue name
- `apps/backend/src/infrastructure/eventRouter/tasks.ts` — `ensure-object-published` task definition (exists, unused)
- `apps/backend/src/core/objects/nodes.ts` — `ensureObjectPublished()` recovery function (exists, unused)
- `apps/backend/src/infrastructure/services/reconciliationJob.ts` — pattern to follow for the new job
- `packages/models/src/objects/object.ts` — `objectStatus()` computed status derivation

EmilFattakhov:
## fix: publishing recovery, AccountManager race condition, and queue priority

This PR addresses three related issues affecting the publish pipeline reliability:

### 1. Periodic publishing recovery job (Closes #701)

Objects that require multiple batches of `publish-nodes` tasks can become permanently stuck in "Publishing" when a batch fails and exhausts its retries. The failed task is dead-lettered to `frontend-errors` (a queue with no consumer), leaving the object partially published with no automated recovery path. **18 objects** are currently stuck in production.

**New files:**
- `core/objects/publishingRecovery.ts` — Detects partially-published objects and re-publishes only their unpublished nodes. Processes up to 5 objects per cycle (configurable via `PUBLISHING_RECOVERY_MAX_PER_CYCLE`). Has a concurrency guard to prevent overlapping runs.
- `infrastructure/services/publishingRecoveryJob.ts` — Periodic job (every 5 minutes, configurable via `PUBLISHING_RECOVERY_INTERVAL_MS`) that enqueues `recover-publishing` tasks when the pipeline queue is idle.

**Modified files:**
- `config.ts` — Added `publishingRecovery.intervalMs` and `publishingRecovery.maxObjectsPerCycle`
- `infrastructure/eventRouter/tasks.ts` — Added `recover-publishing` task schema and createTask case
- `infrastructure/eventRouter/processors/frontend.ts` — Added handler routing for `recover-publishing`
- `infrastructure/repositories/objects/nodes.ts` — Added `getStuckPublishingRootCids()` and `getUnpublishedNodeCidsByRootCid()`
- `app/servers/frontendWorker.ts` — Starts/stops `publishingRecoveryJob` alongside `reconciliationJob`

### 2. AccountManager race condition fix (Closes #555)

The `AccountManager` in the onchain publisher had a race condition where concurrent `removeAccount()` calls (triggered by failed transactions) could drain the account pool while `registerTransaction()` was reading it, causing `TypeError: Cannot read properties of undefined (reading 'address')`. This is likely the root cause of the batch failures that created the 18 stuck objects.

**Changes in `onchainPublisher/accounts.ts`:**
- Serialized `removeAccount` through the same `pLimit(1)` limiter as `registerTransaction` — all pool mutations are now mutually exclusive
- Added empty-pool guard in `registerTransaction` with auto-reinit
- Made `initAccounts` a plain async function (not `uniqueExec`-wrapped) to avoid `pLimit(1)` deadlock when called from within serialized operations
- Atomic account swap: only assigns fresh accounts after all nonces are successfully fetched, preventing partially-initialized state on `getOnChainNonce` failure

**Changes in `onchainPublisher/transactionManager.ts`:**
- Wrapped `removeAccount` call with try/catch so account removal failures don't swallow transaction results

### 3. Queue-based priority for fresh uploads

Recovery and reconciliation jobs now defer to active pipeline work. Fresh uploads always take priority.

**Changes:**
- `infrastructure/drivers/rabbit.ts` — Added `Rabbit.getMessageCount()` exposing `channel.checkQueue()` message count
- `reconciliationJob.ts` / `publishingRecoveryJob.ts` — Check `task-manager` queue depth before enqueuing. If any pipeline tasks are pending, recovery is deferred to the next interval
- Deferral lives in the job schedulers (not the use cases), so self-rescheduled reconciliation fast-drain batches always execute immediately

### Relationship to #700

This PR is based on `fix/reconciliation-transport-and-circuit-breaker` (#702). After the publishing recovery job re-publishes the stuck nodes, objects will move from "Publishing" to "Archiving", where the reconciliation job (fixed in #700/#702) will resolve their archival data.

Closes #701, Closes #555

### DIFF
--- a/apps/backend/.env.sample
+++ b/apps/backend/.env.sample
@@ -1,5 +1,5 @@
 DATABASE_URL=postgresql://postgres:postgres@localhost:5432/postgres
-OBJECT_MAPPING_ARCHIVER_URL=wss://indexer.mainnet.autonomys.xyz/ # update to your if you want
+OBJECT_MAPPING_ARCHIVER_URL=wss://indexer.mainnet.autonomys.xyz/ws # update to your own; /ws path is required for HTTP RPC transport
 FILES_GATEWAY_URL=http://changeme.com # change to your own (or it'll fail when fetching archived files)
 FILES_GATEWAY_TOKEN=changeme # change to your own (or it'll fail when fetching archived files)
 AUTH_SERVICE_API_KEY=1234567890 # change to your own if updated in auth
--- a/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
+++ b/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
@@ -5,6 +5,7 @@ import {
   DeleteObjectCommand,
   GetObjectCommand,
   HeadObjectCommand,
+  ListBucketsCommand,
   PutObjectCommand,
   S3Client,
   UploadPartCommand,
@@ -47,6 +48,10 @@ describe('AWS S3 - SDK', () => {
 
     s3Client = new S3Client({
       region: 'us-east-1',
+      // endpoint is required so that operations with no Bucket param (e.g.
+      // ListBuckets) have a base URL to target; bucketEndpoint:true then
+      // overrides the endpoint with the Bucket value for per-object operations.
+      endpoint: `${BASE_PATH}/s3`,
       credentials: {
         accessKeyId: 'e046e71c8dc3459c8da189e62418203a',
         secretAccessKey: '',
@@ -173,13 +178,68 @@ describe('AWS S3 - SDK', () => {
     )
   })
 
+  // Bucket-prefixed key: first segment becomes the bucket name
+  describe('Bucket-prefixed keys', () => {
+    const BucketedKey = 'my-archive/report.txt'
+    const BucketedBody = Buffer.from('Archived content')
+
+    it('should upload an object with a bucket-prefixed key', async () => {
+      const command = new PutObjectCommand({
+        Bucket,
+        Key: BucketedKey,
+        Body: BucketedBody,
+      })
+      const result = await s3Client.send(command)
+      expect(result).toMatchObject({ ETag: expect.any(String) })
+    })
+
+    it('should download a bucket-prefixed object', async () => {
+      const command = new GetObjectCommand({ Bucket, Key: BucketedKey })
+      const result = await s3Client.send(command)
+      expect(result.Body).toBeDefined()
+      expect(Buffer.from(await result.Body!.transformToByteArray())).toEqual(
+        BucketedBody,
+      )
+    }, 15_000)
+
+    it('should upload objects into a second bucket', async () => {
+      const command = new PutObjectCommand({
+        Bucket,
+        Key: 'another-bucket/file.txt',
+        Body: Buffer.from('Another bucket'),
+      })
+      const result = await s3Client.send(command)
+      expect(result).toMatchObject({ ETag: expect.any(String) })
+    })
+  })
+
+  describe('ListBuckets', () => {
+    it('should list all buckets', async () => {
+      const command = new ListBucketsCommand({})
+      const result = await s3Client.send(command)
+
+      expect(result.Buckets).toBeDefined()
+      expect(result.Buckets!.length).toBeGreaterThan(0)
+
+      const bucketNames = result.Buckets!.map((b) => b.Name)
+      // Flat keys land in 'default'; bucket-prefixed keys create named buckets
+      expect(bucketNames).toContain('default')
+      expect(bucketNames).toContain('my-archive')
+      expect(bucketNames).toContain('another-bucket')
+    })
+
+    it('each bucket should have a creation date', async () => {
+      const command = new ListBucketsCommand({})
+      const result = await s3Client.send(command)
+      for (const bucket of result.Buckets!) {
+        expect(bucket.CreationDate).toBeInstanceOf(Date)
+      }
+    })
+  })
+
   describe('DeleteObject', () => {
     it('should return 403 AccessDenied - storage is immutable', async () => {
-      const command = new DeleteObjectCommand({
-        Bucket,
-        Key,
-      })
-
+      const command = new DeleteObjectCommand({ Bucket, Key: 'test.txt' })
       await expect(s3Client.send(command)).rejects.toMatchObject({
         $metadata: { httpStatusCode: 403 },
         Code: 'AccessDenied',
--- a/apps/backend/__tests__/unit/core/s3.spec.ts
+++ b/apps/backend/__tests__/unit/core/s3.spec.ts
@@ -38,6 +38,7 @@ describe('S3UseCases', () => {
         .mockResolvedValue(null)
 
       const result = await S3UseCases.getObject({
+        Bucket: 'my-bucket',
         Key: 'nonexistent/file.txt',
       } as any)
 
@@ -46,7 +47,7 @@ describe('S3UseCases', () => {
     })
 
     it('should return error from download use case', async () => {
-      const mapping = { key: 'file.txt', cid: 'cid123' }
+      const mapping = { bucket: 'my-bucket', key: 'file.txt', cid: 'cid123' }
       jest
         .spyOn(s3ObjectMappingsRepository, 'findByKey')
         .mockResolvedValue(mapping as any)
@@ -58,6 +59,7 @@ describe('S3UseCases', () => {
         )
 
       const result = await S3UseCases.getObject({
+        Bucket: 'my-bucket',
         Key: 'file.txt',
       } as any)
 
@@ -65,7 +67,7 @@ describe('S3UseCases', () => {
     })
 
     it('should return download result successfully', async () => {
-      const mapping = { key: 'file.txt', cid: 'cid123' }
+      const mapping = { bucket: 'my-bucket', key: 'file.txt', cid: 'cid123' }
       const mockReadable = { read: jest.fn() }
 
       jest
@@ -77,6 +79,7 @@ describe('S3UseCases', () => {
         .mockResolvedValue(ok(mockReadable) as any)
 
       const result = await S3UseCases.getObject({
+        Bucket: 'my-bucket',
         Key: 'file.txt',
       } as any)
 
@@ -84,7 +87,7 @@ describe('S3UseCases', () => {
     })
 
     it('should pass byte range to download use case', async () => {
-      const mapping = { key: 'file.txt', cid: 'cid123' }
+      const mapping = { bucket: 'my-bucket', key: 'file.txt', cid: 'cid123' }
       const downloadSpy = jest
         .spyOn(DownloadUseCase, 'downloadObjectByAnonymous')
         .mockResolvedValue(ok({} as any) as any)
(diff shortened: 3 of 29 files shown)


EmilFattakhov:
needs to be merged after https://github.com/autonomys/auto-drive/pull/702

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM

Like the method of recovery jobs yielding to primary jobs.

PR #711, opened 2026-05-22
fix(s3): advance ListObjectsV2 continuation token past folded prefix

jim-counter:
## Summary

Fixes the duplicate-`CommonPrefixes` bug Cursor Bugbot flagged on [PR #696](https://github.com/autonomys/auto-drive/pull/696#discussion_r3265123834) and [PR #709](https://github.com/autonomys/auto-drive/pull/709#discussion_r3288865588).

When the DB batch in `listObjects` is exhausted with every row folding into a single `CommonPrefix`, the truncation fallback was setting `nextContinuationToken` to the **last raw key scanned** — a key that lives *inside* the folded virtual directory. The next page's `key > token` query then returned the remaining keys in the same directory and re-emitted the same `CommonPrefixes` entry.

The fix detects when the last scanned key folded into a `CommonPrefix` and advances the token to `<foldedPrefix>+￿`, so the next page resumes after the entire virtual directory. Non-delimiter and non-folded cases keep the previous last-key fallback.

The earlier reply in [#696 (comment)](https://github.com/autonomys/auto-drive/pull/696#discussion_r3265173584) considered dropping the truncation guard to avoid the duplicate — but that would re-introduce silent data loss when a prefix group is larger than the DB fetch limit (>10,000 keys). This change keeps the guard *and* eliminates the duplicate. ~26 lines in the use case + 123 lines of unit tests, no test surgery elsewhere.

## Test plan

- [x] `yarn workspace backend test __tests__/unit/core/s3.spec.ts` — 15/15 pass, including 3 new tests:
  - `advances continuation token past a folded CommonPrefix when the DB batch is exhausted inside one prefix group` (regression test for the bug)
  - `uses the raw last key as the token when the last scanned key did not fold into a prefix`
  - `uses the raw last key as the token when no delimiter is set`
- [x] `yarn workspace backend lint` clean
- [x] `tsc --noEmit -p tsconfig.test.json` clean
- [x] CI green
- [ ] Integration smoke against staging once deployed: list a virtual directory with >10k keys under `maxKeys=2` and confirm each page advances past the previous directory rather than repeating it

### DIFF
--- a/apps/backend/__tests__/unit/core/s3.spec.ts
+++ b/apps/backend/__tests__/unit/core/s3.spec.ts
@@ -369,4 +369,127 @@ describe('S3UseCases', () => {
       )
     })
   })
+
+  describe('listObjects', () => {
+    // dbLimit for delimiter listings is min(maxKeys * 10 + 100, 10_000), so
+    // maxKeys=2 yields dbLimit=120 — small enough to construct test data for.
+    const DELIMITER_DB_LIMIT = (maxKeys: number) =>
+      Math.min(maxKeys * 10 + 100, 10_000)
+
+    const makeListing = (key: string) => ({
+      key,
+      cid: 'cid',
+      size: 0n,
+      lastModified: new Date(0),
+    })
+
+    it('advances continuation token past a folded CommonPrefix when the DB batch is exhausted inside one prefix group', async () => {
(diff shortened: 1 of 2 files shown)


EmilFattakhov:
bugbot run

EmilFattakhov (approved):
Looks good!

Issue #712, opened 2026-05-22
Publishing pipeline stalled: NULL `encoded_node` crashes entire batch, stranding 1,981 valid nodes


EmilFattakhov:
### Root cause

The new periodic **publishing recovery job** (PR #701, merged May 20) re-enqueues `publish-nodes` tasks for the same 5 "stuck" objects every 5 minutes, but **some of those nodes have `encoded_node IS NULL` in the DB**, causing `publishNodes` to crash on `Buffer.from(null, 'base64')` and dump the failed tasks into `frontend-errors` after retries are exhausted.

### Evidence

The worker container (`auto-drive-backend-worker-1`, restarted at `2026-05-25T11:16:51Z`, image built `2026-05-22`) is producing this error in a tight loop:

```
TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received null
  at Function.from (node:buffer:320:9)
  at .../infrastructure/services/upload/onchainPublisher/index.js:21:31
  at Array.map (<anonymous>)
  at Object.publishNodes (.../infrastructure/services/upload/onchainPublisher/index.js:20:42)
  at .../infrastructure/eventRouter/utils.js:20:13
  at .../infrastructure/drivers/rabbit.js:79:17
```

Deployed `publishNodes` does not guard against null `encoded_node`:

```20:30:apps/backend/src/infrastructure/services/upload/onchainPublisher/index.ts
  const publishedRows =
    await nodesRepository.getNodesBlockchainDataBatch(uniqueCids)
  const publishedCidSet = new Set(publishedRows.map((r) => r.cid))

  const repeatedNodes = nodes.filter((node) => publishedCidSet.has(node.cid))
  await NodesUseCases.handleRepeatedNodes(repeatedNodes)

  // Nodes that are not known as published anywhere should be sent for publishing
  const unpublishedNodes = nodes.filter(
    (node) => !publishedCidSet.has(node.cid),
  )
```

(Production runs the pre-patch version where line ~20 immediately calls `Buffer.from(node.encoded_node, 'base64')` after the filter — your local working tree already has an in-progress guard for this.)

### Why these nodes have null `encoded_node`

`encoded_node` is set to `NULL` for every node in an object once the object is marked archived:

```179:186:apps/backend/src/infrastructure/repositories/objects/nodes.ts
const removeNodeDataByRootCid = async (rootCid: string) => {
[...]

PR #713, opened 2026-05-25
fix(backend): guard against NULL encoded_node in publishing pipeline

EmilFattakhov:
### Problem

The publishing recovery job (PR #701) re-enqueues `publish-nodes` tasks for objects stuck in partial-publishing state every 5 minutes. However, some of those nodes have `encoded_node = NULL` in the DB because `onObjectArchived` unconditionally strips node data via `removeNodeDataByRootCid` — even for nodes that were never published on-chain.

This causes `publishNodes` to crash on `Buffer.from(null, 'base64')`, dumping ~12 failed tasks per cycle into `frontend-errors` after retries exhaust. Over ~66 hours since the May 22 deploy this produced ~9,540 dead-letter messages, all from the same 5 stuck root CIDs (370 nodes total).

### Root cause

`removeNodeDataByRootCid` runs `UPDATE nodes SET encoded_node = NULL WHERE root_cid = $1` with no regard for publication status. When an object is archived while some nodes still have `block_published_on IS NULL`, those nodes become permanently un-publishable: they look unpublished to the recovery job but have no data to publish.

### Changes

**Commit 1 — Stop the crash and the re-enqueue loop**

- **`onchainPublisher/index.ts`**: Filter out nodes with `encoded_node == null` before `Buffer.from()`, log a warning for visibility, early-return when no publishable nodes remain.
- **`getUnpublishedNodeCidsByRootCid`**: Add `AND encoded_node IS NOT NULL` so the recovery job never enqueues un-publishable nodes.
- **`getStuckPublishingRootCids`**: Add a `FILTER` clause excluding objects where every unpublished node has already been stripped — they would only produce no-op task cycles.

**Commit 2 — Fix the root cause + surface existing broken data**

- **`removeNodeDataByRootCid`**: Add `AND block_published_on IS NOT NULL` so archival only strips `encoded_node` from nodes that are confirmed on-chain. Unpublished nodes retain their data for recovery.
- **New `getUnrecoverablePublishingRootCids` query**: Detects objects with nodes in the broken state (`block_published_on IS NULL AND encoded_node IS NULL`) from before this fix.
- **`publishingRecovery.ts`**: Logs a `warn` each cycle listing any unrecoverable objects by root_cid and node count, surfacing them for operational attention.

### Operational follow-up (not in this PR)

The 5 existing broken root CIDs need a one-off DB heal to write verified `piece_index`/`piece_offset` values. This PR prevents new objects from entering the broken state and stops the `frontend-errors` flood, but the existing 370 rows require manual resolution. After this PR deploys, those objects will appear in the recovery job warning logs.

Resolves step 1 in https://github.com/autonomys/auto-drive/issues/712

### DIFF
--- a/apps/backend/__tests__/unit/repositories/nodes.spec.ts
+++ b/apps/backend/__tests__/unit/repositories/nodes.spec.ts
@@ -370,22 +370,22 @@ describe('Nodes Repository', () => {
     expect(result?.piece_offset).toBe(100)
   })
 
-  it('should remove nodes by root CID', async () => {
+  it('should remove encoded_node only for published nodes by root CID', async () => {
     const rootCid = 'test-root-cid-remove'
     const nodes: Node[] = [
       {
-        cid: 'test-cid-remove-1',
+        cid: 'test-cid-remove-published',
         root_cid: rootCid,
         head_cid: 'test-head-cid-remove',
         type: 'file',
         encoded_node: 'test-encoded-node-remove-1',
         piece_index: null,
         piece_offset: null,
-        block_published_on: null,
+        block_published_on: 100,
         tx_published_on: null,
       },
       {
-        cid: 'test-cid-remove-2',
+        cid: 'test-cid-remove-unpublished',
         root_cid: rootCid,
         head_cid: 'test-head-cid-remove',
         type: 'file',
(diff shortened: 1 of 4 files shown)


EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM

PR #715, opened 2026-05-26
fix(backend): clear encoded_node when recovery-published node's objec…

EmilFattakhov:
…t is already archived

After onObjectArchived runs, removeNodeDataByRootCid only strips encoded_node for nodes already published (block_published_on IS NOT NULL). Nodes published later via recovery get block_published_on set but never lose encoded_node, causing indefinite local payload retention.

Fix updateNodePublishedOn to conditionally clear encoded_node when the node's metadata is already marked archived. This is safe because at that point the data is confirmed on-chain and the archival cleanup won't fire again.


EmilFattakhov:
Need to merge this one prior to https://github.com/autonomys/auto-drive/pull/714

jim-counter (approved):
LGTM

PR #708, opened 2026-05-22
fix(s3): add /s3 location block to public nginx config

jim-counter:
## Summary

- The S3 controller has been mounted at `/s3` on the download API since PR #446 (Aug 2025), but **no nginx location block was ever added** to route external `/s3` requests to Express on any host.
- The download API runs on `auto-drive-mainnet-public` at port 3000. This PR adds the missing `/s3` block to `backend.mainnet.public.conf`, proxying to `127.0.0.1:3000` — the same upstream used by `/file` and `/folder`.
- Includes `proxy_request_buffering off` and 600s timeouts for streaming large multipart uploads, plus S3-appropriate CORS headers (PUT/DELETE/HEAD + AWS SDK headers like `x-amz-content-sha256`, `Content-MD5`, etc.).

**No existing routing is affected** — this is a pure insertion of a new location block.

## Post deploy tests

- [ ] `sudo nginx -t` passes on the public box after copying the config
- [ ] `curl -I https://public.auto-drive.autonomys.xyz/s3/test` returns 401 (Express auth rejection, not nginx 404)

PR #698, opened 2026-05-19
docs: add rclone integration guide

jim-counter:
## Summary

- Adds `docs/rclone/` with a complete rclone integration guide for Auto Drive
- Covers all features that have shipped: ListObjectsV2, bucket model, MD5 ETags, immutable storage

## Contents

- **README.md** — Quickstart, operations overview table, bucket model, immutable storage summary
- **setup.md** — Account creation, API keys, credits, install instructions, full config reference
- **operations.md** — Every rclone command: what works, what doesn't, and why
- **workflows.md** — Archival, cron, browse/verify, mount, migration, CI/CD, and streaming recipes
- **troubleshooting.md** — Common errors and solutions (checksum mismatches, 403 on delete, auth, timeouts)
- **examples/rclone.conf.example** — Annotated config file
- **examples/archive-script.sh** — Cron-ready archive script with configurable bucket and prefix
- **examples/docker-compose.yml** — rclone HTTP server in Docker

## Note

- Should be last item merged in https://github.com/autonomys/auto-drive/milestone/4

PR #716, opened 2026-05-27
fix(s3): support AWS CLI clients (dispatch fallbacks + raw body)

jim-counter:
## Summary

The S3-compatible API only worked with clients that send the `x-id` query param **and** a `Content-Type` header (the AWS JS SDK does both). The **AWS CLI / botocore omits both**, which is why the `scripts/test-s3-compat.sh` smoke test failed against production on every object read/write while DeleteObject (hardcoded 403) and CreateMultipartUpload (no body) passed.

Two distinct bugs, both pre-existing in `main`, both in the public `download-api` code path:

1. **Dispatch** — `getS3Method` resolved the operation only from `x-id`, with fallbacks for `HEAD`/`DELETE` only. A plain `GET`/`PUT` with no `x-id` fell through to `undefined` → **"Method not found"** (the GetObject/PutObject 500s).
2. **Request body** — the S3 raw parser used `type: '*/*'`, but `type-is` returns `false` when there is **no `Content-Type` header**, so the body was never read. Combined with the global `express.json()` (which sets `req.body = req.body || {}` before its own content-type check), `req.body` arrived as `{}` → `Buffer.concat` TypeError deep in the upload path (the UploadPart 500s). The `Invalid part index` error was a cascade of the failed first part.

## Changes

- `getS3Method`: add `GET → GetObject` and `PUT → PutObject` fallbacks, ordered **after** the existing multipart (`uploadId`/`partNumber`) and `list-type=2` checks so those aren't misrouted.
- S3 raw body parser: `type: () => true` — object bodies are opaque binary; always read them as raw bytes regardless of Content-Type.
- `download.ts`: mount the S3 controller **before** the global JSON/urlencoded parsers so they never touch `/s3` (and moved `cors()` above it so `/s3` still gets CORS).

No behavior change for non-`/s3` routes. Build + lint pass.

## Test plan

- [ ] **Needs a real round-trip with the AWS CLI** (not just the JS SDK, which masks both bugs). Run `scripts/test-s3-compat.sh` against a deploy of this branch:
  - [ ] PutObject / GetObject / HeadObject succeed (no more "Method not found" / 500)
  - [ ] Multipart upload + composite ETag succeeds
  - [ ] DeleteObject still returns 403
  - [ ] ListBuckets / ListObjectsV2 unaffected
- [ ] TODO (follow-up): add an automated test that exercises the no-`x-id` / no-`Content-Type` path so this can't regress — the existing SDK-based integration tests won't catch it.

Related: nginx exposure of `/s3` (#708, merged) and rclone docs (#698).

### DIFF
--- a/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
+++ b/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
@@ -415,4 +415,100 @@ describe('AWS S3 - SDK', () => {
       expect(result.Metadata?.cid).toBeDefined()
     })
   })
+
+  // Raw HTTP requests that mimic the AWS CLI / botocore, which (unlike the JS
+  // SDK used above) does NOT send the `x-id` query param for GetObject/
+  // PutObject and sends object bodies with no Content-Type header. These guard
+  // two regressions:
+  //   1. getS3Method must fall back to the HTTP method (GET->GetObject,
+  //      PUT->PutObject) when `x-id` is absent — otherwise dispatch returns
+  //      "Method not found".
+  //   2. The request body must be read as raw bytes regardless of Content-Type;
+  //      a missing Content-Type previously left req.body as {} and broke
+  //      uploads deep in the IPLD chunker.
+  describe('Raw HTTP requests (AWS CLI style: no x-id, no Content-Type)', () => {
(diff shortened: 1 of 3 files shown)


jim-counter:
bugbot run

EmilFattakhov (approved):
LGTM!

PR #695, opened 2026-05-18
feat(s3): return MD5 ETag and expose CID via x-amz-meta-cid

jim-counter:
## ⚠️ Breaking Change

**ETag values returned by `PutObject`, `HeadObject`, `GetObject`, and `CompleteMultipartUpload` change from Autonomys CIDs to MD5 hashes.**

Before: `ETag: bafy...` (CID)
After: `ETag: "d41d8cd98f00b204e9800998ecf8427e"` (quoted MD5 hex — standard AWS S3 format)

**The CID is not lost** — it moves to a new `x-amz-meta-cid` response header on all object responses. S3 clients that used the ETag as a CID reference must switch to reading that header.

Objects uploaded before this migration have `NULL` md5 in `object_mappings` and will not receive an `ETag` header on `HEAD`/`GET` until they are re-uploaded.

---

## Summary

- MD5 computed at upload time and stored in `object_mappings` (new `md5` column)
- `PutObject` and `HeadObject` return `ETag: "md5hex"` (standard S3 format)
- `UploadPart` now returns MD5 of the part body instead of a random UUID v4
- `CompleteMultipartUpload` computes the standard AWS composite ETag: `"<md5_of_part_md5s>-<N>"`
- All object responses set `x-amz-meta-cid` so the CID remains accessible
- `rclone check` and `rclone md5sum` work without `--ignore-checksum`

## Changes

- `apps/backend/migrations/20260513000000-s3-md5-etag.*` — migration (nullable `md5` column)
- `apps/backend/src/infrastructure/repositories/s3/objectMappings.ts` — `md5` in `createMapping` and `S3KeyMapping`
- `apps/backend/src/core/s3/index.ts` — MD5 computation, composite ETag logic, local type extensions
- `apps/backend/src/app/controllers/s3/s3.ts` — `ETag` + `x-amz-meta-cid` headers; XML body parsing for multipart parts
- `apps/backend/__tests__/unit/core/s3.spec.ts` — MD5 format assertions, known-value MD5 test
- `apps/backend/__tests__/integration/s3-sdk/index.spec.ts` — ETag regex assertions, `x-amz-meta-cid` on HeadObject

## Stacked on

This PR is based on `feat/s3-bucket-support` (#684). It must be merged after #684 and rebased onto `main` at that point.

Closes #682
Part of milestone: [rclone S3 Compatibility](https://github.com/autonomys/auto-drive/milestone/4)

PR #717, opened 2026-05-27
fix(s3): return MD5 ETag from ListObjectsV2 (not the CID)

jim-counter:
## Summary

HeadObject was fixed to return the stored MD5 as the ETag (#695), but the **ListObjectsV2 listing path was missed** and still returned the **CID** as the ETag. Clients that verify checksums from a listing (rclone, AWS CLI) can't use a CID as an MD5, so they fall back to a `HeadObject` per object to get the hash.

The `md5` is already stored on `S3.object_mappings` (HeadObject reads it) — the listing query simply didn't select it.

## Changes
- `repositories/s3/objectMappings.ts`: add `md5` to `S3ObjectListing`, select `om.md5` in the `listObjects` query, map it through.
- `controllers/s3/s3.ts`: `listObjectsV2Handler` now emits `ETag: obj.md5 ? "<md5>" : "<cid>"` — real MD5, falling back to the CID for legacy objects with no stored md5 (matching HeadObject; CID stays in `x-amz-meta-cid`).
- Test: assert listed ETags are MD5 format **and equal HeadObject's ETag** for the same object. Updated the `buildListResult` unit fixture for the new field.

`md5` rides along on `S3ObjectListing` through `buildListResult`, so no use-case changes were needed.

## Context
Surfaced while validating rclone against the deployed S3 API. The production smoke test (incl. `rclone check`) already passes **because** rclone falls back to HeadObject for the hash — so this is a **correctness + performance** improvement (self-consistent listings, no N extra HEAD requests), not a functional blocker.

Note: rclone also requires `list_version = 2` in its remote config, because our API implements ListObjectsV2 only (not V1). Documented separately in the rclone docs; server-side ListObjectsV1 support is a possible follow-up.

## Test plan
- [x] `yarn backend test --testPathPattern='s3-sdk|core/s3'` → 40/40 pass
- [x] Build + lint clean
- [ ] Post-deploy: re-run `scripts/test-s3-compat.sh` (should still pass, now without per-object HEAD fallback) and an `rclone check`

### DIFF
--- a/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
+++ b/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
@@ -283,6 +283,24 @@ describe('AWS S3 - SDK', () => {
       }
     })
 
+    it('should return the MD5 ETag (not the CID), matching HeadObject', async () => {
+      // Regression: the listing previously returned the CID as the ETag, so
+      // checksum-verifying clients (rclone, AWS CLI) could not validate against
+      // a listing. Every listed ETag must be a quoted 32-hex MD5 (a CID would
+      // not match), and must equal the ETag HeadObject returns for that object.
+      const list = await s3Client.send(
+        new ListObjectsV2Command({ Bucket: ListBucket }),
+      )
+      const entry = list.Contents!.find((c) => c.Key === 'a.txt')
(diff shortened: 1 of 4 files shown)


EmilFattakhov (approved):
Looks good!

PR #719, opened 2026-05-28
chore: consume Auto SDK S3 helpers + add live integration tests

jim-counter:
## Summary

Closes #718. Three commits:

1. **`refactor(s3): consume ListObjectsV2 + ETag helpers from @autonomys/file-server`** — bumps all `@autonomys/*` deps to `1.6.12` and replaces the local copies of `buildListResult` / `md5Hex` / `formatETag` / `multipartETag` / `ListObjectsParams` / `ListObjectsResult` in `apps/backend/src/core/s3/index.ts` with imports from the SDK's new `s3/` module. `listObjects` collapses to a three-line consumer (`computeListObjectsDbLimit` + `finalizeListObjects`). The repository's `S3ObjectListing` is re-exported from the SDK rather than re-declared (SDK 1.6.12 added the `md5?: string | null` field). Net `-291 / +144`.

2. **`test(s3): add live integration tests for the S3 API`** — new `scripts/live-integration/test-s3-api.sh` covering every method in the controller's `S3HandlerConfig` plus negative paths and unsupported-operation probes. Black-box AWS CLI tests against a deployed endpoint.

3. **`test(s3): add live integration tests for the rclone client`** — new `scripts/live-integration/test-s3-rclone.sh` covering rclone listing variants (`ls`, `lsd`, `lsjson`, `lsf -R`, `tree`, `size`, `hashsum md5`), read (`cat` + range), write (`copyto`, `copy`, `rcat`), and immutability rejection (`deletefile`, `delete`, `purge`, `moveto`).

## Bug fixes inherited from SDK 1.6.12

- `MaxKeys=0` no longer crashes `buildListResult` (`sortedObjects[scanIdx - 1]` indexed `-1` on empty input). The backend controller clamps `maxKeys` to `[1, 1000]` so the bug couldn't reach the SDK from this code path, but the fix benefits any direct caller.
- Multi-character delimiters now fold correctly (`delimIdx + delimiter.length` instead of hard-coded `+ 1`). The backend only passes `/` today so this didn't bite in production.

## Live integration tests — design notes

These tests are **not hermetic**: they hit a real deployed S3 endpoint with a real API key. They are not wired into CI — intended to be run by developers and operators against staging or production. They live under `scripts/live-integration/` (not `apps/backend/__tests__/integration/`, which is for hermetic Jest + TestContainers).

[...]

PR #720, opened 2026-05-29
chore: harden update-autonomys script + bump Auto SDK @autonomys/* to 1.6.13

jim-counter:
## Summary

Two commits, rebased onto `main` after #719 merged:

1. **`chore(scripts): harden update-autonomys`** — three structural fixes to the bump script, the most important being switching from `yarn workspace <name> up` to `yarn workspace <name> add`. `yarn up` is project-wide despite the workspace prefix; `yarn add` is actually workspace-scoped.
2. **`chore(deps): bump @autonomys/* to 1.6.13`** — live validation by running the hardened script. The diff is what real-world usage produces, end-to-end.

## What was actually wrong (and why the first attempt at this PR was incomplete)

When I first wrote this PR I thought the bug from #719 was that `yarn workspaces list --json` enumerated submodule workspaces, and that filtering them out would prevent the manifests from being touched. That's necessary but not sufficient.

Bundling the `1.6.13` bump caught it: with only the filter in place, running `yarn update:autonomys 1.6.13` *still* mutated every submodule manifest. The reason is buried in yarn's docs:

> `yarn up` upgrades the packages matching the list of specified patterns to their latest available version **across the entire project** (regardless of whether they're part of your top-level workspace or not).

The `yarn workspace <name>` prefix only sets the working directory for the command — `up` itself scans every workspace that declares the package and rewrites their manifests. So even after filtering submodule workspaces out of the iteration set, the very first `yarn workspace backend up @autonomys/rpc@1.6.13` call still rewrote the submodule's `@autonomys/rpc: ^1.4.31` to `1.6.13`.

The real fix is `yarn workspace <name> add <descriptor>` instead. `add` only modifies the named workspace's `package.json`. The lockfile still re-resolves project-wide so the submodule's caret ranges absorb the new version through dedup — which is exactly what we want.

## Why caret-range absorption is the right model

All `@autonomys/*` deps inside `submodules/files-gateway/` use **caret ranges**, not exact pins:

| Workspace | `@autonomys/*` deps |
|---|---|
| `auto-files-gateway` | `@autonomys/rpc: ^1.4.31` |
| `file-retriever` | `@autonomys/{asynchronous: ^1.4.36, auto-dag-data: ^1.4.31, auto-utils: ^1.4.31, file-caching: ^1.4.36, rpc: ^1.4.31}` |
| `object-mapping-indexer` | `@autonomys/{auto-dag-data: ^1.4.35, rpc: ^1.4.31}` |

When Auto Drive's main-repo workspaces pin `@autonomys/rpc: "1.6.13"` exact, yarn's lockfile solver finds one version satisfying both the exact pin and the caret range — `1.6.13` — and emits a single lockfile entry covering every consumer:

```
"@autonomys/rpc@npm:1.6.13, @autonomys/rpc@npm:^1.4.31, @autonomys/rpc@npm:^1.5.6":
  version: 1.6.13
```

The submodule's `package.json` files never need to change for the submodule to pick up new SDK versions. So the submodule effectively has two update paths, neither of which involves this script mutating its manifests:

- **Implicit, every bump (common):** main repo bumps the exact pin, lockfile re-resolves, the submodule's caret ranges silently absorb the new version.
- **Explicit floor-raise (rare):** if the submodule's authors decide `^1.4.31` is too permissive and want it raised, they edit the manifest *in the files-gateway repo*, commit, push, and Auto Drive advances its submodule pointer to the new SHA.

The old script's behaviour — replacing an intentional `^1.4.31` range with an exact pin — was *narrowing* a contract the submodule's authors had deliberately left flexible.

## Three improvements in commit 1

1. **`yarn add` instead of `yarn up`** (the actual root cause).
2. **Skip workspaces under `submodules/`** when iterating — necessary alongside #1 to keep the iteration set honest.
3. **Derive the `@autonomys/*` package list from manifests** instead of hardcoding it. The old array was a silent footgun in both directions:
   - A newly-adopted `@autonomys/*` package would not be bumped unless someone remembered to add it.
   - The list drifted the other way: `@autonomys/design-tokens` was listed but consumed nowhere.

Plus up-front logging of the skipped workspaces and the derived package list, so the operator sees what is and is not being bumped before any mutation happens.

## Live validation (commit 2)

`yarn update:autonomys 1.6.13` produced exactly the diff in commit 2 against `main` (post-#719, at 1.6.12):

- **5 main-repo manifests** rewritten: `apps/backend`, `apps/frontend`, `packages/{models,s3,ui}`
- **`yarn.lock`** re-resolved (caret-range entries now point at `1.6.13`)
- **0 submodule manifests touched** — `git -C submodules/files-gateway status` is clean
- **`yarn install --immutable` passes** locally; the `lockfile would have been modified` failure from #719 does not recur
- **Consumption layer verified**: `yarn backend build` clean and all 15 tests in `apps/backend/__tests__/unit/core/s3.spec.ts` pass against #719's SDK migration code at 1.6.13.

## Test plan

- [x] `node -c scripts/update-autonomys.js`
- [x] Read-only verification of discovery logic against `main`
- [x] Live run with `1.6.13`: 0 submodule manifests touched
- [x] `yarn install --immutable` passes locally
- [x] `yarn backend build` clean at 1.6.13
- [x] 15/15 S3 unit tests pass at 1.6.13
- [x] CI green on this branch


jim-counter:
bugbot run

EmilFattakhov (approved):
Awesome 🚀

PR #726, opened 2026-05-31
fix(uploads): support out-of-order multipart part arrival

### LINKED ISSUE #725: S3 multipart upload: out-of-order parts produce corrupted assembled object
## Summary

S3 multipart uploads where parts arrive in any order other than strict ascending `PartNumber` produce a **corrupted assembled object**. Parts that arrive out of sequence are silently rejected at the processing layer, and `CompleteMultipartUpload` still succeeds — but the downloaded object contains only the parts that happened to arrive in the expected sequence.

This violates a core S3 contract: clients are explicitly allowed to upload parts in **any order**, in parallel, and to retry individual parts. `PartNumber` identifies a part's *position* in the final assembly, not its arrival order.

Surfaced by the live integration test added in #719:

```
PR 695 — Multipart upload
  ✓  Multipart 3-part — completed with parts listed out of order
  ✗  Multipart 3-part — assembled bytes do not match expected order
      local=df11bcd2…  remote=1bd4db45…
```

The test uploaded three 16 KiB blocks of distinct content as `[PartNumber: 2, 1, 3]` and completed with parts listed `[3, 1, 2]`. Expected assembled body = `A→B→C` (in `PartNumber` order); got something else (only chunk A made it through processing).

## Root cause

[`apps/backend/src/core/uploads/uploadProcessing.ts:67-71`](apps/backend/src/core/uploads/uploadProcessing.ts):

```ts
const expectedPartIndex =
  lastProcessedPartIndex == null ? 0 : lastProcessedPartIndex + 1
if (index !== expectedPartIndex) {
  throw new Error(`Invalid part index: ${index} !== ${expectedPartIndex}`)
}
```

`processChunk` enforces strict sequential arrival. The flow today:

1. `S3UseCases.uploadPart` calls `UploadsUseCases.uploadChunk` with `zeroIndexedPartNumber = PartNumber - 1`.
2. `uploadChunk` calls `processChunk` **first**, then `filePartsRepository.addChunk`.
3. `processChunk` throws on any non-sequential index, so the chunk is never persisted in `uploads.file_parts` either.
4. The error surfaces as a 500 from the upload-part request, but only after the per-part ETag was already returned to the client — and clients (including AWS CLI) keep going regardless.
5. `CompleteMultipartUpload` runs against whatever did make it through the blockstore — typically just whichever contiguous prefix of parts arrived in order.

The streaming IPLD-during-upload optimization (intentional for the single-PUT and well-behaved-in-order cases) is fundamentally incompatible with S3's any-order multipart contract.

## Why this hasn't been caught before

- Existing hermetic tests in `__tests__/integration/s3-sdk/index.spec.ts` only exercise single-PUT and single-part multipart.
- rclone, AWS CLI, and the SDKs all upload parts **in `PartNumber` order serially by default** unless the client opts into concurrent uploads (e.g. AWS CLI's `s3 cp` with `--multipart-chunksize` and many small parts, or any client tuned for high throughput).
- Production clients that have used multipart have been small enough or sequential enough not to trigger it.

## Blast radius

Anywhere clients upload S3 multipart with non-monotonic arrival:
- Parallel uploads (the most common multipart pattern in production)
- Retried parts after a transient failure
- Any client that intentionally reorders for throughput

## Fix

Decouple "store the raw chunk" from "process into the IPLD tree":

1. In `uploadChunk`, **store the raw chunk first** via `filePartsRepository.addChunk`. The PK is `(upload_id, part_index)` so each part lands at its intended position regardless of arrival order.
2. Replace the strict-sequence check in `processChunk` with a **drain** that processes contiguous chunks in `part_index` order, starting from `last_processed_part_index + 1`. Each call to `uploadChunk` opportunistically drains as much as it can — for an in-order client, it processes exactly the chunk it just stored (streaming optimisation preserved); for an out-of-order client, it processes nothing until the next expected chunk arrives, then catches up.
3. In `completeUpload`, run the drain unconditionally before finalising — covers any remaining pending chunks.

Single-PUT (`putObject` calls `uploadChunk` with `index=0` once) is unaffected: store at 0, drain at 0, finalise.

## Out of scope (separate issues if needed)

- **Concurrent uploadPart calls for the same upload_id** — the existing `processChunk` check-then-act on `last_processed_part_index` is already a classic race; this fix makes the same race visible at the drain step. Real-world S3 clients upload parts concurrently, so this needs an advisory lock per upload — file as follow-up.
- **Part retries (same `PartNumber` uploaded twice)** — `filePartsRepository.addChunk` currently `INSERT`s, which would fail on the PK after a retry. S3 allows part retries; we should `INSERT ... ON CONFLICT (upload_id, part_index) DO UPDATE`. File as follow-up.

## Regression test

Add a hermetic test in `__tests__/integration/s3-sdk/index.spec.ts` that uploads three parts as `[2, 1, 3]`, completes with parts listed `[3, 1, 2]`, downloads, and asserts the body equals `A‖B‖C` (where each block has distinct content so misordering is detectable).

## Verification plan

- [x] New hermetic out-of-order multipart test passes
- [x] All existing backend tests still pass
- [x] Re-run `scripts/live-integration/test-s3-api.sh` against the deploy; "Multipart 3-part — assembled bytes do not match expected order" becomes `ok`

jim-counter:
Closes #725.

## What was wrong

S3 multipart explicitly allows parts to arrive in any order — parallel uploads, retries, throughput-optimised clients all rely on this. `PartNumber` identifies a part's *position* in the final assembly, not its arrival order.

The previous implementation rejected non-monotonic arrivals at the processing layer:

```ts
// apps/backend/src/core/uploads/uploadProcessing.ts
const expectedPartIndex =
  lastProcessedPartIndex == null ? 0 : lastProcessedPartIndex + 1
if (index !== expectedPartIndex) {
  throw new Error(`Invalid part index: ${index} !== ${expectedPartIndex}`)
}
```

`uploadChunk` called `processChunk` *before* persisting the raw bytes via `filePartsRepository.addChunk`, so an out-of-order chunk threw before ever landing in `uploads.file_parts`. The error surfaced as a 500 from upload-part, but only after the per-part ETag had already been computed and returned — and clients (including AWS SDK and CLI) treat upload-part as fire-and-forget per Part. `CompleteMultipartUpload` then ran against whatever did make it through and silently produced a corrupted assembled object.

Surfaced by the live integration test added in #719 (a 3-part upload arriving as `[PartNumber: 2, 1, 3]`); only chunk A made it through and the assembled SHA256 differed from local `A‖B‖C`.

## The fix

Decouple "store the raw chunk" from "process into the IPLD tree":

1. **In `uploadChunk`**, persist the raw chunk first via `filePartsRepository.addChunk`. The `(upload_id, part_index)` primary key means each chunk lands at its declared position regardless of arrival order. `addChunk` is now an UPSERT (`ON CONFLICT (upload_id, part_index) DO UPDATE`) so part retries replace the stored bytes instead of 500'ing on the PK.
2. **After storing**, run a drain that processes chunks in `part_index` order starting from `last_processed_part_index + 1`. For an in-order client this processes exactly the chunk just stored and exits immediately — the streaming-during-upload optimisation is preserved. For an out-of-order client it is a no-op until the next expected chunk arrives, at which point it catches up by processing every already-stored chunk that now follows in sequence.
3. **Per-upload in-process drain lock.** Concurrent `uploadPart` requests for the same `upload_id` chain through a `Map<uploadId, Promise>` so their drains do not interleave reads and updates of `last_processed_part_index`. Without this, two concurrent drains could read the same `next` index, both fetch the same chunk, and both call `processChunk` — silently double-processing bytes into the IPLD tree.
4. **In `completeUpload`** (for `FILE` uploads only), run the drain unconditionally, then `assertNoUnprocessedParts` refuses to finalise when any stored part is unprocessed. Without this check a gap in the parts list — e.g. client uploaded 1 and 3 but never 2 — would emit a CID whose stored byte total includes bytes that are absent from the IPLD tree. The assertion raises a 400 with the offending `part_index` values.
5. Folder uploads have no `file_processing_info` row and skip the drain entirely; their finalisation runs through `processFolderUpload` as before.

## Bugbot findings — addressed

The original commit (`28109199`) drew three valid findings from Cursor Bugbot:

| # | Severity | Finding | Status |
|---|---|---|---|
| 1 | Medium | Completion ignored unprocessed stored parts | Fixed via `assertNoUnprocessedParts` |
| 2 | Medium | Persisted part blocked S3 retry | Fixed via `INSERT ... ON CONFLICT DO UPDATE` |
| 3 | High | Parallel part drains raced | Fixed via in-process `withDrainLock` |

The original PR body framed #2 and #3 as "known follow-ups". That was wrong: #2 is a regression my changes *introduced* (pre-this-branch, `processChunk` threw before `addChunk` so retries simply re-attempted everything), and #3 was made *worse* by my changes — pre-fix the race usually threw `Invalid part index` (visible failure); post-fix it could silently corrupt (worse failure mode). Both belong in this fix.

## Regression tests (three new)

`apps/backend/__tests__/integration/s3-sdk/index.spec.ts` — under the `Out-of-order multipart upload` describe block:

- **`uploaded as [2,1,3] and completed as [3,1,2] should assemble to A‖B‖C`** — the original test, locking the core out-of-order assembly behaviour.
- **`completion should refuse when there is a gap in the parts sequence`** — uploads PartNumbers 1 and 3, expects 400 on CompleteMultipartUpload.
- **`should accept a part retry with the same PartNumber`** — uploads the same PartNumber twice, completes, verifies assembled body matches retried bytes. Pre-fix 500'd on PK violation.
- **`concurrent uploadParts should still assemble correctly`** — Promise.all of three UploadPart commands then complete, asserts body is `A‖B‖C`. Stochastic but catches a removed-lock regression.

## Remaining known limitation

Multi-process serialisation. The drain lock is in-process only; concurrent `uploadPart` for the same `upload_id` arriving at two different backend processes would still race. A PostgreSQL advisory lock would be the cross-process fix but requires routing every drain step through a single connection — out of scope here. The express server is currently one instance per region in production, so in-process serialisation removes the realistic race.

## Verification

- [x] Original regression test fails on pre-fix code with `Invalid part index: 1 !== 0`; passes with this commit
- [x] All four new regression tests pass
- [x] `yarn backend test` — 26 suites, **408 tests pass** (404 → 408 with four new regressions)
- [x] Targeted `__tests__/integration/s3-sdk` suite: 29/29 (was 25; +4 regression tests)
- [x] Re-run `scripts/live-integration/test-s3-api.sh` against the redeploy

### DIFF
--- a/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
+++ b/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
@@ -181,6 +181,78 @@ describe('AWS S3 - SDK', () => {
     )
   })
 
+  // Regression for #725: S3 multipart explicitly allows parts to arrive in any
+  // order (parallel uploads, retries, etc.) — PartNumber identifies position,
+  // not arrival order.  Pre-fix, the processing layer rejected non-sequential
+  // arrivals and the assembled object only contained whichever contiguous
+  // prefix of parts happened to arrive in order.
+  describe('Out-of-order multipart upload', () => {
+    const OutOfOrderKey = 'multipart-out-of-order.bin'
+    const PART_SIZE = 16 * 1024
+    // Distinct, deterministic per-part content so misordered assembly is
+    // detectable byte-for-byte.
+    const PartA = Buffer.alloc(PART_SIZE, 'A')
+    const PartB = Buffer.alloc(PART_SIZE, 'B')
+    const PartC = Buffer.alloc(PART_SIZE, 'C')
+    const ExpectedBody = Buffer.concat([PartA, PartB, PartC])
+
+    it('uploaded as [2,1,3] and completed as [3,1,2] should assemble to A‖B‖C', async () => {
+      const create = await s3Client.send(
+        new CreateMultipartUploadCommand({ Bucket, Key: OutOfOrderKey }),
+      )
+      const uploadId = create.UploadId!
+
+      // Upload parts in non-monotonic arrival order: 2, then 1, then 3.
+      const upPart = (partNumber: number, body: Buffer) =>
+        s3Client.send(
+          new UploadPartCommand({
+            Bucket,
+            Key: OutOfOrderKey,
+            UploadId: uploadId,
+            PartNumber: partNumber,
+            Body: body,
+          }),
+        )
+      const e2 = await upPart(2, PartB)
+      const e1 = await upPart(1, PartA)
+      const e3 = await upPart(3, PartC)
+
+      expect(e1.ETag).toMatch(MD5_ETAG_RE)
+      expect(e2.ETag).toMatch(MD5_ETAG_RE)
+      expect(e3.ETag).toMatch(MD5_ETAG_RE)
+
+      // Complete with parts listed in yet another order (3, 1, 2).  The
+      // server must sort by PartNumber regardless of either arrival order
+      // or completion-list order.
+      const complete = await s3Client.send(
+        new CompleteMultipartUploadCommand({
+          Bucket,
+          Key: OutOfOrderKey,
+          UploadId: uploadId,
+          MultipartUpload: {
+            Parts: [
+              { ETag: e3.ETag!, PartNumber: 3 },
+              { ETag: e1.ETag!, PartNumber: 1 },
+              { ETag: e2.ETag!, PartNumber: 2 },
+            ],
+          },
+        }),
+      )
+      expect(complete.ETag).toMatch(MULTIPART_ETAG_RE)
+
+      // Download and verify the assembled body is A‖B‖C in PartNumber
+      // order, not arrival or completion-list order.
+      const downloaded = await s3Client.send(
+        new GetObjectCommand({ Bucket, Key: OutOfOrderKey }),
+      )
+      const body = Buffer.from(
+        await downloaded.Body!.transformToByteArray(),
+      )
+      expect(body.length).toBe(ExpectedBody.length)
+      expect(body).toEqual(ExpectedBody)
+    }, 30_000)
+  })
+
   // Bucket-prefixed key: first segment becomes the bucket name
   describe('Bucket-prefixed keys', () => {
     const BucketedKey = 'my-archive/report.txt'
--- a/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
+++ b/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
@@ -401,6 +401,56 @@ describe('AWS S3 - SDK', () => {
       })
     }, 30_000)
 
+    // Regression for the second Bugbot finding on #726 ("stale part guard
+    // races drain"): two concurrent uploads of the SAME part_index used to
+    // both read a pre-processing cursor, both bypass the divergent-content
+    // guard, then race the store against the drain.  With the guard + store +
+    // drain now serialised under withDrainLock, concurrent identical uploads
+    // of the same part must converge to a single processed chunk and a
+    // correct assembled body (order-independent, so deterministic to assert).
+    it('concurrent identical uploads of the same part should not corrupt', async () => {
+      const KeySameIdx = 'multipart-same-index-concurrent.bin'
+      const create = await s3Client.send(
+        new CreateMultipartUploadCommand({ Bucket, Key: KeySameIdx }),
+      )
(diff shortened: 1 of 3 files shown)


jim-counter:
bugbot run

jim-counter:
bugbot run

EmilFattakhov (approved):
I like how descriptive the tests are, LGTM!

PR #728, opened 2026-05-31
fix(s3): HeadObject returns 200 with ContentLength and Last-Modified

### LINKED ISSUE #727: S3 HeadObject returns ContentLength=0 and omits Last-Modified
## Summary

`HeadObject` on a non-empty object returns `Content-Length: 0` and no `Last-Modified` header. Surfaced by the S3 live-integration test (#719):

```
HeadObject header round-trips
  ✗  HeadObject — ContentLength wrong        head=0 file=43
  ✗  HeadObject — LastModified missing from response
  ℹ  HeadObject — ContentType is '' (PUT set 'text/plain')
```

GET is unaffected — only HEAD.

### 1. HEAD responds 204, which strips content headers

`headObjectHandler` ([apps/backend/src/app/controllers/s3/s3.ts](apps/backend/src/app/controllers/s3/s3.ts)) calls the shared header helpers (which correctly `res.set('Content-Length', metadata.size)` and `res.set('Content-Type', …)`), then ends with:

```ts
res.sendStatus(204)
```

A `204 No Content` is defined as bodiless, so Node strips `Content-Length` and `Content-Type` from the response. The headers the helper set are discarded, and the AWS SDK reports `ContentLength: 0` / empty `ContentType`. S3 `HeadObject` returns **200** with full headers and an empty body. GET works because it streams a 200/206 body, so the headers survive.

The `ContentType` empty (reported as info, not fail) has the same cause and is fixed by the same change.

### 2. Last-Modified is set by neither helper

`handleDownloadResponseHeaders` / `handleS3DownloadResponseHeaders` in `@autonomys/file-server` never set `Last-Modified` (confirmed by reading the dist). It must be set by the backend from the S3 mapping's `updatedAt` (the IPLD metadata the SDK sees has no S3-mapping timestamp). `getObject` already fetches the mapping; `updatedAt` is bumped on every PUT, so it is the correct object-last-modified source.

## Fix

- `headObjectHandler`: `res.sendStatus(204)` → `res.status(200).end()`, preserving the headers the helpers set (fixes ContentLength and ContentType).
- Thread `mapping.updatedAt` through `getObject` as `lastModified` and set `Last-Modified` (RFC 7231 date) on both GET and HEAD.

## Regression test

Add a hermetic HEAD test in `apps/backend/__tests__/integration/s3-sdk/index.spec.ts` against a plain object asserting `ContentLength` equals the body length, `LastModified` is a `Date`, and `ContentType` is populated.

## Verification plan

- [x] Backend suite green
- [x] New HEAD test passes
- [x] Re-run `scripts/live-integration/test-s3-api.sh`; the two HEAD failures become `ok`, the ContentType info becomes a pass

jim-counter:
Closes #727.

## What was wrong

`HeadObject` on a non-empty object returned `Content-Length: 0` and no `Last-Modified`. From the S3 live-integration test (#719):

```
HeadObject header round-trips
  ✗  HeadObject — ContentLength wrong        head=0 file=43
  ✗  HeadObject — LastModified missing from response
  ℹ  HeadObject — ContentType is '' (PUT set 'text/plain')
```

GET was unaffected — only HEAD.

## Two root causes

**1. HEAD responded 204, stripping content headers.** `headObjectHandler` called the shared header helpers (which correctly set `Content-Length` from `metadata.size` and `Content-Type`), then ended with `res.sendStatus(204)`. A `204 No Content` is defined as bodiless, so Node strips `Content-Length` and `Content-Type` — the AWS SDK then reports `ContentLength: 0` and empty `ContentType`. GET works because it streams a 200/206 body, so the headers survive. The empty `ContentType` (reported as info) shares this cause.

**2. `Last-Modified` was set by neither helper.** `handleDownloadResponseHeaders` / `handleS3DownloadResponseHeaders` in `@autonomys/file-server` never set it (confirmed against the dist). The SDK's IPLD metadata carries no S3-mapping timestamp, so it belongs in the backend.

## Fix

- `headObjectHandler`: `res.sendStatus(204)` → `res.status(200).end()`, preserving the headers the helpers set (fixes ContentLength **and** ContentType).
- Thread `mapping.updatedAt` through `getObject` as `lastModified` and set `Last-Modified` (`toUTCString()`, RFC 7231) on both GET and HEAD. `updatedAt` is bumped on every PUT (`ON CONFLICT … updated_at = NOW()`), so it is the correct object-last-modified source.

## Test

New hermetic HEAD test against a plain object asserts `ContentLength === Body.length`, `LastModified instanceof Date`, and `ContentType` truthy. The existing HEAD test covers a compressed+encrypted object (where the stored size has different semantics), so this uses a clean plain object.

## Verification

- [x] `yarn backend test` — 26 suites, **405 tests** (404 → 405)
- [x] backend/auth/frontend lint pass
- [x] Re-run `scripts/live-integration/test-s3-api.sh`; the two HEAD failures become `ok` and the ContentType info becomes a pass

### DIFF
--- a/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
+++ b/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
@@ -432,6 +432,17 @@ describe('AWS S3 - SDK', () => {
       // CID must be present in the custom header
       expect(result.Metadata?.cid).toBeDefined()
     })
+
+    // HeadObject must return 200 with the headers GET would send. A prior
+    // 204 stripped Content-Length/Content-Type; Last-Modified was never set.
+    it('should return ContentLength, ContentType and LastModified on HeadObject', async () => {
+      // `Key`/`Body` is a plain (uncompressed, unencrypted) object.
+      const result = await s3Client.send(new HeadObjectCommand({ Bucket, Key }))
+
+      expect(result.ContentLength).toBe(Body.length)
+      expect(result.LastModified).toBeInstanceOf(Date)
+      expect(result.ContentType).toBeTruthy()
+    })
   })
 
   // Raw HTTP requests that mimic the AWS CLI / botocore, which (unlike the JS
--- a/apps/backend/src/app/controllers/s3/s3.ts
+++ b/apps/backend/src/app/controllers/s3/s3.ts
@@ -142,6 +142,7 @@ export const getObjectHandler = async (req: Request, res: Response) => {
     byteRange: resultingByteRange,
     cid,
     etag,
+    lastModified,
   } = downloadResult.value
 
   handleDownloadResponseHeaders(req, res, metadata, {
@@ -155,6 +156,7 @@ export const getObjectHandler = async (req: Request, res: Response) => {
   if (etag) res.set('ETag', etag)
   // Always expose the CID so clients that understand Autonomys can use it.
   res.set('x-amz-meta-cid', cid)
+  res.set('Last-Modified', lastModified.toUTCString())
 
   pipeline(await startDownload(), res, (err: Error | null) => {
     if (err) {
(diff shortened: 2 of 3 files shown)


EmilFattakhov (approved):
Makes sense 👍

PR #730, opened 2026-05-31
fix(s3): reject unsupported operations with 501 instead of misrouting

### LINKED ISSUE #729: S3 dispatch misroutes unsupported operations to mutating handlers
## Summary

`getS3Method` ([apps/backend/src/app/controllers/s3/http.ts](apps/backend/src/app/controllers/s3/http.ts)) selects the handler from query parameters **without considering the HTTP method**. Several operations the backend doesn't implement are therefore routed to a handler for a similarly-shaped supported request — and because the target handlers mutate state, a safe or unsupported request causes an unintended write.

Surfaced by the S3 live-integration test (#719), where these were reported as "accepted (contract may have changed)":

| Request | Misrouted to | Unintended effect |
|---|---|---|
| `GET /?uploads` — ListMultipartUploads | `CreateMultipartUpload` | creates an orphan multipart upload |
| `GET /key?uploadId` — ListParts | `CompleteMultipartUpload` | finalizes the upload |
| `DELETE /key?uploadId` — AbortMultipartUpload | `CompleteMultipartUpload` | finalizes instead of discarding |
| `PUT /key` + `x-amz-copy-source` — CopyObject | `PutObject` | writes an **empty** object at the destination |

The two GET cases are the most serious: an HTTP-safe method performs a write. CopyObject silently produces an empty object where the caller expected a copy.

## Root cause

```ts
const getS3Method = (req: Request) => {
  const { 'x-id': s3Method } = req.query
  if (!s3Method) {
    if ('uploads' in req.query) return 'CreateMultipartUpload'        // any method
    if ('uploadId' in req.query && 'partNumber' in req.query) return 'UploadPart'
    if ('uploadId' in req.query) return 'CompleteMultipartUpload'     // any method
    if (req.query['list-type'] === '2') return 'ListObjectsV2'
    if (req.method === 'HEAD') return 'HeadObject'
    if (req.method === 'DELETE') return 'DeleteObject'
    if (req.method === 'GET') return 'GetObject'
    if (req.method === 'PUT') return 'PutObject'
  }
  return s3Method
}
```

The `uploads` / `uploadId` checks fire regardless of method, so `GET ?uploads`, `GET ?uploadId`, and `DELETE ?uploadId` reach the multipart write handlers. CopyObject has no copy-source check, so it falls through to the `PUT → PutObject` branch.

(The `x-id` fast path masks this for the AWS JS SDK on some operations — it sends `x-id=CopyObject` etc., which then miss the handler config and 500 — but the AWS CLI / botocore sends no `x-id` and hits the misroute.)

## Fix

Make the resolver method-aware (switch on `req.method`, then disambiguate by query params / `x-amz-copy-source`). Operations the backend does not implement — CopyObject, ListParts, ListMultipartUploads, AbortMultipartUpload, POST-object — resolve to a dedicated `NotImplemented` handler returning **501** with an S3 `<Error><Code>NotImplemented</Code>` body and **no side effects**.

## Regression tests

Hermetic tests asserting CopyObject, ListParts, ListMultipartUploads, and AbortMultipartUpload each return 501, plus a check that CopyObject does not create the destination object (HEAD → 404 afterwards).

## Out of scope (possible future features)

CopyObject is a natural cheap operation here (a new key → same CID mapping) and AbortMultipartUpload is useful cleanup. Implementing them is a feature; this issue only stops the misrouting and rejects them cleanly.

## Verification plan

- [x] Backend suite green; new 501 tests pass
- [x] Existing supported-operation tests unaffected
- [x] Re-run `scripts/live-integration/test-s3-api.sh`; the three "accepted" probes become "rejected"

jim-counter:
Closes #729.

## What was wrong

`getS3Method` selected the handler from query parameters **without considering the HTTP method**, so operations the backend doesn't implement reached a handler for a similarly-shaped supported request. Because those handlers mutate state, a safe or unsupported request caused an unintended write:

| Request | Misrouted to | Unintended effect |
|---|---|---|
| `GET /?uploads` — ListMultipartUploads | `CreateMultipartUpload` | created an orphan multipart upload |
| `GET /key?uploadId` — ListParts | `CompleteMultipartUpload` | **finalized** the upload |
| `DELETE /key?uploadId` — AbortMultipartUpload | `CompleteMultipartUpload` | **finalized** instead of discarding |
| `PUT /key` + `x-amz-copy-source` — CopyObject | `PutObject` | wrote an **empty** object at the destination |

The two GET cases are the worst — an HTTP-safe method performing a write.

## Fix

Resolve the operation **method-first**, then disambiguate by query param / `x-amz-copy-source`. CopyObject, ListParts, ListMultipartUploads, AbortMultipartUpload, and POST-object route to a `notImplementedHandler` returning **501** with an S3 `<Error><Code>NotImplemented</Code>` body and no side effects.

Bucket-level `ListMultipartUploads` (`GET /?uploads`) lands on the separate `GET /` ListBuckets route when the bucket is the endpoint (so it returned the bucket list, not a write — but still wrong); guarded there too.

## Tests

Hermetic tests assert 501 for CopyObject, ListParts, ListMultipartUploads, and AbortMultipartUpload — including that CopyObject creates no destination object, and that ListParts/Abort run against a **live in-progress upload** so a misroute would actually finalize it (proving it no longer does).

## Notes

- CopyObject (a new key→same-CID mapping) and AbortMultipartUpload (cleanup) are natural cheap features here; this PR only stops the misrouting and rejects them cleanly. Worth separate feature issues if desired.
- The CopyObject "no side effect" check asserts the destination HEAD *rejects* rather than a specific status, since the missing-key 404 vs 500 behaviour is fixed independently in #724.

## Verification

- [x] `yarn backend test` — 26 suites, **408 tests** (404 → 408 with 4 new)
- [x] backend/auth/frontend lint pass
- [x] Re-run `scripts/live-integration/test-s3-api.sh`; the CopyObject / ListMultipartUploads / ListParts probes flip from "accepted" to "rejected"

### DIFF
--- a/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
+++ b/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
@@ -1,12 +1,16 @@
 import { dbMigration } from '../../utils/dbMigrate.js'
 import {
+  AbortMultipartUploadCommand,
   CompleteMultipartUploadCommand,
+  CopyObjectCommand,
   CreateMultipartUploadCommand,
   DeleteObjectCommand,
   GetObjectCommand,
   HeadObjectCommand,
   ListBucketsCommand,
+  ListMultipartUploadsCommand,
   ListObjectsV2Command,
+  ListPartsCommand,
   PutObjectCommand,
   S3Client,
   UploadPartCommand,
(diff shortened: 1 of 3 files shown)


jim-counter:
bugbot run

EmilFattakhov (approved):
Looks good!

PR #732, opened 2026-06-04
Fix/706 confirmation depth publishing

### LINKED ISSUE #706: # Phantom Published Nodes: Transactions Recorded as Published But Never Included On-Chain
Follow up in #700 

## Summary

1,309 nodes in the database have `block_published_on` and `tx_published_on` set, indicating they were published on-chain, but **the transactions were never actually included in the claimed blocks**. These phantom nodes block the reconciliation job from making any progress, preventing 13,006 legitimately published (and indexer-resolvable) nodes from being reconciled.

## Impact

- **164 objects** are affected (all showing `is_archived = false`), belonging to uploads across 5 date windows
- **Reconciliation is completely stalled** — the job has made 0 progress for 8+ consecutive runs and counting
- **13,006 older nodes** (blocks 1.7M–5.5M) that ARE on-chain and resolvable by the indexer are permanently blocked from reconciliation
- Affected objects will never reach `is_archived = true` status

### How `block_published_on` is set prematurely

In `src/infrastructure/services/upload/onchainPublisher/transactionManager.ts`, line 70:

```typescript
if (status.isInBlock || status.isFinalized) {
```

When a transaction reaches `isInBlock` status, the code immediately records it as published:

```typescript
// line 91-100
logger.info('In block: %s', status.asInBlock.toString())
const blockHash = status.asInBlock.toString()
const { block } = await api.rpc.chain.getBlock(blockHash)
resolve({
  success: true,
  txHash: transaction.hash.toString(),
  blockHash: status.asInBlock.toString(),
  blockNumber: block.header.number.toNumber(),
  status: status.type,
})
```

The `isInBlock` status in Substrate means the transaction was included in a **non-finalized** block. If a chain reorganization (fork) occurs, that block can be replaced by a different block that does **not** include the transaction. When this happens:

1. The database records `block_published_on = <block number>` and `tx_published_on = <tx hash>`
2. The block gets reorged — the transaction is silently dropped
3. The transaction is never re-submitted (the system considers it "done")
4. The node is permanently stuck: marked as published, but the data is not on-chain

### Why reconciliation is completely blocked

The reconciliation job (`src/core/objects/reconciliation.ts`) queries unreconciled nodes ordered by `ORDER BY m.created_at DESC LIMIT 500` — newest first. All 1,309 phantom nodes are newer than the 13,006 legitimately stuck nodes. Since the phantom nodes always fill the 500-node batch, and the indexer correctly reports "not found" for all of them (the data was never actually archived), the job resolves 0/500 every run and never reaches the older, resolvable backlog.

### 1. Initial observation

The reconciliation job logs showed zero progress for 8+ consecutive runs:

```
Reconciliation batch complete: resolved 0/500 nodes, 0 objects affected
Reconciliation made no progress for 8 consecutive runs (500 unreconciled nodes).
```

### 2. Database state

| Metric                  | Count   |
|-------------------------|---------|
| Total published nodes   | 251,407 |
| Reconciled (have piece) | 237,089 (94.3%) |
| Unreconciled (stuck)    | 14,318 (5.7%)   |

### 3. Random sampling revealed indexer has the data

Sampled 500 **random** unreconciled nodes against the indexer:

- **457 (91.4%) FOUND** in indexer
- 43 (8.6%) not found

This proved the indexer is working and has data — the job should be resolving most nodes.

### 4. Reconciliation's "newest first" query only hits recent nodes

The 500 newest unreconciled nodes (what the reconciliation job actually queries) are all from blocks 7,226,757–7,845,279 (only 39 distinct blocks). Testing these against the indexer:

- **0 out of 500 found** — the indexer has no data for any of them

### 5. Age distribution confirmed two populations

| Category         | Count  | Block Range       |
|------------------|--------|-------------------|
| Old (resolvable) | 13,006 | 1,706,609–5,525,681 |
| Recent (phantom) | 1,309  | 7,226,757–7,845,279 |

### 6. On-chain verification proved transactions don't exist

Exhaustively verified all 1,309 distinct (block, tx) pairs:

- **0 out of 1,309 transactions found** in their claimed blocks
- All 39 blocks were checked; the extrinsics in each block do not include the claimed transaction hashes

For comparison, 30 randomly sampled old unreconciled nodes: **30 out of 30 confirmed on-chain**.

### 7. Root cause traced to `isInBlock` vs `isFinalized`

The `transactionManager.ts` resolves with `success: true` on `isInBlock`, which is susceptible to chain reorganizations. The phantom nodes were created across 5 date windows suggesting 5 separate reorg events:

| Date Window          | Blocks              | Phantom Nodes | Objects |
|----------------------|----------------------|---------------|---------|
| 2026-04-01 06:58–07:15 | 7,226,757–7,226,971 | 435 | 2 |
| 2026-04-02 02:32–05:31 | 7,238,360–7,240,034 | 366 | 158 |
| 2026-04-17 16:45     | 7,458,231           | 40 | 1 |
| 2026-05-14 12:10     | 7,836,513–7,836,686 | 169 | 1 |
| 2026-05-15 01:57     | 7,844,820–7,845,279 | 138 | 1 |

## Affected Objects Detail

Top affected objects (by phantom node count):

| head_cid (truncated) | Total Nodes | Archived | Phantom | Unpublished |
|---|---|---|---|---|
| `bafkr6ifmhxu...` | 355 | 80 | 275 | 0 |
| `bafkr6ih223e...` | 384 | 160 | 224 | 0 |
| `bafkr6ifdmox...` | 879 | 610 | 169 | 100 |
| `bafkr6iec43l...` | 1,279 | 1,119 | 160 | 0 |
| `bafkr6ibp7bw...` | 1,214 | 1,026 | 138 | 50 |

Several objects have a mix of legitimately archived nodes and phantom nodes, meaning parts of those files are on-chain but other parts are lost.

### Immediate: Clear phantom nodes for re-publishing

Reset `block_published_on` and `tx_published_on` to NULL for the 1,309 phantom nodes so they re-enter the publishing pipeline:

```sql
-- DRY RUN: verify count first
SELECT COUNT(*) FROM nodes
WHERE block_published_on >= 7200000
  AND piece_index IS NULL
  AND tx_published_on IS NOT NULL;
-- Expected: 1309

-- EXECUTE: clear phantom publication data
UPDATE nodes
SET block_published_on = NULL,
    tx_published_on = NULL
WHERE block_published_on >= 7200000
  AND piece_index IS NULL
  AND tx_published_on IS NOT NULL;
```

### Preventive: Wait for finalization before recording publication

Change `transactionManager.ts` to only resolve on `isFinalized` instead of `isInBlock`:

```typescript
// Before (vulnerable to reorgs):
if (status.isInBlock || status.isFinalized) {

// After (safe):
if (status.isFinalized) {
```

Note: this increases latency per transaction (finalization takes ~1 minute on mainnet vs ~6 seconds for block inclusion). Consider whether the tradeoff is acceptable, or whether a hybrid approach is better (record on `isInBlock`, verify on `isFinalized`, revert if not finalized).

### Secondary: Make reconciliation resilient to permanently unresolvable nodes

Even with the finalization fix, nodes could become unresolvable for other reasons. The reconciliation job should have a mechanism to skip nodes that have repeatedly failed resolution (e.g., after N consecutive failures across runs, flag them for manual review and exclude from the query).

EmilFattakhov:
## Summary

Fixes #706. On-chain publishing recorded a node as published the moment its transaction reached `isInBlock`. Because Autonomys uses Nakamoto-style (probabilistic) consensus, a non-finalized block can be dropped by a chain reorg — so transactions were marked `block_published_on`/`tx_published_on` in the DB while the data never actually landed on-chain. This produced ~1,309 "phantom" published nodes that stalled the reconciliation job and blocked 13,006 legitimately-stuck nodes from being reconciled.

This PR makes publication durable: a transaction is only recorded as published after `CHAIN_CONFIRMATION_DEPTH` blocks build on top of its inclusion block **and** that inclusion block is still canonical.

## Approach

Rather than waiting for GRANDPA `isFinalized` (Autonomys finality is probabilistic/k-deep, so this is slow and not the right signal), we count best-chain confirmations — the same N-confirmation pattern already used by the EVM `paymentManager`:

1. On `isInBlock`, capture the inclusion block hash + number — but do **not** resolve.
2. Subscribe to new heads; once the chain is `confirmationDepth` blocks past the inclusion block, re-fetch the canonical block hash at the inclusion height.
3. If it matches the captured hash → the transaction survived → resolve `success: true`.
4. If it differs → the block was reorged out → resolve `success: false` so the node re-enters the publishing pipeline instead of being silently lost.

The default depth is **25 blocks** (~2.5 min at ~6s/block) — over 2x the largest reorg ever observed on the network (~12 blocks).

## Changes

- **`transactionManager.ts`** — wait for confirmation depth + canonical re-check before resolving success; replace the hardcoded 60s timeout with a configurable budget; tear down both the extrinsic and new-heads subscriptions on every exit path.
- **`accounts.ts`** — add `resyncAccount()` which re-fetches the on-chain nonce and keeps the account in the pool (instead of removing it).
- **Reorg/transient handling** — reorg, timeout, and usurped outcomes are treated as chain/timing events, not account faults: they `resyncAccount` (resync nonce, keep account) rather than `removeAccount`. This prevents the signer pool from draining during the very congestion/reorgs this change exists to survive, and keeps in-memory nonces in sync with chain state.
- **`config.ts` / `.env.sample`** — new env-configurable settings.

## Configuration

```
# Reorg safety margin. A node is only recorded as published once this many
# blocks build on its inclusion block AND that block is still canonical.
# Default 25 (~2.5 min at 6s/block). Clamped to >= 1.
CHAIN_CONFIRMATION_DEPTH=25
# Max time a single transaction may wait to be confirmed. Must cover both
# time-to-inclusion (congestion) and confirmationDepth blocks on top.
CHAIN_TRANSACTION_TIMEOUT_MS=300000
```

## Testing

- `tsc --noEmit` — clean
- `eslint` on changed files — clean
- Reviewed by two independent senior-engineer passes; both critical findings (signer-pool drain + nonce desync on reorg) resolved and verified.

## Trade-offs / operational notes

- **Throughput:** each transaction now holds its concurrency slot for the full confirmation window (~2.5 min) instead of ~6s, lowering steady-state publish throughput. If clearing the backlog bottlenecks, a follow-up can decouple confirmation-waiting from the submission concurrency limit (confirm out-of-band, free the slot at inclusion).
- **Latency:** publication is acknowledged ~2.5 min after inclusion by design — this is the cost of reorg safety and is tunable via `CHAIN_CONFIRMATION_DEPTH`.

### DIFF
--- a/apps/backend/.env.sample
+++ b/apps/backend/.env.sample
@@ -6,6 +6,21 @@ AUTH_SERVICE_API_KEY=1234567890 # change to your own if updated in auth
 RABBITMQ_URL=amqp://guest:guest@localhost:5672
 
 # ---------------------------------------------------------------------------
+# On-chain publishing (consensus chain)
+# ---------------------------------------------------------------------------
+
+# Reorg safety margin for publishing. A transaction is only recorded as
+# published once this many blocks have been built on top of its inclusion
+# block AND that block is still canonical. Prevents "phantom" published nodes
+# from chain reorgs (see issue #706). Default 25 (~2.5 min at 6s/block).
+CHAIN_CONFIRMATION_DEPTH=25
+
+# Max time a single transaction may wait to reach CHAIN_CONFIRMATION_DEPTH
+# before timing out. Must exceed CHAIN_CONFIRMATION_DEPTH * block time with
+# margin. Default 300000ms (5 min).
+CHAIN_TRANSACTION_TIMEOUT_MS=300000
+
+# ---------------------------------------------------------------------------
 # Pay-with-AI3 / purchased credits feature
 # ---------------------------------------------------------------------------
 
--- a/apps/backend/__tests__/unit/onchainPublisher/confirmation.spec.ts
+++ b/apps/backend/__tests__/unit/onchainPublisher/confirmation.spec.ts
@@ -1,0 +1,59 @@
+import { describe, it, expect } from '@jest/globals'
+import {
+  hasReachedConfirmationDepth,
+  isStillCanonical,
+  isAccountFault,
+} from '../../../src/infrastructure/services/upload/onchainPublisher/confirmation.js'
+
+describe('confirmation helpers', () => {
+  describe('hasReachedConfirmationDepth', () => {
+    it('is false until the head reaches inclusion + depth', () => {
+      // inclusion at 100, depth 25 -> need head >= 125
+      expect(hasReachedConfirmationDepth(124, 100, 25)).toBe(false)
+      expect(hasReachedConfirmationDepth(110, 100, 25)).toBe(false)
+    })
+
+    it('is true exactly at inclusion + depth', () => {
+      expect(hasReachedConfirmationDepth(125, 100, 25)).toBe(true)
+    })
+
+    it('is true beyond inclusion + depth', () => {
+      expect(hasReachedConfirmationDepth(200, 100, 25)).toBe(true)
+    })
+
+    it('still requires one confirmation at minimum depth of 1', () => {
+      expect(hasReachedConfirmationDepth(100, 100, 1)).toBe(false)
+      expect(hasReachedConfirmationDepth(101, 100, 1)).toBe(true)
+    })
+  })
+
+  describe('isStillCanonical', () => {
+    it('is true when the canonical hash at the inclusion height is unchanged', () => {
+      expect(isStillCanonical('0xabc', '0xabc')).toBe(true)
+    })
+
+    it('is false when a reorg replaced the inclusion block', () => {
+      expect(isStillCanonical('0xdef', '0xabc')).toBe(false)
+    })
+  })
+
+  describe('isAccountFault', () => {
+    it('treats only Invalid as an account fault that evicts the signer', () => {
+      expect(isAccountFault('Invalid')).toBe(true)
+    })
+
+    it('treats chain/infrastructure failures as transient (no eviction)', () => {
+      for (const status of [
+        'Reorged',
+        'Timeout',
+        'Usurped',
+        'ConfirmationError',
+        'Failed',
+        'InBlock',
+        undefined,
+      ]) {
+        expect(isAccountFault(status)).toBe(false)
+      }
+    })
+  })
+})
--- a/apps/backend/__tests__/unit/utils/misc.spec.ts
+++ b/apps/backend/__tests__/unit/utils/misc.spec.ts
@@ -1,0 +1,39 @@
+import { describe, it, expect, afterEach } from '@jest/globals'
+import { positiveIntEnv } from '../../../src/shared/utils/misc.js'
+
+const KEY = 'TEST_POSITIVE_INT_ENV'
+
+describe('positiveIntEnv', () => {
+  afterEach(() => {
+    delete process.env[KEY]
+  })
+
+  it('returns the fallback when the variable is unset', () => {
+    expect(positiveIntEnv(KEY, 25)).toBe(25)
+  })
+
+  it('parses a valid positive integer', () => {
+    process.env[KEY] = '50'
+    expect(positiveIntEnv(KEY, 25)).toBe(50)
+  })
+
+  it('floors a decimal value', () => {
+    process.env[KEY] = '12.9'
+    expect(positiveIntEnv(KEY, 25)).toBe(12)
+  })
+
+  it('falls back on a non-numeric value (NaN guard)', () => {
+    process.env[KEY] = 'not-a-number'
+    expect(positiveIntEnv(KEY, 25)).toBe(25)
+  })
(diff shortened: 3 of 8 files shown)


EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

jim-counter (dismissed):
LGTM. Have we considered adding test coverage for this behaviour?

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM. Thanks for the tests and the refactor to support them!

PR #735, opened 2026-06-08
fix(frontend): preview compressed non-encrypted files

EmilFattakhov:
Compressed (non-encrypted) files failed to preview because the public gateway serves them with `Content-Encoding: deflate`, which browsers fail to decode (broken  / ERR_CONTENT_DECODING_FAILED).

Route compressed previews through the backend download API (ignoreEncoding=true + client-side decompress) instead of the gateway URL, and pass gatewayUrl=null for those files so the previewer uses the decompressed blob. Adds a raw-bytes fallback for files whose stored data isn't actually compressed despite ZLIB metadata, and prevents a superseded preview run from clobbering state with a stale error.


EmilFattakhov:
At first I wanted to roll out both frontend change (defensive) and a gateway change, but I believe the pure gateway fix in PR https://github.com/autonomys/auto-files-gateway/pull/171 is clearer, closing this PR

Issue #736, opened 2026-06-08
Previews for compressed non-encrypted files are still flaky


EmilFattakhov:
# previews for compressed non-encrypted files are still broken

## Summary
Previews (and direct fetches via the public file endpoint) for files that are **compressed but not encrypted** are still failing in the browser. The public gateway responds with `Content-Encoding: deflate` while the body on the wire is **uncompressed** (raw file bytes), so the browser tries to inflate plaintext and fails (`ERR_CONTENT_DECODING_FAILED` / broken `<img>`).

This persists even after deploying `@autonomys/file-server@1.6.14` to the relevant services.

## Impact
- File previews for affected objects render broken in the web UI.
- Direct `GET /file/:cid` of these objects is undecodable by any client that honors `Content-Encoding`.

## Reproduction
```bash
CID=bafkr6ia5gftwkofm6vqu36qen3ps23ha63h5ucwjnro6d2orj4p7p7srq4

# Browser-like subresource request (what an <img> sends)
curl -sS -D - -o /dev/null \
  -H "Accept-Encoding: gzip, deflate, br" \
  -H "Sec-Fetch-Dest: image" \
  "https://gateway.autonomys.xyz/file/$CID"
```

**Actual:** `HTTP/2 200`, `content-encoding: deflate`, but the body begins with the PNG magic number `89 50 4E 47 0D 0A 1A 0A` — i.e. raw, *uncompressed* bytes served under a `deflate` label.

**Expected:** no `Content-Encoding` header (server serves plain/decompressed bytes), browser renders the preview.

Additional observations against the live endpoint:
- `Content-Encoding: deflate` is only set when the request includes `Accept-Encoding: deflate` (i.e. it negotiates purely on `Accept-Encoding`).
- `?raw=true` and `?ignoreEncoding=true` are **ignored** (still returns `Content-Encoding: deflate`).
- Response has no `x-file-origin` header.

## Root cause (understood)
Some stored objects carry `compression: ZLIB` metadata while their node bytes are actually uncompressed (upload-side metadata/data mismatch). Trusting the flag makes the service advertise `Content-Encoding: deflate` over plaintext, which breaks any client that honors the encoding.

## The fix that should address this (already in code)
[...]

PR #737, opened 2026-06-08
fix(frontend): preview compressed non-encrypted files

EmilFattakhov:
Compressed (non-encrypted) files failed to preview because the public gateway serves them with Content-Encoding: deflate, which browsers fail to decode (broken  / ERR_CONTENT_DECODING_FAILED). All necessary fixes are implemented on the gateway side, nevertheless, the decoding is still affected. 

Make previews transport-agnostic so they render regardless of how the gateway/backend labels the body:

- Keep a zero-cost fast path: plain, directly-renderable media still streams straight from the gateway URL (native range support, nothing buffered into memory).
- Route everything else (compressed, encrypted, or text) through one defensive path that fetches the raw stored bytes (ignoreEncoding=true + skipDecryption) and decrypts/decompresses client-side, never trusting the wire encoding.
- Sniff the zlib header before inflating, so files flagged ZLIB but stored uncompressed pass through instead of throwing — no second fetch, no exception-driven fallback.
- Decrypt via decryptFile directly to avoid the SDK's decryptFileData always-decompressing
- Pass gatewayUrl=null for compressed non-encrypted files so the previewer uses the decoded blob, and prevent a superseded preview run from clobbering state with a stale error.

Addresses the front-end part of: https://github.com/autonomys/auto-drive/issues/736

### DIFF
--- a/apps/frontend/__tests__/unit/components/filePreviewDecode.spec.ts
+++ b/apps/frontend/__tests__/unit/components/filePreviewDecode.spec.ts
@@ -250,6 +250,26 @@ describe('loadPlaintextBytes', () => {
     ).rejects.toThrow('Aborted');
   });
 
+  it('aborts when the signal fires after the final chunk but before return', async () => {
+    const controller = new AbortController();
+    // The abort lands once the stream has fully drained — after the last chunk
+    // is yielded but before collectStream assembles/returns the buffer. The
+    // loop-top guard never sees it, so only a post-loop re-check throws.
+    const api = {
+      downloadObject: jest.fn().mockResolvedValue(
+        (async function* () {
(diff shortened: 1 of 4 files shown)


EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM!

PR #740, opened 2026-06-11
Admin credits: dedicated screen, combined refunds, mandatory refund tx hash

EmilFattakhov:
### Summary

Expands the existing "mark as refunded" feature for credits purchased with AI3 that expired unused:

1. **Dedicated credits screen** — purchased credits move out of the admin dashboard into `/{chain}/drive/admin/credits`. Batches are grouped by account, each rendered as a single summary line (batch count, storage purchased / remaining / expired unused) with a per-account toggle to expand the full batch list, plus Expand all / Collapse all.
2. **Combined refunds** — multiple unused batches of the same account can be selected and marked as refunded in one atomic operation, while single-batch refunds still work as before.
3. **Mandatory refund transaction hash** — every refund must record the on-chain tx hash of the AI3 transfer (strict `0x` + 64 hex format). Without a valid hash the request is rejected and nothing is marked refunded.

### Backend

- New migration `20260611000000` adds `purchased_credits.refund_tx_hash VARCHAR(255)` (NULL only on rows refunded before this column existed).
- `POST /credits/batches/:id/refund` now requires `{ refundTxHash }` in the body; 400 on missing/malformed hash.
- New `POST /credits/batches/refund` with `{ batchIds, refundTxHash }`:
  - All-or-nothing transaction: if any id is unknown → 404 listing the missing ids, nothing updated.
  - **Rejects batches spanning multiple accounts (400)** — one on-chain transfer goes to a single wallet, so cross-account combined refunds are blocked server-side.
  - Already-refunded batches are skipped (idempotent), same as the single endpoint; the same hash is recorded on every batch.
- Repository: `markAsRefunded(id, refundTxHash)` and new `markManyAsRefunded(ids, refundTxHash)` (row-locking transaction with rollback on missing ids / multiple accounts).
- `@auto-drive/models`: `PurchasedCredit.refundTxHash` added.

### Frontend

- New route + page `/{chain}/drive/admin/credits` (`ROUTES.adminCredits`); the dashboard section is replaced by a link card.
- Per-user purchase history: checkboxes on non-refunded batches, select-all, and a "Mark N Refunded" action; both single and combined refunds go through a modal that requires a validly-formatted tx hash before the confirm button enables.
- Refunded rows display the refund timestamp and tx hash on both screens.
- Account ID on the purchase history header and an "Account →" shortcut on each group line link to the existing account details page (uploads/downloads) via the `accountId` already on batch rows.

### Tests

13 new unit tests in `apps/backend/__tests__/unit/useCases/credits.spec.ts` covering: admin-only access, missing/malformed tx hashes (verifying the repository is never called), hash trimming, 404 on unknown ids, multi-account rejection, idempotent no-ops, id de-duplication, and refunded/already-refunded counts.

### Deployment notes

- Run `yarn backend db-migrate up` before deploying.
- Existing refunded rows keep `refund_tx_hash = NULL`; all new refunds require a hash.

### DIFF
--- a/apps/backend/migrations/sqls/20260611000000-purchased-credits-refund-tx-hash-up.sql
+++ b/apps/backend/migrations/sqls/20260611000000-purchased-credits-refund-tx-hash-up.sql
@@ -1,0 +1,11 @@
+-- Add refund transaction hash tracking to purchased_credits.
+--
+-- refund_tx_hash: On-chain transaction hash of the AI3 refund transfer the
+--                 admin executed before marking the batch as refunded.
+--                 Mandatory for every new refund — the refund endpoints
+--                 reject requests without it. Multiple batches refunded in
+--                 a single on-chain transaction share the same hash.
+--                 NULL only on rows refunded before this column existed.
+
+ALTER TABLE purchased_credits
+  ADD COLUMN refund_tx_hash VARCHAR(255);
--- a/apps/backend/src/core/users/credits.ts
+++ b/apps/backend/src/core/users/credits.ts
@@ -1,4 +1,10 @@
-import { PurchasedCredit, User, UserRole, UserWithOrganization } from '@auto-drive/models'
+import {
+  EVM_TX_HASH_REGEX,
+  PurchasedCredit,
+  User,
+  UserRole,
+  UserWithOrganization,
+} from '@auto-drive/models'
 import {
   AdminCreditBatchRow,
   AdminUserCreditBatchRow,
@@ -187,10 +193,9 @@ const getUserBatches = async (
 // ---------------------------------------------------------------------------
 // Refund transaction hash validation
 // Every refund must record the on-chain transaction hash of the AI3 transfer
-// the admin executed. Strict EVM format: 0x followed by 64 hex characters.
+// the admin executed. Strict EVM format (EVM_TX_HASH_REGEX from
+// @auto-drive/models, shared with the frontend refund modal).
 // ---------------------------------------------------------------------------
-
-const REFUND_TX_HASH_REGEX = /^0x[0-9a-fA-F]{64}$/
 
 // purchased_credits.id is a UUID column. Validating ids here returns a clean
 // 400 instead of letting Postgres fail the `::uuid` cast (which would
@@ -209,7 +214,7 @@ const validateRefundTxHash = (
     )
   }
   const trimmed = refundTxHash.trim()
-  if (!REFUND_TX_HASH_REGEX.test(trimmed)) {
+  if (!EVM_TX_HASH_REGEX.test(trimmed)) {
     return err(
       new BadRequestError(
         'Invalid refund transaction hash: expected 0x followed by 64 hex characters',
--- a/apps/backend/src/core/users/credits.ts
+++ b/apps/backend/src/core/users/credits.ts
@@ -192,6 +192,12 @@ const getUserBatches = async (
 
 const REFUND_TX_HASH_REGEX = /^0x[0-9a-fA-F]{64}$/
 
+// purchased_credits.id is a UUID column. Validating ids here returns a clean
+// 400 instead of letting Postgres fail the `::uuid` cast (which would
+// surface as a 500). Any UUID version is accepted.
+const UUID_REGEX =
+  /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i
+
 const validateRefundTxHash = (
   refundTxHash: unknown,
 ): Result<string, BadRequestError> => {
@@ -231,6 +237,10 @@ const refundBatch = async (
 ): Promise<Result<void, ForbiddenError | NotFoundError | BadRequestError>> => {
   if (executor.role !== UserRole.Admin) {
     return err(new ForbiddenError('Admin access required'))
+  }
+
+  if (!UUID_REGEX.test(batchId)) {
+    return err(new BadRequestError('Invalid batch id: expected a UUID'))
   }
 
   const txHashResult = validateRefundTxHash(refundTxHash)
@@ -306,6 +316,15 @@ const refundBatches = async (
   }
 
   const uniqueIds = [...new Set(batchIds.map((id) => id.trim()))]
+
+  const invalidIds = uniqueIds.filter((id) => !UUID_REGEX.test(id))
+  if (invalidIds.length > 0) {
+    return err(
+      new BadRequestError(
+        `Invalid batch ids (expected UUIDs): ${invalidIds.join(', ')}`,
+      ),
+    )
+  }
 
   const txHashResult = validateRefundTxHash(refundTxHash)
   if (txHashResult.isErr()) {
--- a/apps/backend/src/core/users/credits.ts
+++ b/apps/backend/src/core/users/credits.ts
@@ -6,7 +6,11 @@ import {
 } from '../../infrastructure/repositories/users/purchasedCredits.js'
 import { AccountsUseCases } from './accounts.js'
 import { config } from '../../config.js'
-import { ForbiddenError, NotFoundError } from '../../errors/index.js'
+import {
+  BadRequestError,
+  ForbiddenError,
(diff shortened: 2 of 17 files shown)


EmilFattakhov:
A few screenshots of a new interface:

Screenshot 2026-06-11 at 10 44 28 AM

Screenshot 2026-06-11 at 10 47 07 AM

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

apps/backend/migrations/sqls/20260611000000-purchased-credits-refund-tx-hash-up.sql
```diff
@@ -0,0 +1,11 @@
...
+--
+-- refund_tx_hash: On-chain transaction hash of the AI3 refund transfer the
+--                 admin executed before marking the batch as refunded.
+--                 Mandatory for every new refund — the refund endpoints
+--                 reject requests without it. Multiple batches refunded in
+--                 a single on-chain transaction share the same hash.
+--                 NULL only on rows refunded before this column existed.
+
+ALTER TABLE purchased_credits
+  ADD COLUMN refund_tx_hash VARCHAR(255);
```
jim-counter, line 11: So a single refund tx can cover multiple credit refunds. I assume the intended logic is that we can only refund the entire remaining storage balance on a purchsed_credit based on the amount of AI3 originally paid and the refund AI3 being a pro-rated mount relative to the amount of storage purchased/used?

apps/frontend/src/components/views/AdminPanel/AdminCredits.tsx
```diff
@@ -120,40 +133,166 @@ const OverCapPanel = ({
 };
 
 // ---------------------------------------------------------------------------
-// All credit batches table
+// Per-account batch group
```
jim-counter, line 136: The grouping needs to be by purchasing wallet address?

apps/frontend/src/components/views/AdminPanel/AdminUserCredits.tsx
```diff
@@ -56,29 +63,111 @@ export const AdminUserCredits = ({
...
 
+  // Batches are loaded by userPublicId, so they can span more than one
+  // account when the user purchased credits under different org accounts
+  // over time. Each distinct account links to the existing account
+  // (organization) page with the uploads/downloads history.
+  const accountIds = useMemo(
+    () => [...new Set(batches.map((b) => b.accountId))],
+    [batches],
+  );
+  const hasMultipleAccounts = accountIds.length > 1;
```
jim-counter, line 88: In a world of organisations with potentially multiple users each with multiple payment wallets the use of "Accounts" here is hard to parse. "PurchasingWallet" would improve readability.

jim-counter: Check the comment below for an example :)

apps/frontend/src/components/views/AdminPanel/RefundTxHashModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { useEffect, useState } from 'react';
+import { Button } from '@auto-drive/ui';
+import { AlertTriangle, X } from 'lucide-react';
+
+/**
+ * Strict EVM transaction hash format: 0x followed by 64 hex characters.
+ * Mirrors the backend validation — the backend rejects anything else and
+ * the batch is NOT marked as refunded.
+ */
+export const REFUND_TX_HASH_REGEX = /^0x[0-9a-fA-F]{64}$/;
```
jim-counter: This is the 3rd time I've seen this declared. Could `EVM_TX_HASH_REGEX` live as a single util somewhere?

apps/backend/src/core/users/credits.ts
```diff
@@ -216,6 +262,91 @@ const refundBatch = async (
...
+    return err(new ForbiddenError('Admin access required'))
+  }
+
+  if (
+    !Array.isArray(batchIds) ||
+    batchIds.length === 0 ||
+    !batchIds.every((id) => typeof id === 'string' && id.trim() !== '')
+  ) {
+    return err(
+      new BadRequestError('batchIds must be a non-empty array of batch ids'),
```
jim-counter, line 320: What about if the non-empty strings are invalid UUIDs?

apps/backend/src/infrastructure/repositories/users/purchasedCredits.ts
```diff
@@ -587,6 +592,110 @@ const markAsRefunded = async (
...
+    // two overlapping combined-refund transactions cannot deadlock by
+    // locking the same rows in opposite orders.
+    const existing = await client.query<{
+      id: string
+      account_id: string
+      refunded_at: Date | null
+    }>(
+      `SELECT id, account_id, refunded_at
+       FROM purchased_credits
+       WHERE id = ANY($1::uuid[])
```
jim-counter: If `ids` has any invalid format UUIDs this blows up and we'll see 500 instead of 404 as expected.

jim-counter (changes requested):
How do we calculate the refund amount? Looks like my EVM wallet has been downgraded from admin so I can't see the feature on the Netlify preview to check the flow for real :(

apps/backend/migrations/sqls/20260611000000-purchased-credits-refund-tx-hash-up.sql
EmilFattakhov: > So a single refund tx can cover multiple credit refunds. 

Yes, we batch them all expired credits together and one refund can cover several purchased credit batches. The corresponding AI3 amount is calculated separately and is based on the original purchase price in AI3. If purchased credits are partially used and partially expired (e.g. purchased 1GB used half of it), we refund half of the original paid AI3 amount, so this

> refund AI3 being a pro-rated mount relative to the amount of storage purchased/used

is exactly right.

apps/backend/src/core/users/credits.ts
EmilFattakhov: Good catch! Fixed:  both refund use cases now validate ids as UUIDs

apps/backend/src/infrastructure/repositories/users/purchasedCredits.ts
EmilFattakhov: Fixed across two layers!

apps/frontend/src/components/views/AdminPanel/AdminUserCredits.tsx
EmilFattakhov: Hm, let me implement the changes and see if they make sense in the context :)

apps/frontend/src/components/views/AdminPanel/AdminCredits.tsx
EmilFattakhov: Good idea - I introduced some additional changes to that bit, we not group by Account first, and within account we also subgroup by `fromWallet` (or `purchasingWallet`) to only refund the transactions to the same wallet that was initially used for the payment. Nice improvement!

apps/frontend/src/components/views/AdminPanel/RefundTxHashModal.tsx
EmilFattakhov: Done 👍

EmilFattakhov (commented):
Thanks for a thorough review Jim! I'm not sure where we specify the admin wallet for previews, but we definitely need to add you there 😄

EmilFattakhov:
bugbot run

apps/frontend/src/components/views/AdminPanel/AdminUserCredits.tsx
jim-counter: The `refundGroupKey` replaces most of the ambiguous references and the comment at the top makes everything clearer for anyone reading the code later. Thanks!

jim-counter (approved):
LGTM! This will make the refund process much easier. Thanks for the fixes.

Issue #747, opened 2026-06-15
Pricing: createIntent computes & locks USDC quote when method=USDC


EmilFattakhov:
## Context
Reuse the existing 10-minute price-lock pattern. `getPrice()` already yields shannons/byte
(AI3-denominated). For USDC intents, convert to a USD-denominated quote and lock it on the
intent. Credit math at confirmation stays proportional.

> **Updated 2026-08-07** — the rate source changed again (#746): AI3/USD is now the
> **volume-weighted average of the last N pool swaps**, read from the subgraph, not a v4
> Quoter result for this specific size. The **structure** of this issue is unchanged for the
> third time running: USDC converts to AI3 at the boundary and reuses
> `paymentAmount / shannons_per_byte`. What changes is where the rate comes from and which
> failures can occur. Superseded wording kept in ~~strikethrough~~.
>
> **Updated 2026-08-06** — revised after #807 replaced the CEX-aggregate oracle with the
> Uniswap v4 WAI3/USDC pool.

## The rate is now size-independent — but the effective rate column stays

~~`getExecutableQuote(shannons)` prices a specific size, because slippage is a function of
the amount.~~ A VWAP over recent trades is one number, the same for a \$5 purchase and a
\$500 one. Two consequences:

1. **The marginal-vs-executable gap is gone.** It existed because the treasury swapped at
   quote time and paid fee + price impact the marginal rate did not show. Conversion is now
   manual and batched (#751), so there is no per-intent swap to under-price. The realized
   swap prices the VWAP is built from already carry the fee and impact of *those* trades.
2. **`USD_QUOTE_MARGIN` is now the only wedge** between the displayed rate and the charged
   rate — and it is still a wedge, so the effective-rate column is still required. Convert a
   received payment at `usd_rate_at_creation` and the margin is handed back as free storage
   on every purchase.

So: persist **both**. `usd_rate_at_creation` = the raw VWAP, for display and reconciliation.
The effective rate = margin-inclusive, and it is what confirmation converts at.

`requestedBytes` (#808) is still **required** on the USDC path — not because the rate depends
on size any more, but because a USDC charge is a fixed amount and you cannot compute one
without knowing what is being bought.

## Tasks
- [ ] Migration: add an effective-rate column to `intents` (e.g. `effective_shannons_per_token`
      `numeric(78,0)`, or a scaled USD-per-AI3 rate following the `USD_RATE_SCALE` convention).
[...]

Issue #742, opened 2026-06-15
[EPIC] USDC payments support for PayWithAI3


EmilFattakhov:
Tracking issue for accepting **USDC on Ethereum mainnet** as a payment method for
credits.

> **Revised 2026-08-07.** Three changes, all pulling in the same direction: the rate now
> comes from *indexed trade history* instead of a live pool read, and the treasury becomes a
> **manually managed, hard-capped** position instead of an automated one. Superseded wording
> is kept in ~~strikethrough~~ where the delta is worth reviewing.

## Goal
Users buy storage credits with USDC on Ethereum. Credits are granted immediately at a rate
locked on the intent; the treasury accumulates USDC and converts to AI3 **manually, out of
band**, under an exposure cap that shuts the payment path off on its own.

## What changed on 2026-08-07

**1. Rate source: The Graph, not a live pool read.**
AI3/USD comes from the **volume-weighted average of the last N swaps** (default 10, floor 5)
in the Uniswap WAI3/USDC pool, read from the
[subgraph](https://thegraph.com/explorer/subgraphs/DiYPVdygkfjDWhbxGSqAQxwBKmfKnkWQojqeM2rkLb3G)
via the Graph gateway. Same pool on Ethereum mainnet as before — only the *access path* and
the *statistic* change. (`chain=arbitrum-one` in that link is where the subgraph is
published on The Graph Network, not the chain it indexes.)
~~Live `slot0` marginal price + a v4 Quoter `quoteExactOutputSingle` per intent, gated by a
24h TWAP rebuilt from `Swap` logs over `eth_getLogs`.~~

Realized swap prices are the right statistic *because* conversion is now manual: an
executable quote answers "what would it cost to swap this exact size right now", and we no
longer swap per intent. A trade-weighted average of what the pool actually filled recently
is both closer to what the treasury will eventually realize and far cheaper to read.

**2. No automatic USDC→AI3 conversion.**
~~Instant/automated conversion, `sweepToTreasury` + swapper-contract path.~~ USDC sits in
the receiver and grows. An operator sweeps and converts by hand. The FX risk this opens
(price moves between purchase and conversion) is bounded by change 3 rather than hedged.

**3. Two-gate kill switch, one manual and one automatic.**
USDC is accepted only when **both** gates are open:
- **Manual gate** — an admin toggle on the dashboard, DB-backed so it takes effect without a
  redeploy. It **latches**: once an admin turns it off, only an admin turns it back on.
[...]

PR #755, opened 2026-06-16
feat: shared PaymentMethod, token intent fields & payment event ABIs (USDC epic — step 1)

### LINKED ISSUE #743: Models: add PaymentMethod enum, intent payment fields, and ERC20 event ABI
## Context
USDC support introduces a first-class "payment method" concept shared by backend and frontend. Centralize the new types and the ERC20 event ABI in `@auto-drive/models` so both sides stay in sync (mirrors how `EVM_TX_HASH_REGEX` was shared in #_/commit 0878b612).

## Tasks
- [ ] Add `PaymentMethod` enum: `AI3_NATIVE`, `USDC_ETH`.
- [ ] Extend the shared `Intent` type with: `paymentMethod`, `tokenAmount` (string/bigint),
      `usdRateAtCreation`, `quotedTokenAmount`. Keep them optional/nullable for back-compat
      with AI3_NATIVE rows.
- [ ] Export the `IntentTokenPaymentReceived(bytes32 intentId, address token, uint256 amount, address payer)`
      event ABI fragment for reuse by the payment manager and frontend.
- [ ] (If not already shared) confirm `EVM_TX_HASH_REGEX` is reused, not duplicated.

## Acceptance criteria
- `yarn models build` passes; types are importable from both backend and frontend.
- No duplicated PaymentMethod/ABI literals in app code.

EmilFattakhov:
**Part of #742 (USDC Payments for PayWithAI3 epic).**
**Closes #743 — Models: add PaymentMethod enum, intent payment fields, and ERC20 event ABI.**

Step 1 of the epic: the shared `@auto-drive/models` types the backend and frontend will build on to accept USDC on Ethereum alongside native AI3. **Additive and behaviour-preserving** — no payment logic changes yet.

## Changes
**Models (`@auto-drive/models`)**
- Add `PaymentMethod` enum — `ai3_native` / `usdc_eth` (lowercase, matching the existing `IntentStatus` casing).
- Extend `IntentSchema` with:
  - `paymentMethod` — **optional**, mirroring how `expiresAt`/`fromAddress` are handled (DB will be `NOT NULL DEFAULT 'ai3_native'`, but optional in the type for back-compat until the column + wiring land).
  - `tokenAmount`, `quotedTokenAmount`, `usdRateAtCreation` — `bigint`, USDC-only (NULL for AI3 rows).
- Add `USD_RATE_SCALE` (`1e18`) / `USD_RATE_DECIMALS`: the AI3/USD rate is carried as a **scaled bigint** so quote math stays integer-only, consistent with `shannonsPerByte`/`paymentAmount`.
- New `paymentEvents.ts`: relocate `IntentPaymentReceived` ABI out of the backend, and add the new `IntentTokenPaymentReceived` ABI — `intentId` + `payer` indexed. `payer` travels in the event (not read from `receipt.from`) because ERC20 payments can be relayed.

**Backend**
- Payment manager imports the relocated ABI from `@auto-drive/models`; `event-abi.ts` deleted.
- New unit test `PaymentEventAbis.spec.ts` guarding both ABI shapes (event name + indexed topics) and the new `IntentSchema` fields.

## Design decisions (agreed up front)
| Decision | Choice |
| --- | --- |
| USD rate representation | Scaled bigint (×1e18) |
| `PaymentMethod` casing | Lowercase (`ai3_native` / `usdc_eth`) |
| `IntentTokenPaymentReceived` indexed fields | `intentId` + `payer` |
| Event ABI scope | Relocate **both** ABIs into models |

## Verification
- `models` builds clean (`tsc`).
- Backend `tsc --noEmit`: zero errors from this change. (A pre-existing `@polkadot/wasm-crypto` "cannot find module" error is unrelated — an incomplete-install artifact in this environment.)
- New unit test: **10/10 passing** (run in isolation; the full backend suite needs Docker/TestContainers).

## Notes for downstream issues
- `paymentMethod` can be tightened to required in #747/#748 once the column and read/write wiring exist.
- `intentCreationSchema` is intentionally untouched — adding `paymentMethod` to the creation input belongs to #747 (pricing / `createIntent`).


EmilFattakhov:
bugbot run

packages/models/src/users/intent.ts
```diff
@@ -14,20 +14,60 @@ export enum IntentStatus {
...
+  // Native AI3 on Auto EVM — the original PayWithAI3 flow.
+  AI3_NATIVE = "ai3_native",
+  // USDC (ERC20) on Ethereum mainnet — deferred-conversion flow.
+  USDC_ETH = "usdc_eth",
+}
+
+/**
+ * Fixed-point scale for `usdRateAtCreation`. The AI3/USD rate is stored as an
+ * integer scaled by 1e18 (e.g. 0.0064 USD/AI3 -> 6_400_000_000_000_000n) so it
+ * can be carried as a bigint and used in integer-only quote math, matching how
```
jim-counter, line 34: What's the specific intended use-case for storing the USD rate? I know it says quote math here but if we always refund in AI3 on Auto EVM we won't ever need any conversions beyond creating the intent?

packages/models/src/users/intent.ts
```diff
@@ -14,20 +14,60 @@ export enum IntentStatus {
...
+  // --- Token-payment fields: set for USDC_ETH only; NULL for AI3_NATIVE ---
+  // Raw on-chain token amount actually received, in the token's smallest unit
+  // (USDC has 6 decimals). Set by the payment manager on confirmation.
+  tokenAmount: z.bigint().optional(),
+  // Token amount quoted to the user at creation, in the token's smallest unit.
+  quotedTokenAmount: z.bigint().optional(),
+  // AI3/USD rate locked at creation, scaled by USD_RATE_SCALE (1e18). Used to
+  // convert the received token amount to an AI3-equivalent for the existing
+  // proportional credit math.
+  usdRateAtCreation: z.bigint().optional(),
```
jim-counter, line 70: This gets inaccurate when we add a convenience/conversion markup?

jim-counter (commented):
Great start Emil. Couple of questions added.

Might be worth adding an issue to pull anything useful for developers in this feature out into the Auto SDK. I expect most value would be delivered when the epic is complete or nearing completion. The flow that jumps out at me immediately is having a 3rd party developer accepting their own token and handling intent payment.

packages/models/src/users/intent.ts
EmilFattakhov: This would allow us to track and debug in case there's any mismatch between USDC payments and the expected value of storage granted. Basically, for any intent we're storing both AI3 price and it's USD conversion at a time of purchase. For refunds - this number would allow us to convert USDC into the exact amount of AI3 that needs refunding

packages/models/src/users/intent.ts
EmilFattakhov: Yes, that's true. This might require adjustment later, or we might decide to store a proper USD / AI3 conversion rate and then add a markup on top of that. For now I'm thinking of doing latter, but the implementation might evolve as the feature gets developed

packages/models/src/users/intent.ts
jim-counter: Refunds would be:

intent cost in AI3 * (current storage amount / original storage amount)

Unless you're suggesting we refund in USDC?

packages/models/src/users/intent.ts
jim-counter: I think the important thing is that if we store the data we can make the calculations line up.

jim-counter (approved):

packages/models/src/users/intent.ts
EmilFattakhov: No, I think we should refund in AI3 as per design to avoid having an AI3 -> USDC conversion for the second time.

PR #760, opened 2026-06-18
fix(s3): handle control-char keys in ListObjectsV2 via encoding-type=url

### LINKED ISSUE #758: S3: ListObjectsV2 returns 500 on keys with control characters; implement encoding-type=url
## Summary

`ListObjectsV2` returns **HTTP 500 with an HTML body** when a bucket contains a key with control
characters. The XML serializer (`js2xmlparser`) cannot emit characters that XML 1.0 cannot represent
(control chars `0x00–0x08`, `0x0B`, `0x0C`, `0x0E–0x1F`) inside a `<Key>` element, so the listing
handler throws and Express returns its default HTML 500 page.

This is faithful-handling defect: Auto Drive **accepts** these keys on `PutObject` (any UTF-8 key),
but then **cannot list** the bucket that contains them — and because storage is immutable, the key
cannot be removed, so the bucket's listing is permanently broken.

## Reproducer

```bash
# Store a key containing control characters (SOH, STX)
aws --endpoint-url https://public.auto-drive.autonomys.xyz/s3 s3api put-object \
  --bucket <bucket> --key $'recon-ctrl-\x01\x02-test.txt' --body somefile

# List the bucket — fails
AWS_MAX_ATTEMPTS=1 aws --endpoint-url https://public.auto-drive.autonomys.xyz/s3 \
  s3api list-objects-v2 --bucket <bucket>
# => An error occurred (500) when calling the ListObjectsV2 operation
#    (reached max retries: 0): Internal Server Error
```

The underlying exception surfaces to SDK clients as a deserialization failure because the 500 body is
HTML, not S3 XML:

```
operation error S3: ListObjectsV2, ... StatusCode: 500, ...
error while deserializing xml error response :
XML syntax error on line 6: element <meta> closed by </head>
```

The AWS SDK then retries 10× (its default `maxAttempts`), which also makes affected listings very slow.

## Why this happens

- The handler ignores the `encoding-type` query parameter entirely (the rclone integration currently
  sets `list_url_encode: false` as a workaround quirk).
- Keys are serialized as raw bytes into `<Key>` elements.
- XML 1.0 cannot represent most control characters **at all** — not even via numeric character
  references. `encoding-type=url` exists precisely so S3 can return such keys percent-encoded.

## Proposed fix — implement `encoding-type=url` properly

1. Parse the `encoding-type` query parameter on every `ListObjectsV2` request.
2. For each page about to be returned, scan `Contents` keys and `CommonPrefixes` for XML-illegal
   characters (control chars except `\t`, `\n`, `\r`).
3. **Clean page:** return as normal — no behavioural change for the common case.
4. **Affected page + client sent `encoding-type=url`:** URL-percent-encode the affected keys and
   `CommonPrefixes`, include `<EncodingType>url</EncodingType>` in the response body, return 200.
5. **Affected page + client did NOT send `encoding-type=url`:** return **HTTP 400** with a clean S3 XML
   error, code `InvalidRequest`, message e.g. *"This listing contains keys requiring encoding-type=url;
   retry with encoding-type=url set."*

`PutObject` stays permissive (accepts any UTF-8 key, no write-time gate). Storing the key was never the
problem; the listing serialization was. The fix makes the system faithfully handle the data it accepts.

This also lets the upstream rclone `Autonomys.yaml` flip `list_url_encode` from `false` back to the
default `true`, which is required for a clean upstream listing.

## Acceptance criteria

- [x] `aws s3api list-objects-v2 --encoding-type url` on a bucket containing a control-char key returns
      **200** with the affected key URL-encoded and `<EncodingType>url</EncodingType>` in the body.
- [x] `aws s3api list-objects-v2` (no `encoding-type`) on the same bucket returns **HTTP 400** with a
      well-formed S3 XML error body (`InvalidRequest`).
- [x] Listings of buckets with no affected keys are unchanged (no `EncodingType` element, same output).
- [x] `CommonPrefixes` containing control chars handled symmetrically.
- [x] Unit tests covering all three cases (clean page, affected page with opt-in, affected page without
      opt-in).

## Pointers

- Handler: `apps/backend/src/app/controllers/s3/s3.ts` → `listObjectsV2Handler`.
- Dispatch: `apps/backend/src/app/controllers/s3/http.ts` → `getS3Method`.
- XML helper: `apps/backend/src/app/controllers/s3/utils.ts` → `sendXML`.

jim-counter:
Fixes #758

## Problem

`ListObjectsV2` returned **HTTP 500 with an HTML body** when a bucket contained a key with control
characters. XML 1.0 cannot represent most control characters (`0x00–0x08`, `0x0B`, `0x0C`, `0x0E–0x1F`)
at all, so the serializer threw and Express returned its default HTML 500 page. Because storage is
immutable, such a key cannot be removed, so the bucket's listing was permanently broken.

Reproducer:

```bash
aws --endpoint-url https://public.auto-drive.autonomys.xyz/s3 s3api put-object \
  --bucket <bucket> --key $'recon-ctrl-\x01\x02-test.txt' --body somefile
AWS_MAX_ATTEMPTS=1 aws --endpoint-url https://public.auto-drive.autonomys.xyz/s3 \
  s3api list-objects-v2 --bucket <bucket>
# => An error occurred (500) ... Internal Server Error  (HTML body, not S3 XML)
```

## Fix

Implement the standard `encoding-type=url` parameter in `ListObjectsV2`:

- **`encoding-type=url`**: percent-encode every key and `CommonPrefix` and return
  `<EncodingType>url</EncodingType>` (encoding all keys, not just the affected ones, so clients that
  url-decode the whole listing stay correct for keys containing `+`, `%`, or `/`).
- **No opt-in, affected page**: return **400 `InvalidArgument`** (`<ArgumentName>encoding-type</ArgumentName>`)
  instead of throwing a 500.
- **No opt-in, clean page**: unchanged.

`PutObject` stays permissive (any UTF-8 key); storing the key was never the problem, the listing
serialization was.

This also lets the upstream rclone `Autonomys.yaml` flip `list_url_encode` from `false` back to the
default `true`.

## Tests

Unit tests for `hasXmlIllegalChars`, `encodeS3Key` (control chars, `%20`/`%2B`/`%2F`, round-trip,
unreserved set), and `planListingEncoding` (clean → plain, opt-in → encode, affected without opt-in →
reject, illegal `CommonPrefix` symmetric).

### DIFF
--- a/apps/backend/__tests__/unit/controllers/s3-listing-encoding.spec.ts
+++ b/apps/backend/__tests__/unit/controllers/s3-listing-encoding.spec.ts
@@ -1,0 +1,79 @@
+import { describe, it, expect } from '@jest/globals'
+import {
+  encodeS3Key,
+  hasXmlIllegalChars,
+  planListingEncoding,
+} from '../../../src/app/controllers/s3/utils.js'
+
+describe('hasXmlIllegalChars', () => {
+  it('returns false for ordinary keys', () => {
+    expect(hasXmlIllegalChars('dir/file.txt')).toBe(false)
+    expect(hasXmlIllegalChars('hello world (1).txt')).toBe(false)
+    expect(hasXmlIllegalChars('Hello, 世界/ " \' @ < > & ? + ≠')).toBe(false)
+  })
+
+  it('treats TAB, LF and CR as legal (XML permits them)', () => {
+    expect(hasXmlIllegalChars('a\tb')).toBe(false)
+    expect(hasXmlIllegalChars('a\nb')).toBe(false)
+    expect(hasXmlIllegalChars('a\rb')).toBe(false)
+  })
+
+  it('detects control characters XML 1.0 cannot represent', () => {
+    expect(hasXmlIllegalChars('recon-ctrl-\x01\x02-test.txt')).toBe(true)
+    expect(hasXmlIllegalChars('\x00')).toBe(true)
+    expect(hasXmlIllegalChars('\x0B')).toBe(true)
+    expect(hasXmlIllegalChars('\x0C')).toBe(true)
+    expect(hasXmlIllegalChars('\x1F')).toBe(true)
+  })
+})
+
+describe('encodeS3Key', () => {
+  it('percent-encodes control characters', () => {
+    expect(encodeS3Key('a\x01b')).toBe('a%01b')
+  })
+
+  it('encodes spaces as %20 (never +) and encodes a literal +', () => {
+    expect(encodeS3Key('a b')).toBe('a%20b')
+    expect(encodeS3Key('a+b')).toBe('a%2Bb')
+  })
+
+  it('encodes reserved characters so decoding round-trips', () => {
+    expect(encodeS3Key('dir/file')).toBe('dir%2Ffile')
+    expect(encodeS3Key('a%b')).toBe('a%25b')
+  })
+
+  it('round-trips back to the original via decodeURIComponent', () => {
+    const key = 'dir/weird key+\x01\x02-世界.txt'
+    expect(decodeURIComponent(encodeS3Key(key))).toBe(key)
+  })
+
+  it('leaves the unreserved set untouched', () => {
+    expect(encodeS3Key('abcABC123-_.!~*\'()')).toBe('abcABC123-_.!~*\'()')
+  })
+})
+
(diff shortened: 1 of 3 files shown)


EmilFattakhov (approved):
Looks good!

PR #761, opened 2026-06-18
fix(s3): parse CompleteMultipartUpload XML with a real parser

### LINKED ISSUE #759: S3: multipart CompleteMultipartUpload XML parser fails on aws-sdk-go-v2 clients; persists wrong composite ETag
## Summary

Multipart uploads driven by **aws-sdk-go-v2 clients (e.g. rclone)** persist a **wrong composite ETag**:
`d41d8cd98f00b204e9800998ecf8427e-N` — the MD5 of an *empty* buffer, followed by the part count.
`parseMultipartParts` in `apps/backend/src/app/controllers/s3/s3.ts` uses a regex to extract
`<PartNumber>`/`<ETag>` pairs from the `CompleteMultipartUpload` request body; it matches the aws-cli
(boto3) XML but mis-handles the aws-sdk-go-v2 XML, so the per-part MD5s decode to empty bytes and the
composite collapses to the empty-input MD5.

**This is a persisted wrong value, not a transient verification failure.** Every multipart upload an
aws-sdk-go-v2 client has made against Auto Drive in production has stored a wrong ETag, returned on
every subsequent `HeadObject`/`GetObject`/`ListObjectsV2`.

## Root cause

Go's `encoding/xml` escapes double quotes as the numeric character reference `&#34;` in element text;
boto3 (Python) emits literal quotes. The part ETag returned by `UploadPart` is quoted (`"<hex>"`), and
the client echoes it back inside `<ETag>` in `CompleteMultipartUpload`:

- **boto3:** `<ETag>"d41d…"</ETag>` → regex captures `"d41d…"` → quotes stripped → clean hex ✓
- **aws-sdk-go-v2:** `<ETag>&#34;d41d…&#34;</ETag>` → regex captures `&#34;d41d…&#34;`. `parseMultipartParts`
  only un-escapes `&quot;` (not `&#34;`), so the value keeps the `&#34;` entities, the hex decode yields
  empty bytes, and the composite computation hashes an empty concatenation → `d41d8cd9…-N`.

## Reproducer

```bash
# Multipart upload via rclone (aws-sdk-go-v2), any file > 5 MiB
rclone copy largefile.bin autodrive:<bucket>/  --s3-chunk-size 5M --s3-upload-cutoff 5M

aws --endpoint-url https://public.auto-drive.autonomys.xyz/s3 s3api head-object \
  --bucket <bucket> --key largefile.bin
# => "ETag": "d41d8cd98f00b204e9800998ecf8427e-N"   (WRONG — MD5 of empty input)
```

Side-by-side, the **same** multipart upload via aws-cli stores the **correct** composite ETag
(MD5 of the concatenated per-part MD5s, dash, part count) — confirmed by hand:
`upload-part` ETags fed to `complete-multipart-upload` via aws-cli yield e.g. `fa4f9db…-2`, matching the
locally-computed composite. Only the aws-sdk-go-v2 XML path is broken.

## Proposed fix — replace the regex with a proper XML parser

Replace `parseMultipartParts`'s regex with a real XML parser so entity decoding, namespaces, attribute
ordering, and whitespace are handled correctly. `fast-xml-parser` is already present in the dependency
tree (transitively, via the AWS SDK); reuse it rather than adding a new top-level dependency where
possible. The parser must handle:

- aws-cli/boto3 output (no namespaces, ETags in literal double quotes).
- aws-sdk-go-v2 output (possible namespaces, whitespace/attribute differences, `&#34;`-escaped quotes).
- Any other valid `<CompleteMultipartUpload>` body per the AWS spec.

After parsing, validate that every `<Part>` has an integer `<PartNumber>` and a non-empty `<ETag>`, and
sort by `PartNumber`. Reject malformed bodies with **`400 MalformedXML`** (well-formed S3 XML error)
rather than silently falling back to an empty parts array and storing a wrong ETag.

## Historical data (flag for awareness, out of scope here)

Existing multipart objects uploaded by aws-sdk-go-v2 clients already carry wrong ETags in storage
(`md5` column / mapping). A backfill/recompute is **out of scope for this issue** — raise separately if
desired.

## Acceptance criteria

- [x] `rclone copy <largefile>` (aws-sdk-go-v2) against Auto Drive produces a stored ETag matching the
      AWS composite spec (MD5 of concatenated part MD5s + `-` + part count).
- [x] aws-cli multipart behaviour is unchanged (still a correct composite).
- [x] Malformed `CompleteMultipartUpload` XML (missing `<PartNumber>`, empty `<ETag>`, non-integer part
      number, missing root element) returns **`400 MalformedXML`** with a well-formed S3 XML error body.
- [x] Unit tests with fixture bodies from **both** aws-cli and aws-sdk-go-v2.
- [x] Existing multipart tests pass unchanged.

## Pointers

- `apps/backend/src/app/controllers/s3/s3.ts` → `parseMultipartParts`, `completeMultipartUploadHandler`.
- Composite computation: `multipartETag` (from `@autonomys/file-server`) — unchanged; it is correct
  given clean per-part hex. The defect is purely in extracting the ETags from the request XML.

jim-counter:
Fixes #759

## Problem

Multipart uploads driven by aws-sdk-go-v2 clients (e.g. rclone) persisted a **wrong composite ETag**:
`d41d8cd98f00b204e9800998ecf8427e-N` — the MD5 of empty input, followed by the part count. This is a
persisted wrong value, returned on every subsequent `HeadObject`/`GetObject`/`ListObjectsV2`.

Root cause: `parseMultipartParts` used a regex that only un-escaped `&quot;`. Go's XML encoder escapes
the ETag's surrounding quotes as `&#34;` in element text (boto3 emits literal quotes), so the leftover
entities broke the per-part hex decode and the composite collapsed to the empty-input MD5. aws-cli
worked; rclone did not.

## Fix

Replace the regex with `fast-xml-parser` (already resolved transitively via `@aws-sdk/xml-builder`,
promoted to a direct dependency), which decodes entities and handles namespaces, attribute ordering,
and whitespace uniformly across clients. Validate that each `<Part>` has an integer `PartNumber` (>= 1)
and a non-empty `<ETag>`; reject anything else with **400 `MalformedXML`** rather than silently storing
a wrong ETag.

## Tests

Fixtures for both the aws-cli/boto3 body (literal quotes) and the aws-sdk-go-v2 body (`&#34;`-escaped
quotes, reversed element order, namespaced) assert they parse to identical parts, that `&#34;` is gone,
and — end to end — that feeding the parsed go-sdk ETags through `multipartETag` yields the correct AWS
composite and not the empty-MD5 value. Plus namespace-prefixed parsing, part sorting, and malformed-body
cases (missing root, empty body, no parts, missing PartNumber, empty ETag, non-integer PartNumber).

Note: the SDK fixtures are reconstructed from each library's serialization, not captured off the wire.

### DIFF
--- a/apps/backend/__tests__/unit/controllers/s3-multipart-parser.spec.ts
+++ b/apps/backend/__tests__/unit/controllers/s3-multipart-parser.spec.ts
@@ -1,0 +1,122 @@
+import { describe, it, expect } from '@jest/globals'
+import { createHash } from 'crypto'
+import { multipartETag } from '@autonomys/file-server'
+import {
+  MalformedMultipartError,
+  parseMultipartParts,
+} from '../../../src/app/controllers/s3/s3.js'
+
+// Two known per-part MD5s: md5("") and md5("a").
+const M1 = 'd41d8cd98f00b204e9800998ecf8427e'
+const M2 = '0cc175b9c0f1b6a831c399e269772661'
+
+// The AWS composite ETag = md5( raw(M1) ++ raw(M2) ) + "-" + partCount.
+const expectedComposite = `${createHash('md5')
+  .update(Buffer.concat([Buffer.from(M1, 'hex'), Buffer.from(M2, 'hex')]))
+  .digest('hex')}-2`
+
+const buf = (s: string) => Buffer.from(s, 'utf-8')
+
+// aws-cli / boto3: PartNumber before ETag, quotes emitted literally.
+const botoBody = `<?xml version="1.0" encoding="UTF-8"?>
+<CompleteMultipartUpload xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Part><PartNumber>1</PartNumber><ETag>"${M1}"</ETag></Part><Part><PartNumber>2</PartNumber><ETag>"${M2}"</ETag></Part></CompleteMultipartUpload>`
+
+// aws-sdk-go-v2 / rclone: ETag before PartNumber, quotes escaped as &#34;.
+const goSdkBody = `<CompleteMultipartUpload xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Part><ETag>&#34;${M1}&#34;</ETag><PartNumber>1</PartNumber></Part><Part><ETag>&#34;${M2}&#34;</ETag><PartNumber>2</PartNumber></Part></CompleteMultipartUpload>`
+
+describe('parseMultipartParts', () => {
+  it('parses the aws-cli (boto3) body', () => {
+    expect(parseMultipartParts(buf(botoBody))).toEqual([
+      { PartNumber: 1, ETag: `"${M1}"` },
+      { PartNumber: 2, ETag: `"${M2}"` },
+    ])
+  })
+
+  it('parses the aws-sdk-go-v2 (rclone) body, decoding &#34; to literal quotes', () => {
+    const parts = parseMultipartParts(buf(goSdkBody))
+    expect(parts).toEqual([
+      { PartNumber: 1, ETag: `"${M1}"` },
+      { PartNumber: 2, ETag: `"${M2}"` },
+    ])
+    // The bug was that &#34; survived and broke the hex decode.
+    expect(parts.every((p) => !p.ETag.includes('&#34;'))).toBe(true)
+  })
+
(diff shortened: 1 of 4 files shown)


EmilFattakhov (approved):
Looks good!

Issue #762, opened 2026-06-19
Move S3-compatible API to dedicated subdomain (s3.auto-drive.autonomys.xyz)


jim-counter:
## Summary

The S3-compatible API is currently served at the clumsy `https://public.auto-drive.autonomys.xyz/s3` path. This moves it to a dedicated subdomain, `https://s3.auto-drive.autonomys.xyz`, served at the root.

The legacy `/s3` path stays alive **indefinitely** (no sunset) — existing clients keep working untouched. We simply stop *advertising* it. This is a dual-serve, not a redirect: AWS SigV4 signs the Host header and path, so an HTTP redirect to a different host would break SDK uploads. Both endpoints proxy to the same backend.

Scope: **mainnet/production only.** Staging is explicitly out of scope — it gets its own S3 endpoint in the separate staging refactor.

## Dependency ordering ⚠️

These must land in order. Do not merge the docs changes until the endpoint resolves and serves a valid cert.

### 1. Infra (gating) — `autonomys/infra`
- [x] DNS record: proxied Cloudflare A record `s3.auto-drive.autonomys.xyz` → same origin as `public.auto-drive` (reuses `autodrive_public`). **Live** — resolves to the same Cloudflare proxy IPs as `public.auto-drive`.
- Infra PR: autonomys/infra#569 (merged + applied)

**Cert reality (two layers — neither is `certbot --expand` alone).** The auto-drive records are proxied through Cloudflare, so:
- **Edge cert (what clients see):** issued/served by Cloudflare, not the origin. The existing deep subdomains (`public.`, `mainnet.`) each have a per-host 90-day Let's Encrypt edge cert — the signature of Total TLS / per-host auto-issuance. So `s3.` should auto-issue once the proxied record applies. **Verify** (see step 3); only needs Cloudflare-dashboard action if it does *not* auto-issue.
- **Origin cert (Cloudflare → box):** certbot on the EC2 box (not infra-as-code — gap tracked in #765). `certbot --expand` to add `s3.` is required only under Full (strict) SSL mode, harmless otherwise — so do it regardless (no-regret). Requires the DNS record live first (HTTP-01). Handled in the auto-drive deploy, not infra.
- [x] Uptime/health monitoring — N/A; no per-endpoint uptime monitoring on the S3 endpoint (confirmed acceptable).

### 2. auto-drive PR — nginx + routing + Location fix
[...]

PR #763, opened 2026-06-19
feat(s3): move S3-compatible API to dedicated subdomain

jim-counter:
## What

Moves the S3-compatible API onto a dedicated subdomain, `https://s3.auto-drive.autonomys.xyz`, served at the root — replacing the clumsy `https://public.auto-drive.autonomys.xyz/s3` path.

Part of #762.

## How

**1. nginx — dedicated S3 subdomain (`feat(s3): serve S3 API on dedicated subdomain`)**
- New `server` block for `s3.auto-drive.autonomys.xyz` serving S3 at the root. A `rewrite ^/(.*)$ /s3/$1 break;` maps requests onto the backend's existing `/s3` mount, so **no backend routing change is needed**.
- HTTP→HTTPS redirect block for the subdomain.
- The legacy `location /s3` block is **left intact and served in parallel** — this is dual-serve, **not** a redirect. AWS SigV4 signs the Host header and path, so redirecting to a different host would invalidate SDK signatures. Existing clients keep working untouched; we simply stop advertising the old path.
- Live-integration smoke-test scripts now default to the subdomain.

**2. Host-aware `Location` (`feat(s3): return absolute, host-aware Location from CompleteMultipartUpload`)**
- `CompleteMultipartUpload` previously returned a relative `/s3/<bucket>/<key>` `Location`, which is non-compliant (S3 spec wants an absolute URL) and would resolve to a non-existent `/s3/s3/...` on the root-served subdomain.
- Now built in the HTTP layer from proxy forwarding headers (`X-Forwarded-Proto`, `Host`, `X-Forwarded-Prefix`) so it reflects the public endpoint the client used rather than the backend's internal `/s3` mount. The S3 use case drops `Location` and stays host-agnostic. nginx sets the forwarding headers on both S3 vhosts (`/s3` on the legacy host, root on the subdomain).

## Notes

- **The `/s3` CORS + proxy config is intentionally duplicated** between the legacy and subdomain blocks. nginx's only DRY mechanism is `include` (separate files), which would add deployment coupling and hurt at-a-glance readability of each vhost; the duplication is bounded and the blocks are self-documenting.
- Scope is **mainnet/production only** — staging gets its own S3 endpoint in the separate staging refactor.

## Verification

- `yarn backend build` — clean (exit 0)
- S3 unit tests — 53/53 pass
- S3 SDK integration test (real multipart upload via AWS SDK) — 38/38 pass
- `yarn backend lint` — clean

## Depends on (infra, tracked in #762)

This cannot serve traffic until the infra side lands:
1. DNS record `s3.auto-drive.autonomys.xyz` → same host as `public.auto-drive.autonomys.xyz`
2. TLS cert expanded to cover the new name (`certbot --expand -d public.auto-drive.autonomys.xyz -d s3.auto-drive.autonomys.xyz`)

Post-deploy verification (round-trip + S3-key encoding-fidelity check) and the docs updates (rclone #698, developer_docs) are tracked in #762.

### DIFF
--- a/apps/backend/src/app/controllers/s3/s3.ts
+++ b/apps/backend/src/app/controllers/s3/s3.ts
@@ -61,6 +61,37 @@ const getUploadOptions = (req: Request) => {
   return UploadOptions
 }
 
+/**
+ * Build the absolute, host-aware object URL for the CompleteMultipartUpload
+ * `Location` response field — per the S3 spec, the URL of the newly created
+ * object as seen from the endpoint the client used.
+ *
+ * The backend mounts the S3 API at `/s3`, but the public-facing endpoint
+ * differs per vhost (the dedicated `s3.` subdomain serves it at the root; the
+ * legacy host serves it under `/s3`) and nginx rewrites both onto `/s3` before
+ * the request reaches here. So the URL is reconstructed from the forwarding
+ * headers nginx sets rather than from the request's own (rewritten) path:
+ *   - X-Forwarded-Proto — scheme at the edge (nginx terminates TLS and proxies
+ *     over http, so req.protocol alone would always read 'http').
+ *   - Host — the public hostname (forwarded via `proxy_set_header Host`).
+ *   - X-Forwarded-Prefix — the public path prefix the S3 API is exposed under:
+ *     '/s3' on the legacy host, root on the subdomain. nginx drops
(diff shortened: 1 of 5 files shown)


EmilFattakhov (dismissed):
s3.auto-drive.autonomys.xyz is a lot nicer!

EmilFattakhov (approved):

PR #768, opened 2026-06-24
feat(frontend): open purchase screen to all users with Google sign-in gate

EmilFattakhov:
## What

Makes the Buy Credits flow discoverable for everyone, while keeping the actual purchase action gated to Google-verified accounts.

### Changes
- **More visible "Buy more credits" button** — switched from `outline` to the filled `primary` variant (`+ font-semibold`). Primary uses the `--primary` / `--primary-foreground` theme variables, so it renders with consistent, intentional colors in both light and dark schemes instead of a low-contrast outline.
- **Purchase screen open to all users** — removed the `UserProtectedLayout` wrapper from `purchase/page.tsx`. That layout was the only thing blocking unauthenticated access (it redirected logged-out users to `/drive/global` and refused to render until a `user` existed). The Next.js middleware does not gate this route — its matcher only covers `/drive/*`, not the chain-prefixed `/[chain]/drive/*`.
- **Live pricing for everyone** — `api.getCreditPrice()` no longer throws `'No session'`. `GET /intents/price` is already a public backend endpoint (registered before the auth/feature-flag middleware), so auth headers are now optional. Logged-out and non-Google visitors see real prices instead of `—` / `$0.00`.
- **Google sign-in gate** — new `GoogleAuthGate` modal. When a logged-out or non-Google user clicks **Select Package** / **Configure**, they get a prompt to sign in or sign up with Google (with a Google sign-in button) instead of advancing. Detection uses `session.underlyingProvider === 'google'`.
- **Sidebar footer** — "Buy more credits" is now always shown, while the existing "Log In" (logged-out) and "Ask for more credits" (non-Google) buttons are preserved.

## Why

So that the credit packages are discoverable by all visitors (driving conversions), with authentication enforced at the point of action rather than by hiding the screen entirely.

## Behavior by user type
- **Google-authenticated:** unchanged — packages work exactly as before.
- **Non-Google authenticated / logged-out:** can browse packages and see pricing; selecting one opens the Google sign-in prompt.

## Reviewer notes
- The gate keys on Google auth specifically (per request). On production where `buyCredits` is publicly active this matches current behavior. Edge case: during a *staff-only* rollout, a non-staff Google user would now pass the UI gate and could reach the final step before the backend rejects intent creation. Easy to additionally gate on the `buyCredits` feature flag if preferred.
- Verified with `tsc --noEmit` and ESLint (both pass) on the changed files. Not exercised in a running browser (local stack not provisioned in this worktree).

---

### Follow-up fixes (review feedback)
- **Gate can't be bypassed via query params:** the Google check moved from the Step 1 click handler into the wizard and is enforced at render (`effectiveStep`), so non-Google users can't reach later steps via `?step=`. The sign-in prompt still fires at the point of action (package click).
- **Sign-in returns to the flow:** `useLogIn().signIn` now accepts a `callbackUrl`; after Google OAuth the user lands back on the purchase page with their package preselected (`…/drive/purchase?step=2&packageId=<id>`) instead of the default drive view.

### DIFF
--- a/apps/frontend/src/app/[chain]/drive/purchase/page.tsx
+++ b/apps/frontend/src/app/[chain]/drive/purchase/page.tsx
@@ -1,10 +1,9 @@
 import { PurchaseCredits } from '@/components/views/PurchaseCredits';
-import { UserProtectedLayout } from '@/components/layouts/UserProtectedLayout';
 
-export default async function Page() {
-  return (
-    <UserProtectedLayout>
-      <PurchaseCredits />
-    </UserProtectedLayout>
-  );
+// The purchase screen is intentionally available to all visitors — including
+// unauthenticated and non-Google users — so the credit packages are
+// discoverable. Authentication is enforced at the point of action: selecting a
+// package prompts a Google sign-in (see Step1_SelectPackage / GoogleAuthGate).
+export default function Page() {
+  return <PurchaseCredits />;
 }
--- a/apps/frontend/src/components/atoms/BuyMoreCreditsButton.tsx
+++ b/apps/frontend/src/components/atoms/BuyMoreCreditsButton.tsx
@@ -7,7 +7,11 @@ export const BuyMoreCreditsButton = () => {
 
   return (
     <InternalLink href={ROUTES.purchase(network.id)} className='contents'>
-      <Button variant='outline' size='sm' className='w-full text-xs'>
+      <Button
+        variant='primary'
+        size='sm'
+        className='w-full text-xs font-semibold'
+      >
         Buy more credits
       </Button>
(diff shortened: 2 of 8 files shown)


EmilFattakhov:
bugbot run

EmilFattakhov:
The page is now visible for the non-authenticated users

Screenshot 2026-06-24 at 3 53 28 PM

A message is shown for those who wants to proceed with a purchase, but using non-google authentication, or no authentication

Screenshot 2026-06-24 at 3 53 36 PM

EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM!

PR #769, opened 2026-06-25
fix(backend): make object removal hide and block the file everywhere

EmilFattakhov:
Removing ("Remove") a file previously only set marked_as_deleted on the owner's ownership row, which hid it from "My Files" but left it fully downloadable and discoverable by CID for everyone, including anonymous callers, the S3/SDK path, and global search/explorer.

Treat an object as deleted once it has an owner moved it to trash, and enforce that consistently:

- ObjectUseCases.isObjectDeleted + ownershipRepository.getAdminOwnershipState
- authorizeDownload returns ObjectNotFound for deleted objects (covers sync, anonymous, published and S3 downloads, which all funnel through it)
- async createDownload / asyncDownload worker refuse deleted objects
- GET /:cid, /:cid/metadata, /:cid/summary return 404 for deleted objects
- global searchMetadataByCID / searchMetadataByName / getRootObjects only return objects with an active admin owner

Restoring from Trash clears marked_as_deleted, which reverses all of the above. Adds an e2e test covering the full remove/restore lifecycle.

### DIFF
--- a/apps/hasura/metadata/databases/default/tables/metadata.yaml
+++ b/apps/hasura/metadata/databases/default/tables/metadata.yaml
@@ -30,17 +30,25 @@ select_permissions:
       # ObjectUseCases.isObjectDeleted: an object is removed once it has an admin
       # owner but none that is still active. Objects with no admin ownership at
       # all (internal/child nodes) are kept visible.
+      #
+      # Removal is tracked on the root upload's ownership row, so we evaluate the
+      # ROOT's ownership (root_metadata follows root_cid -> head_cid) rather than
+      # the row's own object_ownership. A folder child keeps a vestigial active
+      # admin row keyed by its own cid, so checking object_ownership directly
+      # would wrongly keep a removed folder's children discoverable.
       filter:
         _or:
           - _not:
+              root_metadata:
+                object_ownership:
+                  is_admin:
+                    _eq: true
+          - root_metadata:
               object_ownership:
                 is_admin:
                   _eq: true
-          - object_ownership:
-              is_admin:
-                _eq: true
-              marked_as_deleted:
-                _is_null: true
+                marked_as_deleted:
+                  _is_null: true
       allow_aggregations: true
     comment: ""
 object_relationships:
--- a/apps/hasura/metadata/databases/default/tables/metadata.yaml
+++ b/apps/hasura/metadata/databases/default/tables/metadata.yaml
@@ -12,7 +12,36 @@ select_permissions:
         - tags
         - created_at
         - updated_at
-      filter: {}
+      # Mirror the anonymous removal filter so authenticated users can't discover
+      # another owner's trashed objects by CID/name via a raw GraphQL query (the
+      # per-document filters only protect the UI). As with the anonymous role,
+      # removal is tracked on the root upload's ownership row, so we evaluate the
+      # ROOT's ownership (root_metadata follows root_cid -> head_cid) rather than
+      # the row's own object_ownership; a folder child keeps a vestigial active
+      # admin row so checking object_ownership directly would leak it.
+      #
+      # The third branch additionally keeps a user's OWN objects visible
+      # regardless of removal state, so their Trash view and own-file detail
+      # pages keep working — this is the only difference from the anonymous role.
+      filter:
+        _or:
+          - _not:
+              root_metadata:
+                object_ownership:
+                  is_admin:
+                    _eq: true
+          - root_metadata:
+              object_ownership:
+                is_admin:
+                  _eq: true
+                marked_as_deleted:
+                  _is_null: true
+          - root_metadata:
+              object_ownership:
+                oauth_provider:
+                  _eq: X-Hasura-Oauth-Provider
+                oauth_user_id:
+                  _eq: X-Hasura-Oauth-User-Id
       allow_aggregations: true
     comment: ""
   - role: anonymous
--- a/apps/hasura/metadata/databases/default/tables/metadata.yaml
+++ b/apps/hasura/metadata/databases/default/tables/metadata.yaml
@@ -25,7 +25,22 @@ select_permissions:
         - tags
         - created_at
         - updated_at
-      filter: {}
+      # Hide objects an owner has removed (moved to Trash) so they can't be
+      # discovered anonymously by CID/name, matching the REST 404 behaviour and
+      # ObjectUseCases.isObjectDeleted: an object is removed once it has an admin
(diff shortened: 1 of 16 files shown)


EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

apps/hasura/metadata/databases/default/tables/object_ownership.yaml
```diff
@@ -19,7 +19,16 @@ select_permissions:
...
+        - is_admin
+        - marked_as_deleted
+      # Anonymous clients may reference is_admin / marked_as_deleted to filter
+      # public listings, but must never be able to read or enumerate ownership
+      # rows that an owner has trashed. Restricting visible rows to the active
+      # (non-deleted) ones makes any direct query for deleted rows return empty
+      # and forces the marked_as_deleted column to always read as null.
+      filter:
+        marked_as_deleted:
+          _is_null: true
```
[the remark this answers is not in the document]
EmilFattakhov: Verdict: the described leak does not actually occur
The premise of the issue is that the metadata / metadata_roots anonymous permission filter evaluates the object_ownership relationship through the anonymous role's view of object_ownership (which hides marked_as_deleted rows). That is not how Hasura works.

Permission boolean expressions evaluate relationship sub-expressions against the RAW related table, NOT through that table's row-level select permission. A role's permission filter never inherits the related table's permission.

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
As we discussed last week, I will address the necessary changes on the S3/rclone layer in a separate PR, this one would be the first step that would allow us to test the behaviour and confirm the functionality via UI.

apps/hasura/metadata/databases/default/tables/metadata.yaml
```diff
@@ -25,7 +25,30 @@ select_permissions:
...
+      # ObjectUseCases.isObjectDeleted: an object is removed once it has an admin
+      # owner but none that is still active. Objects with no admin ownership at
+      # all (internal/child nodes) are kept visible.
+      #
+      # Removal is tracked on the root upload's ownership row, so we evaluate the
+      # ROOT's ownership (root_metadata follows root_cid -> head_cid) rather than
+      # the row's own object_ownership. A folder child keeps a vestigial active
+      # admin row keyed by its own cid, so checking object_ownership directly
+      # would wrongly keep a removed folder's children discoverable.
+      filter:
```
jim-counter, line 68: This removal filter is only added to the `anonymous` role — the `user` role's `select_permission` above (line 15) still has `filter: {}`, and on `object_ownership` the `user` role also still exposes `marked_as_deleted` with no row filter.

Since the frontend queries Hasura as role `user` whenever a token is present (apps/frontend/src/services/gql/index.ts:13), a logged-in user can discover another owner's trashed objects — metadata and ownership rows, including `marked_as_deleted` — by issuing a raw GraphQL query that omits the `where`-clause the PR added to the query documents. The per-document filters protect the UI, but enforcement is no longer at the permission layer for `user` the way it now is for `anonymous`, which doesn't match the PR's goal of hiding removed objects "from every view … for everyone."

Downloads stay blocked server-side via `authorizeDownload`, so this is a discovery/metadata leak rather than a content leak — but I'd either mirror this filter onto the `user` role for `metadata` / `root_metadata` / `object_ownership`, or note explicitly that authenticated metadata discovery is out of scope.

Separately, worth a one-line manual check while you're in here: confirm the `anonymous` filter actually hides removed objects given `object_ownership`'s new `marked_as_deleted IS NULL` row permission — i.e. that Hasura evaluates this `_not: { root_metadata: { object_ownership: { is_admin } } }` predicate against raw rows, not the permission-filtered view. (It should, but it's security-critical enough to verify with a single anonymous query for a trashed CID.)

EmilFattakhov: Thanks for a thorough review @jim-counter!

> This removal filter is only added to the anonymous role — the user role's select_permission above (line 15) still has filter: {}, and on object_ownership the user role also still exposes marked_as_deleted with no row filter.

This is addressed, I mirrored the approach of an `anonymous user` to a regular `user` and verified that the metadata could no longer be accessed 

> Separately, worth a one-line manual check while you're in here: confirm the anonymous filter actually hides removed objects given object_ownership's new marked_as_deleted IS NULL row permission — i.e. that Hasura evaluates this _not: { root_metadata: { object_ownership: { is_admin } } } predicate against raw rows, not the permission-filtered view. (It should, but it's security-critical enough to verify with a single anonymous query for a trashed CID.)

The security-critical point is verified: the trashed local object is hidden from anonymous/non-owner. More elaborate response from Claude upon testing:

`If Hasura had evaluated _not: { root_metadata: { object_ownership: { is_admin } } } against the permission-filtered view, that admin row would have been invisible to the predicate → _not would be true → the object would have leaked. It didn't leak, which confirms Hasura evaluates the permission-filter relationship predicate against raw base-table rows, not the permission-filtered view. Exactly as intended.`

EmilFattakhov: Thanks for a thorough review @jim-counter!

> This removal filter is only added to the anonymous role — the user role's select_permission above (line 15) still has filter: {}, and on object_ownership the user role also still exposes marked_as_deleted with no row filter.

This is addressed, I mirrored the approach of an `anonymous user` to a regular `user` and verified that the metadata could no longer be accessed 

> Separately, worth a one-line manual check while you're in here: confirm the anonymous filter actually hides removed objects given object_ownership's new marked_as_deleted IS NULL row permission — i.e. that Hasura evaluates this _not: { root_metadata: { object_ownership: { is_admin } } } predicate against raw rows, not the permission-filtered view. (It should, but it's security-critical enough to verify with a single anonymous query for a trashed CID.)

The security-critical point is verified: the trashed local object is hidden from anonymous/non-owner. More elaborate response from Claude upon testing:

`If Hasura had evaluated _not: { root_metadata: { object_ownership: { is_admin } } } against the permission-filtered view, that admin row would have been invisible to the predicate → _not would be true → the object would have leaked. It didn't leak, which confirms Hasura evaluates the permission-filter relationship predicate against raw base-table rows, not the permission-filtered view. Exactly as intended.`

jim-counter (approved):
LGTM! Looking forward to trying this out.

PR #771, opened 2026-06-30
fix(backend): hide trashed objects from S3 object-lock endpoints

EmilFattakhov:
objectExists only checked for an S3 mapping row, which persists after an owner removes an object (moves it to Trash). GET and ListObjects already treat such objects as not found via isObjectDeleted / notRemovedByOwnerSQL, but GetObjectRetention and GetObjectLegalHold gate on objectExists, so trashed keys could still return retention/legal-hold XML after GET/list reported NoSuchKey. Resolve the mapping's CID through isObjectDeleted so objectExists mirrors the GET/list deletion semantics.


EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM!

PR #772, opened 2026-07-03
Fix: fully depleted credit batches incorrectly marked as expired

EmilFattakhov:
### Problem

In production, a Pay with AI3 purchase whose credits were **100% consumed (0 B remaining)** was showing in the admin panel as **"Expired · awaiting refund"**.

A fully used-up batch forfeited nothing, so no refund is owed. It should read **"Depleted"**, and it should never prompt admins to process a refund.

### Root cause

Two compounding issues:

1. **Backend** — the credit expiry job (`markExpiredCredits`) set `expired = TRUE` on *every* row past `expires_at`, with no check on remaining balance. A batch that was fully consumed before its expiry date still got flagged as expired once the date passed.
2. **Frontend** — `getBatchStatus` checked the `expired` flag before the depleted (0 bytes remaining) check, so the incorrect flag won. The admin "awaiting refund" badge counted every `expired && !refunded` row, surfacing depleted batches as refundable.

### Changes

**Backend**
- `markExpiredCredits` now only expires rows that still have remaining bytes (`upload_bytes_remaining > 0 OR download_bytes_remaining > 0`). Fully depleted rows are never flagged — nothing was forfeited, and forfeited-bytes metrics are unaffected since depleted rows contribute 0.

**Data migration** (`20260703000000-fix-depleted-marked-expired`)
- Un-expires rows already affected in production: `expired = TRUE`, 0 upload + 0 download remaining, `refunded_at IS NULL`.
- Refunded rows are deliberately excluded — `markAsRefunded` zeros remaining bytes, so a refunded row with `expired = TRUE` was genuinely expired *before* the refund.
- Down migration re-flags past-expiry depleted rows, restoring pre-fix behavior.

**Frontend**
- `getBatchStatus` classifies **depleted before expired** — defense in depth: even if a stale `expired` flag reappears, a fully used batch can never render as "Expired".
- `AdminCredits` "awaiting refund" count now additionally requires remaining bytes > 0, so depleted batches never trigger the refund badge.

### Tests

- **New**: repository spec for `markExpiredCredits` (TestContainers Postgres, like the other repo specs) covering: past-expiry with remaining bytes → expired + forfeited totals reported; past-expiry fully depleted → *not* expired, nothing forfeited; future-expiry → untouched.
- **Updated**: frontend `getBatchStatus` spec — the old "prioritises expired over depleted" assertion is inverted, plus a new case for expired-with-remaining-bytes. All 33 frontend unit tests pass.

### DIFF
--- a/apps/frontend/src/components/views/AdminPanel/AdminCredits.tsx
+++ b/apps/frontend/src/components/views/AdminPanel/AdminCredits.tsx
@@ -190,8 +190,15 @@ const groupBatchesByAccount = (
     ),
     userPublicIds: [...new Set(accountBatches.map((b) => b.userPublicId))],
     walletCount: new Set(accountBatches.map((b) => b.fromAddress ?? '—')).size,
+    // Awaiting refund = expired with unused bytes left. Fully depleted
+    // batches (0 remaining) forfeited nothing, so no refund is owed —
+    // require remaining > 0 as defence in depth against stale expired flags.
     refundableCount: accountBatches.filter(
-      (b) => b.expired && b.refundedAt === null,
+      (b) =>
+        b.expired &&
+        b.refundedAt === null &&
+        BigInt(b.uploadBytesRemaining) + BigInt(b.downloadBytesRemaining) >
+          BigInt(0),
     ).length,
     refundedCount: accountBatches.filter((b) => b.refundedAt !== null).length,
     totalPurchased: accountBatches.reduce(
--- a/apps/frontend/src/components/views/AdminPanel/AdminCredits.tsx
+++ b/apps/frontend/src/components/views/AdminPanel/AdminCredits.tsx
@@ -17,6 +17,7 @@ import {
 import { Button, ROUTES, type NetworkId } from '@auto-drive/ui';
 import {
   getBatchStatus,
+  isBatchRefundable,
   STATUS_CLASSES,
   STATUS_LABEL,
 } from '../../../utils/credits';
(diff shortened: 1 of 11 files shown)


EmilFattakhov:
bugbot run

apps/frontend/src/components/views/AdminPanel/AdminCredits.tsx
```diff
@@ -190,8 +190,15 @@ const groupBatchesByAccount = (
     ),
     userPublicIds: [...new Set(accountBatches.map((b) => b.userPublicId))],
     walletCount: new Set(accountBatches.map((b) => b.fromAddress ?? '—')).size,
+    // Awaiting refund = expired with unused bytes left. Fully depleted
```
jim-counter: **The per-user admin refund flow still offers refunds on depleted batches.**

This PR fixes the account-level view (`AdminCredits.refundableCount` now requires
remaining bytes > 0), but `AdminUserCredits.tsx` was not updated and still gates
refundability solely on `refundedAt === null`:

- line 96: `refundableIds = new Set(batches.filter((b) => b.refundedAt === null)...)`
- line 361: `const isRefundable = batch.refundedAt === null;`
- line 490: the "Mark Refunded" button renders for any `isRefundable` batch

So a fully depleted batch renders a "Depleted" badge yet still shows an enabled
"Mark Refunded" button and is included in bulk-refund selection - the exact
mistake this PR is meant to prevent. Suggest applying the same
remaining-bytes > 0 condition here (or deriving refundability from one shared
helper so the two admin views can't drift).

jim-counter (commented):
There's another flow to cover off here.

apps/frontend/src/components/views/AdminPanel/AdminCredits.tsx
EmilFattakhov: Excellent catch Jim, completely missed that screen when verifying the changes, implemented!

EmilFattakhov:
Review follow-up: refundability is now derived from a single shared `isBatchRefundable` helper used by both admin views, and enforced server-side: the refund endpoints return 400 for fully depleted batches (single and combined), so no client can record a refund on a batch that owes nothing.

EmilFattakhov:
bugbot run

apps/frontend/src/utils/credits.ts
```diff
@@ -67,13 +67,39 @@ export interface BatchStatusFields {
 }
 
 export const getBatchStatus = (batch: BatchStatusFields): BatchStatus => {
-  if (batch.expired) return 'expired';
+  // Depleted wins over expired: a fully used-up batch forfeited nothing, so
+  // it must never surface as "Expired" (which implies a refund is owed),
+  // even if a stale expired flag is set on the row.
   if (BigInt(batch.uploadBytesRemaining) === BigInt(0)) return 'depleted';
+  if (batch.expired) return 'expired';
```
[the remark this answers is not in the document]
EmilFattakhov: Depletion definition is intentionally upload-only, verified and enforced in the entire scope of PR

EmilFattakhov:
bugbot run

EmilFattakhov:
Thanks for the review @jim-counter, this is ready for another round!

jim-counter (approved):
LGTM!

PR #773, opened 2026-07-06
fix(frontend): clear stale account/credit/encryption state on sign-out

EmilFattakhov:
Signing out only reset the Zustand user store, but disabled React Query fetches (enabled: !!session) keep serving previously cached data, and the query cache itself was persisted to localStorage and never cleared. 

This left behind some of the banners that should associate with a session:
- the "credits expiring soon" banner 
- purchase-page credit cap active for signed-out users 
- allowed to briefly render the wallet-connect/payment step before the session resolved.

Also clears the persisted default encryption password on logout, and routes the ToU-decline flow through the shared logout path so it gets the same cleanup.

### DIFF
--- a/apps/frontend/src/components/atoms/ExpiryWarningBanner.tsx
+++ b/apps/frontend/src/components/atoms/ExpiryWarningBanner.tsx
@@ -29,7 +29,13 @@ export const ExpiryWarningBanner = () => {
     enabled: !!session?.data,
   });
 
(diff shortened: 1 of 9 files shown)


EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM, nice find on the scenario.

The ideal is always to have any new user/session state stored in a place that is automatically cleared through the design.

PR #778, opened 2026-07-08
feat(s3): soft-delete, server-side copy/move, abort MPU, and mtime for rclone

EmilFattakhov:
## Why

Auto Drive's data on the DSN is permanent, but the S3 layer previously refused every mutating operation — `DeleteObject` returned **403**, `CopyObject`/`AbortMultipartUpload` returned **501**. That blocked rclone's delete/rename/move/sync and prevented us from passing rclone's integration test suite (a prerequisite for getting listed in rclone's official S3-provider docs).

Now that object removal is a **soft-delete** (content hidden, never erased from the DSN — see #769), the S3 namespace can offer the full mutable surface rclone expects while preserving the permanence guarantee.

## Key design decision: mapping-level delete

Deletion is tracked on the S3 mapping (new `S3.object_mappings.deleted_at`), **not** on object ownership. In content-addressed storage a server-side copy is just a remap (`dst key → same CID`), so `move = copy + delete-source`. If delete were ownership-level (per-CID), deleting the move source would also hide the just-created destination (same CID). Mapping-level deletion keeps the destination live.

On top of that, deleting the **last** S3 key for a CID also moves the object to the owner's web-app **Trash** (unified behaviour), and restoring from Trash clears the mapping's `deleted_at` so the S3 key reappears.

## What changed

**S3 API behaviour**
- **DeleteObject** → soft-deletes the `(bucket, key)` mapping; returns **204**, idempotent. Propagates to web-app Trash when it's the last key for the CID.
- **CopyObject** → content-addressed remap, `200` + `<CopyObjectResult>`. Enables rclone move/rename (`Copy + Delete`).
- **AbortMultipartUpload** → discards an in-progress upload (was 501).
- **mtime** → `x-amz-meta-mtime` stored verbatim, echoed on GET/HEAD; `SetModTime` works via metadata-REPLACE copy-onto-self.
- Re-PUT after delete resurrects the key.

**Layers**
- **Migration** — `object_mappings.deleted_at` + `mtime`.
- **Repo** — `softDeleteMapping` / `restoreMappingsByCid` / `countActiveMappingsByCid`; `findByKey`/`listObjects` filter `deleted_at IS NULL`; `createMapping` resets `deleted_at` on conflict.
- **Core** — `S3UseCases.{deleteObject,copyObject,abortMultipartUpload}`, `UploadsUseCases.abortUpload`, `ObjectUseCases.restoreObject` un-hides mappings.
- **Controller/router** — 204/`CopyObjectResult`/204 responses, `x-amz-copy-source` parsing, `x-amz-meta-mtime` headers.
- **DTOs** — `DeleteObject`, `CopyObject` in `@auto-drive/s3`.

**Tests**
- Unit coverage for the new use cases (delete idempotency + Trash propagation, copy remap/NoSuchKey/mtime inherit-vs-replace, abort delegation).
- Integration tests rewritten from 403/501 assertions to soft-delete / copy / move / mtime round-trips.
- Live-integration scripts reframed from immutability-rejection to soft-delete behaviour.

## Verification

`tsc` build ✅ · test-tsconfig typecheck ✅ · eslint ✅ · shell `bash -n` ✅ · **29/29 S3 unit tests pass** ✅ (run Docker-free).

## Reviewer notes / follow-ups
- **rclone docs (PR #698):** reframed doc content prepared separately (not committed here) — delete/move/rename/sync move into "Supported (soft)".
- **CreateBucket/DeleteBucket:** intentionally **not** implemented — this backend folds the bucket into the key's first path segment, so a bare `PUT/DELETE /name` is indistinguishable from a flat object op. `FsMkdir/FsRmdir` pass anyway via the mandated `no_check_bucket = true` (Mkdir is a no-op). A separate task is deciding whether to adopt "no-slash = bucket-op" routing.
- **PublicLink (presigned GET):** not implemented; document as a known exception for the rclone suite.
- Multipart-uploaded objects carry `mtime` only after a later `SetModTime` (rclone sends it on `CreateMultipartUpload`, which isn't persisted to completion); single-PUT objects carry it directly.


apps/backend/src/core/s3/index.ts
```diff
@@ -241,6 +273,116 @@ const putObject = async (
...
+    const result = await ObjectUseCases.markAsDeleted(user, deleted.cid)
+    if (result.isErr()) {
+      // The mapping is already hidden; a failure to also trash the underlying
+      // object (e.g. the caller isn't its owner) must not fail the delete.
+      logger.warn(
+        'Soft-deleted S3 mapping but could not move object to Trash (cid=%s): %s',
+        deleted.cid,
+        result.error.message,
+      )
+    }
```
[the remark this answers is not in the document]
EmilFattakhov: Addressed per [92ef786](https://github.com/autonomys/auto-drive/pull/778/commits/92ef78697f0f3468dabcfb5f6498271af2ecdc08)

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

apps/backend/src/core/s3/index.ts
```diff
@@ -241,6 +298,185 @@ const putObject = async (
...
+    params.Key,
+    source.cid,
+    source.md5,
+    mtime,
+  )
+
+  // The destination is the copier's object: grant them ownership so it appears
+  // in their Drive and they can manage it. Idempotent for a same-user copy
+  // (the uploader already owns the source's content).
+  await OwnershipUseCases.setUserAsAdmin(user, source.cid)
```
jim-counter: The mapping namespace is global (`findByKey` has no user scoping, GET is `downloadObjectByAnonymous`), and the only gate above is `authorizeDownload(source.cid)` — a global banned/removed check, not a caller check. So any authenticated user who can name another user's `(bucket, key)` can copy it and become an **admin owner** of the CID: it appears in their Drive, they acquired it without uploading, and — since `deleteObject`'s gate is "caller is *an* admin owner" — they can then `DeleteObject` the victim's original key, hiding it globally.

Side effect: `setUserAsAdmin`'s upsert resets `marked_as_deleted = NULL`, so a copy also silently pulls the object out of the copier's own Trash.

Suggestion: require the caller to already be an active owner of the source CID (mirror the gate in `deleteObject`), or scope the source lookup once mappings carry an owner column (see review body, point 3).

apps/backend/src/core/uploads/uploads.ts
```diff
@@ -501,6 +505,39 @@ const scheduleCachePopulation = async (cid: string): Promise<void> => {
...
+  ) {
+    return err(
+      new ForbiddenError('User does not have permission to abort this upload'),
+    )
+  }
+
+  // removeUploadArtifacts deletes the blockstore entries, the upload rows keyed
+  // by this root_upload_id (a standalone file upload is its own root), the
+  // buffered file parts, and the file-processing-info row.
+  await removeUploadArtifacts(uploadId)
```
jim-counter, line 551: No status guard: after `CompleteMultipartUpload` the upload row still exists with status `MIGRATING`, and the async `migrate-upload-nodes` worker still needs the blockstore entries. An abort issued after complete (S3 clients do this on retry/cleanup; per the S3 spec it should return `NoSuchUpload`) deletes the blockstore mid-migration — the client holds a CID for an object whose nodes may never publish to the DSN.

Suggestion: return `ObjectNotFoundError` unless the upload is still in an in-progress state (i.e. reject once it reaches `MIGRATING`/completed).

apps/backend/src/core/s3/index.ts
```diff
@@ -241,6 +298,185 @@ const putObject = async (
...
+    return ok()
+  }
+
+  // Ownership gate: the S3 namespace is shared across API keys, so only an
+  // ADMIN owner (the uploader) may hide a key. Shared recipients are stored as
+  // non-admin owners and must not be able to hide a key for everyone. A caller
+  // without active admin ownership is a no-op (still reported as success for
+  // idempotency) rather than letting a sharee or other tenant hide the object.
+  const owners = await ownershipRepository.getOwnerships(mapping.cid)
+  const isAdminOwner = owners.some(
```
jim-counter: This authorizes per-CID, but the thing being hidden is a per-key mapping. With dedup, an
unrelated user who uploaded identical content is also an admin owner of this CID and passes this gate for *your* key. See review body point 3 — an owner column on `object_mappings` would let this check, `countActiveMappingsByCid`, and `restoreMappingsByCid` all scope correctly.

jim-counter (changes requested):
Requesting changes for three issues, two anchored inline and one design-level here because it spans several queries:

1. **`CopyObject` privilege escalation** (inline in `core/s3/index.ts`) — copy grants the caller admin ownership of any CID reachable by `(bucket, key)` with no source-ownership check, and from there they can hide the victim's keys via `DeleteObject`.
2. **`abortUpload` missing status guard** (inline in `core/uploads/uploads.ts`) — abort-after-complete deletes blockstore artifacts mid-migration and can lose a completed object.
3. **`object_mappings` has no owner column.** Delete authorization is per-CID ("am I an admin owner of the content?") but the thing being deleted is a per-key mapping. With content-addressed dedup, two unrelated users who upload identical bytes are both admin owners of the same CID, so either can hide the other's key. The same gap distorts `countActiveMappingsByCid` (my last-key delete won't Trash-propagate because your identical file's key counts as active) and `restoreMappingsByCid` (my restore un-hides your keys). Adding an owner (`oauth_provider`/`oauth_user_id`) column to `S3.object_mappings` and scoping the delete gate, count, and restore to it fixes all three at once — and gives issue 1 a clean scope too.

Two non-blocking notes, no action needed before merge: `marked_as_deleted` is stamped with app-side `new Date()` while `deleted_at` uses DB `NOW()`, so the `deleted_at >= since` restore comparison is exposed to clock skew (and in the race-guard path the mapping is hidden *before* the trash timestamp, so restore misses it); and the Object Lock flip from `Enabled/COMPLIANCE/ON` to `404/OFF` is client-visible and worth a line in release notes.

apps/backend/src/core/s3/index.ts
EmilFattakhov: Confirmed correct and fixed via the suggested option 👍

apps/backend/src/core/uploads/uploads.ts
EmilFattakhov: Confirmed the finding is valid, fixed as per suggestion 👍

apps/backend/src/core/s3/index.ts
EmilFattakhov: Confirmed correct and fixed - migration adds `owner_oauth_provider/owner_oauth_user_id` to `S3.object_mappings`, `deleteObject`(index.ts) gates on the per-mapping owner, trash-count is owner-scoped, restore un-hides only the restorer's own keys

EmilFattakhov:
bugbot review

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
Okay now it's gotten really ginormous, but all the points (and million bugbot findings) have been addressed, ready for another round @jim-counter

apps/backend/__tests__/unit/controllers/s3-object-lock.spec.ts
```diff
@@ -50,35 +46,16 @@ describe('getS3Method — Object Lock dispatch', () => {
   })
 })
 
+// Object Lock is not enforced now that the S3 namespace is mutable (delete /
+// overwrite / rename succeed), so the read endpoints report "no lock": the
+// configuration/retention handlers return 404 ObjectLockConfigurationNotFound /
+// NoSuchObjectLockConfiguration (exercised via the endpoints), and legal hold is
+// always OFF.
 describe('Object Lock response bodies', () => {
```
jim-counter, line 54: We are underselling one of our value props here. See https://github.com/autonomys/auto-drive/issues/781.

apps/backend/src/app/controllers/s3/s3.ts
```diff
@@ -262,39 +305,32 @@ export const headObjectHandler = async (req: Request, res: Response) => {
...
-export const objectLockConfigurationBody = () => ({
-  ObjectLockEnabled: 'Enabled',
-  Rule: { DefaultRetention: { Mode: 'COMPLIANCE', Years: 100 } },
-})
-
-export const objectRetentionBody = () => ({
-  Mode: 'COMPLIANCE',
-  RetainUntilDate: OBJECT_LOCK_RETAIN_UNTIL,
-})
+// Object Lock is NOT enforced. The S3 namespace is mutable — DeleteObject
```
jim-counter, line 308: We are underselling one of our value props here. See https://github.com/autonomys/auto-drive/issues/781.

apps/backend/src/app/controllers/s3/s3.ts
```diff
@@ -61,6 +61,43 @@ const getUploadOptions = (req: Request) => {
...
+ */
+const getMtime = (req: Request): string | null => {
+  const mtime = req.headers['x-amz-meta-mtime']
+  return typeof mtime === 'string' ? mtime : null
+}
+
+/**
+ * Parse the source object of a CopyObject request from its x-amz-copy-source
+ * header. The value is "/{bucket}/{key}" (the leading slash is optional), with
+ * each path segment URL-encoded and an optional "?versionId=..." suffix — Auto
```
jim-counter, line 77: Note the proposal around versioning in https://github.com/autonomys/auto-drive/issues/781.

jim-counter (dismissed):
## Follow-up proposal: scope the S3 namespace per user

First — the fix commits land all three review issues properly (source-gated copy with no
ownership re-grant, the PENDING guard on abort, and the per-mapping owner column with scoped delete/count/restore), and the added cross-owner write guards close a key-stealing hole the review hadn't even flagged. The test coverage for the dedup and cross-owner cases is exactly what was needed. Nothing below blocks this PR.

### Why raise a design change now

Working through the review surfaced a pattern: nearly every hard problem in this PR — the dedup co-owner gates on delete and copy, the guarded upsert with its null-return protocol, `destinationOwnedByAnother`, the cross-owner `AccessDenied` paths, the legacy-fallback branching — exists to referee contention over a single global `(bucket, key)` namespace shared by all users. Each guard is individually correct, but they're symptoms of the same root cause: the namespace has no owner dimension, while everything layered on top of it now does.

The global namespace also produces a weak user experience that will surface quickly once rclone users arrive: whoever PUTs `default/test.txt` first owns that name **forever** — a second user uploading their own `test.txt` gets `403 AccessDenied`, and soft-delete never releases the name. First-come-first-served squatting on common filenames in shared buckets is a support burden waiting to happen, and "your key name is taken by a stranger you can't see" is not explainable behaviour for an S3 provider.

Real S3 doesn't have this problem because bucket namespaces live inside an account. The proposal is the same move here: make the mapping identity `(owner, bucket, key)`.

### What changes

**Migration** — replace the `(bucket, key)` uniqueness with a composite primary key:

```sql
-- Resolve remaining NULL-owner legacy rows by DUPLICATION (see rationale below).
INSERT INTO "S3".object_mappings
  (bucket, "key", cid, md5, mtime, deleted_at, created_at, updated_at,
   owner_oauth_provider, owner_oauth_user_id)
SELECT om.bucket, om."key", om.cid, om.md5, om.mtime, om.deleted_at,
       om.created_at, om.updated_at, oo.oauth_provider, oo.oauth_user_id
FROM "S3".object_mappings om
JOIN object_ownership oo ON oo.cid = om.cid AND oo.is_admin = true
WHERE om.owner_oauth_user_id IS NULL;

DELETE FROM "S3".object_mappings WHERE owner_oauth_user_id IS NULL;

ALTER TABLE "S3".object_mappings
  ALTER COLUMN owner_oauth_provider SET NOT NULL,
  ALTER COLUMN owner_oauth_user_id SET NOT NULL;

ALTER TABLE "S3".object_mappings DROP CONSTRAINT object_mappings_pkey;
ALTER TABLE "S3".object_mappings
  ADD PRIMARY KEY (owner_oauth_provider, owner_oauth_user_id, bucket, "key");
```

Legacy-row rationale: a NULL owner means the backfill found several admin owners for the CID and couldn't tell who created the key. Guessing (e.g. earliest owner) risks silently
404ing the *actual* S3 user's key — their sync scripts break with no self-service recovery. Duplicating gives every admin owner their own copy; the worst case is a surprise key the user can simply soft-delete from their own namespace. Losing access is strictly worse than gaining clutter. (Worth a pre-flight `COUNT(*)` of NULL-owner rows in production — if it's near zero this whole decision is academic.)

**Code — net deletion.** With contention impossible, the guard machinery becomes
unreachable and comes out:

- `destinationOwnedByAnother` + both pre-check call sites; the guarded
  `ON CONFLICT ... WHERE owner` upsert reverts to a plain upsert returning non-null;
  all four `if (!mapping) return ForbiddenError` checks
- The owner-vs-legacy branching in `deleteObject` and the mirrored source-ownership block in `copyObject` — `findByKey(user, bucket, key)` returning null *is* the check
- The three `ForbiddenError → AccessDenied` branches in the put/complete/copy handlers
- `getMappingOwner` in the repository; the dual scoped/unscoped `countActiveMappingsByCid`

Mechanical additions: `findByKey` / `softDeleteMapping` / `listObjects` / `listBuckets`
gain the owner in their WHERE clause, and `getObject` / `listObjects` / `listBuckets` /
`objectExists` take a `user` param — every handler already runs `handleS3Auth`, so the
identity is one argument away. Everything else in this PR is untouched: soft-delete,
copy-as-remap, abort MPU + PENDING guard, mtime (including the staging table),
Object Lock, `UploadPartCopy` guard, DTOs, live-integration scripts. Ballpark: −300 lines of guards, +100 of owner threading, one migration. Rough estimate 2–3 focused days including tests (replace the ~10 cross-owner guard tests with two-users-same-key isolation tests).

### Why this sequencing works

This stacks cleanly on the current PR rather than reworking it — the owner columns and
backfill it depends on are already here, and every mapping this PR creates carries exactly one owner, so the data migrates forward with no row rewrites. Options:

1. **Stack it as a follow-up PR on this branch and release together** (preferred) — #778
   keeps its review history, and rclone users never observe the interim first-writer-wins
   403 behaviour that would otherwise need explaining and later un-explaining.
2. Merge #778 now and follow up — safe data-wise, but the 403-on-collision behaviour ships and becomes a (short-lived) support surface.

### One thing to confirm

GET/HEAD/list become owner-scoped, so if anything today deliberately relies on reading
*another* user's keys through the S3 API, this breaks it. It looks like an artifact of the original table design rather than a feature, but it should be confirmed rather than assumed. (Public sharing by CID via the download API is unaffected.)

EmilFattakhov:
This is a merge commit that would go into main / production, no code changes to that one since your last review @jim-counter

EmilFattakhov:
Sorry actually please give me some time to verify the changes

EmilFattakhov:
Okay my feeling was right, some of the migrations could be rewritten and cleaned up before the final merged, now it's ready for another review, thanks!

jim-counter (approved):
LGTM

PR #779, opened 2026-07-08
docs(s3): record decision to not implement bucket endpoints

EmilFattakhov:
## What & why

The S3 API folds the logical "bucket" into the first path segment of the object key (`parseBucketAndKey`). A true S3 `CreateBucket`/`DeleteBucket`/`HeadBucket` targets a bare `PUT|DELETE|HEAD /{bucket}` with **no** object key — which, under this scheme, is indistinguishable from a flat, default-bucket object operation. That flat-key behaviour is exactly what the `s3-bucket-support` migration and every legacy flat key depend on, and what the integration harness (`bucketEndpoint: true` + flat keys like `test.txt`) exercises.

This came up while implementing S3 soft-delete + copy/move for rclone compatibility. Two options were on the table:

- **(a)** Accept the `no_check_bucket` workaround and **not** implement bucket endpoints.
- **(b)** Adopt a "no-slash path = bucket operation" routing rule — which reinterprets every legacy flat-key mutation (forcing existing objects under a `default/` prefix) and requires rewriting the integration tests.

**Decision: (a).** rclone's `FsMkdir`/`FsRmdir` already pass when the remote is configured with `no_check_bucket = true` (Mkdir becomes a no-op; Rmdir on a sub-path root returns nil), so dedicated bucket endpoints are unnecessary under the mandated config. (a) keeps existing object semantics and legacy flat keys intact at zero risk; (b) changes a shipped contract for a benefit that only applies to clients ignoring the mandated config.

## Changes

This PR contains **no runtime behaviour change** — the server already behaves per (a). It makes the decision explicit and durable so it can't be silently reversed:

- **`http.ts`** — authoritative decision note directly above the handler map (where a future engineer would look to "add CreateBucket").
- **`s3.ts`** — pointer on `parseBucketAndKey`.
- **`docs/architecture.md`** — note under the S3-Compatible Endpoints table.
- **s3-sdk integration test** — locks the decision: a bare, slash-less `PUT`/`HEAD`/`DELETE` is asserted to be an object op (`PutObject`→200 + body round-trip, `HeadObject`→200 + `Content-Length`/MD5 `ETag`, `DeleteObject`→403), so adopting a "no-slash = bucket op" rule later is a conscious, test-breaking choice. (The HEAD case asserts object metadata rather than a bare 404, since a real `HeadBucket` on a missing bucket also 404s and would not break — per senior review.)

## Reviewer notes

- **Accepted trade-off:** a client that ignores `no_check_bucket` and sends a bare `PUT /{bucket}` will silently create an empty object named `{bucket}` in `default`. Closing that is option (b); it stays a documented limitation.
- The user-facing rclone docs that already mandate `no_check_bucket = true` live on the separate, still-unmerged `docs/rclone-integration` branch — consistent with this decision.

### DIFF
--- a/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
+++ b/apps/backend/__tests__/integration/s3-sdk/index.spec.ts
@@ -526,8 +526,8 @@ describe('AWS S3 - SDK', () => {
   // bucket name (parseBucketAndKey), such a request is indistinguishable from a
   // flat, default-bucket object op — the semantics the bucket-support migration
   // and legacy flat keys rely on. Bucket endpoints are therefore left
-  // unimplemented, and clients must set `no_check_bucket = true` (see
-  // docs/rclone) so they never emit CreateBucket/HeadBucket. These tests lock
+  // unimplemented, and clients must set `no_check_bucket = true` so they never
+  // emit CreateBucket/HeadBucket. These tests lock
   // that decision: a bare, slash-less PUT/HEAD/DELETE is handled as a
   // default-bucket object op, so adopting a "no-slash = bucket op" rule later is
   // a conscious, test-breaking change rather than a silent regression.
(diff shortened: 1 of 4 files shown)


jim-counter (approved):
LGTM. If we added deeper bucket support we would need to adjust the DSN storage layer so we could rebuild the full S3 structure if we ever wanted to.

Issue #749, opened 2026-06-15
Config: add Ethereum/USDC/oracle env vars and wire startup


EmilFattakhov:
## Context
New configuration surface for the second chain, the oracle and the payment gates. Existing
keys: `EVM_CHAIN_ENDPOINT`, `EVM_CHAIN_CONTRACT_ADDRESS`, `EVM_CHAIN_CONFIRMATIONS`,
`CREDITS_PRICE_MULTIPLIER` (in `apps/backend/src/config.ts`).

> **Updated 2026-08-07** for the subgraph rate source (#746) and the two-gate kill switch
> (#811). Superseded keys in ~~strikethrough~~.

## Tasks
- [ ] Ethereum / USDC chain:
      - `ETH_CHAIN_ENDPOINT` (already present as `config.ethereum.rpcUrl`, optional by design)
      - `ETH_CHAIN_ID`
      - `ETH_USDC_RECEIVER_ADDRESS`
      - `USDC_TOKEN_ADDRESS`
      - `ETH_CHAIN_CONFIRMATIONS` (default 6)
- [x] Oracle (subgraph VWAP — #746) — **landed on `feat/price-oracle-dex-source`**, exact
      names as implemented:
      - `GRAPH_SUBGRAPH_URL`, `GRAPH_API_KEY` — both read via `process.env` (not `env()`) so
        they stay optional and a non-USDC deployment boots without them. The key is
        **secret**: never logged, and it is billed per query, which is a second reason the
        derived rate is cached rather than read per request
      - `ORACLE_REQUEST_TIMEOUT_MS` (10000) — one HTTP query, replacing the RPC-era pair
      - `ORACLE_SWAP_SAMPLE_SIZE` (10), `ORACLE_MIN_SWAP_SAMPLES` (5)
      - `ORACLE_MAX_SWAP_AGE_MS` (86400000), `ORACLE_MAX_INDEX_LAG_MS` (900000)
      - `ORACLE_MIN_WINDOW_VOLUME_USDC` (1000, whole USDC)
      - `ORACLE_MAX_SWAP_DEVIATION` (25 percent)
      - kept: `ORACLE_CACHE_TTL_MS`, `ORACLE_MAX_STALE_MS`, `ORACLE_MIN_USD_PER_AI3`,
        `ORACLE_MAX_USD_PER_AI3`, `USD_QUOTE_MARGIN`
[...]

PR #780, opened 2026-07-08
feat: AI3/USD price oracle service (USDC epic — step 4)

### LINKED ISSUE #746: Oracle service: AI3/USD from a subgraph VWAP of the last N swaps
## Context

> **Rewritten 2026-08-07.** This issue previously shipped (PR #807) as a live pool read: a
> `slot0` marginal price plus a v4 Quoter call per intent, gated by a 24h TWAP rebuilt from
> `Swap` logs. **Reopened** because the epic no longer converts USDC per intent (#742), so
> pricing against instantaneous pool state is no longer the right statistic — and because the
> log-and-state reconstruction needs an archive RPC to do a job an indexer already does.
> Superseded tasks are kept in ~~strikethrough~~; the guard *philosophy* (fail closed,
> explicit error taxonomy, last-good cache) carries over unchanged.

The rate is now the **volume-weighted average price of the last N swaps** in the Uniswap
WAI3/USDC pool on Ethereum mainnet, read from the
[published subgraph](https://thegraph.com/explorer/subgraphs/DiYPVdygkfjDWhbxGSqAQxwBKmfKnkWQojqeM2rkLb3G)
through the Graph gateway.

Why realized trades rather than pool state: the treasury no longer swaps at quote time
(conversion is manual, see #751), so "what would this exact size cost right now" answers a
question nobody asks. What a purchase should be priced against is what the pool has actually
been filling at — which is exactly what indexed swaps record, fee and price impact included.

## The statistic

For each of the last N swaps take the USDC leg and the AI3 leg, normalized by decimals:

```
usdPerAi3 = Σ |usdc_i| / Σ |ai3_i|          (over the N-swap window)
```

Summing before dividing *is* the volume weighting — a swap's influence is its size — and it
avoids N rounding errors from per-swap division. Keep the `USD_RATE_SCALE` (1e18) integer
representation and the existing `parseDecimalToScaledBigint` conventions.

## Tasks
- [ ] New adapter under `apps/backend/src/infrastructure/services/priceOracle/` (e.g.
      `subgraphSwaps.ts`) that queries the gateway for the last N swaps of the target pool:
      per-swap token amounts, timestamp, block, plus `_meta { block { number timestamp } }`
      for index lag. Auth via `GRAPH_API_KEY`; never log the key.
- [x] **Verify the subgraph covers this pool before building against it** — confirm it
      indexes the mainnet v4 PoolManager (or the correct pool contract), that swaps are
      filterable by pool id, and pin the exact schema field names. This is the one unknown
      that can invalidate the rest of the issue.
- [ ] Compute the VWAP as above. N configurable (`ORACLE_SWAP_SAMPLE_SIZE`, default 10).
- [ ] Guards — each one **fails closed** with its own error, never a silent fallback:
      - **Sample count**: fewer than `ORACLE_MIN_SWAP_SAMPLES` (default 5) usable swaps in
        the response → unavailable. The pool trades ~1.6×/day, so this is a real state.
      - **Freshness**: newest swap older than `ORACLE_MAX_SWAP_AGE_MS` (default 24h) →
        unavailable. Note the window's *span* is data-dependent: 10 swaps at historic rate is
        ~6 days, so freshness must be judged on the newest swap, not the window.
      - **Window volume**: Σ USDC below `ORACLE_MIN_WINDOW_VOLUME_USDC` (default 1000) →
        unavailable. Without this, ten dust swaps set the price of every purchase, and on a
        thin pool that is the cheapest attack available.
      - **Outlier trim**: drop swaps deviating more than `ORACLE_MAX_SWAP_DEVIATION` (default
        25%) from the window median before averaging; re-check the sample-count floor after
        trimming.
      - **Absolute bounds**: keep `ORACLE_MIN_USD_PER_AI3` / `ORACLE_MAX_USD_PER_AI3`.
      - **Index lag**: `_meta` block older than `ORACLE_MAX_INDEX_LAG_MS` → unavailable. An
        indexer that has stopped is indistinguishable from a pool that has stopped trading,
        and the two need different pages.
- [ ] Cache with TTL (`ORACLE_CACHE_TTL_MS`) + last-good fallback within `ORACLE_MAX_STALE_MS`.
      Unlike the old executable quote, a VWAP **may** be served from cache — it is an average
      over days, not a block-scoped read backing a swap.
- [x] Error taxonomy: **one** error, `OracleUnavailableError`, carrying a machine-readable
      `reason` (`gateway` | `misconfigured` | `insufficient-samples` | `stale-window` |
      `market-moved` | `thin-volume` | `out-of-bounds` | `indexer-lag` | `indexer-error` |
      `narrow-window`) as well as a message naming the guard. A dashboard that
      has to regex a string to render a status breaks on the next wording change.
      `market-moved` and `misconfigured` were both added after review (2026-08-10); see the two
      sections below. `misconfigured` exists because "we cannot reach The Graph" and "we are
      pointed at the wrong pool" are fixed by waiting and by a redeploy respectively, and
      collapsing them sends whoever is paged to a status page to debug one of our own constants.
      It covers a missing `GRAPH_API_KEY`, a subgraph without the pool, and currencies in the
      wrong order — deliberately NOT transport failures, which stay `gateway`.
      `QuoteTooLargeError` / `PoolEmptyError` / `InvalidQuoteAmountError` retire with the
      quoter; **`PriceDeviationError` retires too** (decided 2026-08-07) — it existed to judge
      live spot against a trailing average, and with no live spot read there is nothing left
      for it to compare. The outlier trim, the volume floor and the absolute bounds are the
      manipulation controls now.
- [x] **No on-chain read at all** (decided 2026-08-07): no viem client, no `slot0`, no
      advisory cross-check. One source, one failure mode. The Ethereum RPC stays in the
      codebase for #748 and the #811 balance poller, and `ETH_CHAIN_ENDPOINT` no longer needs
      to be archive-capable. Pool identity constants move to `pool.ts`.
- [ ] Retire the archive-RPC machinery this replaces: `fetchSwapPrices` (`eth_getLogs`
      chunking), `readUsdPerAi3At`, the TWAP reference cache and `ORACLE_TWAP_*` config. The
      Ethereum RPC is still needed by #748 and the balance poller, just not by the oracle.
      `timeWeightedAverage` in `quote.ts` goes with them unless the advisory check reuses it.
- [ ] Expose oracle health (last successful read, which guard last fired, sample count,
      window span, window volume) for the admin dashboard in #811.
- [ ] Unit tests, all HTTP mocked: VWAP math incl. weighting, sample-count floor,
      staleness, thin-volume rejection, outlier trim, bounds, index lag, TTL cache, last-good
      fallback, and gateway 4xx/5xx/timeout.

## Verified against the live gateway (2026-08-10)

The subgraph is a **v4** subgraph (`hooks`, `modifyLiquiditys`, pool id = the PoolKey hash)
indexing the Ethereum mainnet PoolManager, and it has our pool:
`token0=WAI3` (18dp) / `token1=USDC` (6dp) — the same ordering `POOL_KEY` pins. Indexing
healthy, `hasIndexingErrors: false`.

Schema as used: `swaps(first, orderBy: timestamp, orderDirection: desc, where: { pool })`
returning `id timestamp amount0 amount1`, plus `_meta { block { number timestamp }
hasIndexingErrors }` and a `pool(id)` identity check in the same round trip. (Per-swap
`transaction { blockNumber }` was listed here in an earlier revision and is **not** requested:
nothing needs a block number per fill, the VWAP is time-ordered, and the indexer's own head is
already captured for the lag guard.) **Amounts are BigDecimal strings in whole tokens and signed by direction** ("199392.024",
"-477.1285"), not base units — the adapter takes the magnitude and scales by the token's
decimals. `amountUSD` is deliberately unused: it is the subgraph's own valuation through its
pricing paths, whereas `amount1` IS the USDC that changed hands.

**Live verdict: the oracle refuses, correctly.** The newest fill is 2026-08-05 08:12:47 UTC —
130 hours old against a 24h bound — so `getPrice()` returns `reason=stale-window`. The pool's
liquidity is still `0`. Further guards would stop this window even so: after the 25% trim removes
three swaps from the tail of the 59% decline, the survivors total **919.53 USDC**, below the 1000
USDC floor — and those three are the three *newest* fills, which is a refusal in its own right
(see below). Every figure is a property of a dead pool, so re-derive every threshold once it
trades again.

The window is also ten consecutive **sells** — WAI3 in, USDC out, all ten — which is a pool being
drained rather than a market with two sides. Nothing here detects one-directional exhaustion;
staleness and the volume floor are what caught this one. Worth holding in mind during
re-derivation.

## Manipulation controls (revised 2026-08-10 after review)

The trim measures deviation against a **count-based median**, so whoever supplies most of a
ten-swap window sets the price. Volume alone does not stop that: an attacker who clears the
1000 USDC floor with their own fills clears it at whatever price they like — roughly $20 of
fees on this 1% pool bought a 64x discount. Two bounds close it:

- **`ORACLE_MIN_WINDOW_SPAN_MS` (2h)** — the surviving fills must span at least this long.
  Time is the part that cannot be faked: a burst of fills in one block is printable on demand,
  while the same fills held across hours must be defended against everyone else trading in
  between, and that delay is what the treasury alerting needs to react. Refuses with
  `narrow-window`.
- **`ORACLE_MAX_WINDOW_AGE_MS` (7d)** — the window's lower time bound. Without it a majority of
  ancient fills carries the median, the trim discards the *recent* fills as outliers, and a rate
  from another era is served and charged. A filter, not a refusal, so one stale fill cannot deny
  an otherwise live window.

What remains, and is pinned as a test: an attacker who sustains a wrong price across hours *can*
still deny the payment path, because trimming their fills leaves too few samples. Denial, not
mispricing — the direction this must fail in. If denial becomes the concern, the answer is a
deeper pool or a second source, not a looser trim.

## A genuine market move is not manipulation, and the trim mishandled it (added 2026-08-10)

The count-based median has a second failure mode, found in review, and it needs no attacker at
all. The trim drops the lone absurd print it was built for — but when the market genuinely
**re-prices**, the fills carrying the new price are the *minority*, so the trim discards exactly
those. What survives is a majority that has not caught up, and its average is calmly wrong in the
direction of the old regime while every other guard is satisfied: enough samples, enough volume,
spanning hours, inside the absolute bounds. `ORACLE_MAX_WINDOW_AGE_MS` does not help, because at
7d it is far wider than the move.

Measured on this pool's own ten real fills, made fresh and given volume over the floor with the
prices untouched: the oracle serves **0.004698 USD/AI3 while the most recent fill was
0.002393** — 96% above anything that traded, from 7 samples and 1,839 USDC spanning 6h. At ~1.6
swaps/day the window needs days to catch up, and every purchase in between is charged at a price
nobody traded. A crash over-charges the buyer; a rally hands `USD_QUOTE_MARGIN` back as free
storage and is arbitrageable — buy credits immediately after AI3 moves up.

- [x] **The newest fill in the window gets a veto.** If the trim discarded it, the window
      describes a regime the market has left, and the oracle refuses with `market-moved` instead
      of quoting. Denial rather than mispricing, the direction everything else here fails in, and
      it costs nothing on a volatile-but-continuous market because then the newest fill sits inside
      the band. `ORACLE_MAX_SWAP_DEVIATION` consequently sets two things — which fills are dropped,
      and how far the market may move before the path shuts — so derive it from how far this pool
      actually moves between consecutive fills.
- [x] Config relationships are validated at import. `ORACLE_MAX_SWAP_AGE_MS >
      ORACLE_MAX_WINDOW_AGE_MS` used to be accepted and *silently weakened* the oracle rather than
      refusing: the window filter runs first, so the freshness guard could never fire and a market
      that stopped six days ago would price as current.

Still open, and deliberately so: nothing detects a **one-directional** window (the live one is
ten consecutive sells). A trend filter and a two-sided-flow requirement are both plausible next
steps, and both want real trading data to size against.

## Acceptance criteria
- Deterministic, unit-tested VWAP with every guard covered and no network in tests.
- Every refusal names the guard that fired; callers can tell an outage from a thin market, and a
  misconfigured deployment from an unreachable one.
- A drained or untraded pool produces a clean `OracleUnavailableError`, not a wrong price.
- A pool that has **re-priced** produces a refusal, not a rate from the previous regime.
- No archive-node requirement remains in the oracle path.
- `GRAPH_API_KEY` absent → the backend still boots; only USDC quoting fails, naming the var.

EmilFattakhov:
**Part of #742 (USDC Payments for PayWithAI3 epic).**
**Closes #746 — Oracle service: AI3/USD price module.**

Step 4: a self-contained backend module that yields a trustworthy AI3/USD rate for the
USDC pricing #747 will add to `createIntent`. **Additive and dormant** — nothing imports
it yet, so behaviour is unchanged until #747 wires it in.

## Scope note
The issue proposed multiple sources with median + divergence detection. This was
**deliberately narrowed to a single source (CoinGecko)**: CoinGecko is itself a
volume-weighted aggregate across venues, and one well-known source keeps the module
simple. Median and cross-source divergence are therefore intentionally omitted — the
trustworthiness guarantees come from sanity bounds, staleness checks, a TTL cache with a
last-good fallback, and a re-fetch throttle.

## Units contract (from step 1)
The oracle returns AI3/USD as a **`bigint` scaled by `USD_RATE_SCALE` (1e18)** — the exact
representation `intents.usd_rate_at_creation` stores — so downstream USDC quote math stays
integer-only.

## Module — `apps/backend/src/infrastructure/services/priceOracle/`
- **`coingecko.ts`** — CoinGecko adapter: `fetchCoingeckoQuote(signal)` (public
  `simple/price` endpoint, no key) + `parseCoingeckoResponse` (zod-validated: `usd`
  positive & finite).
- **`quote.ts`** (pure — the bulk of the tests) — `parseDecimalToScaledBigint` (exact
  decimal→1e18 via **string math, never float**; `Number(x)*1e18` loses precision above
  ~$0.009), `isWithinBounds`, `isQuoteFresh`.
- **`index.ts`** — singleton `getPrice(): Promise<Result<OraclePrice, OracleUnavailableError>>`
  (neverthrow).
- **`types.ts`** — `OraclePrice`, `RawQuote`, `OracleUnavailableError`.

## Behaviour — `getPrice()`
On-request (lazy); no background worker or timer. Each call resolves in order:
1. **Fresh cache hit** while within `cacheTtlMs` (60s) — no network call.
2. **Throttled** — if a recent fetch failed, serve the last-good value (or error) without
   re-hitting CoinGecko.
3. **Single-flight** — concurrent callers share one in-flight refresh.
4. **Refresh** — fetch (per-request timeout + `AbortController`), validate against sanity
   bounds and per-quote staleness, then cache + return.

On a CoinGecko outage it serves the **last-good price** (flagged `stale: true`) for up to
`maxStaleMs` (10 min), re-polling at most once per `cacheTtlMs`; past that horizon it
returns `err(OracleUnavailableError)` so USDC quoting is refused. A failed fetch never
overwrites the cache, and `fromCache: true` means only a fresh TTL hit.

## Pricing / margin
- `shared/utils/pricing.ts` — `applyMarginPercent(amount, percent)` (basis points,
  **ceil-rounded so it never undercharges**; guards negative / NaN / Infinity).
- **The oracle returns the RAW market rate** (to be stored as `usdRateAtCreation`); the
  **5% `USD_QUOTE_MARGIN`** (env %, default 5) is applied to the *charged* amount in
  **#747**, not here. The helper is added and tested now but intentionally uncalled.
- New `config.priceOracle` block + `config.credits.usdQuoteMarginPercent`, documented in
  `.env.sample`.

## Design decisions
| Decision | Choice |
| --- | --- |
| Source | CoinGecko only (keyless); single-source is intentional |
| Median / divergence | Omitted (moot for one source; bounds + staleness + fallback instead) |
| Fetch model | On-demand `getPrice()`; background warm-refresh loop deferred to #749 |
| Rate representation | `bigint` scaled by `USD_RATE_SCALE` (1e18) |
| Margin | Oracle returns raw rate; `USD_QUOTE_MARGIN` (%) applied in #747 |

## Verification
- ESLint clean; `tsc --noEmit` 0 errors.
- **32 unit tests, all passing** — `quote` (precision incl. values > 2^53, bounds,
  staleness); `coingecko` (parser vs captured live fixtures); `getPrice` (cache hit/miss,
  TTL refresh, last-good fallback, outage throttle, unavailable / too-stale errors,
  single-flight); `fetchQuote` (bounds + staleness drop, reject→null, timeout); margin math.
  **No live HTTP in tests.**
- These specs don't require the TestContainers global-setup; CI runs the full suite as usual.

## Deferred
- Calling `getPrice()` + applying `USD_QUOTE_MARGIN` in `createIntent` → **#747**.
- Server-startup wiring and an optional background warm-refresh loop → **#749**.

### DIFF
--- a/apps/backend/.env.sample
+++ b/apps/backend/.env.sample
@@ -82,36 +82,27 @@ CREDIT_EXPIRY_CHECK_INTERVAL=3600000
 INTENT_EXPIRY_MINUTES=10
 
 # ---------------------------------------------------------------------------
-# USDC payments — AI3/USD price oracle & quote margin
+# USDC payments — AI3/USD price oracle (CoinGecko) & quote margin
 # ---------------------------------------------------------------------------
 
-# Comma-separated price sources for the AI3/USD oracle. Only 'coingecko' is
-# implemented today. Default: coingecko.
-ORACLE_SOURCES=coingecko
-
-# Minimum healthy sources required to compute a fresh price. Below this the
-# oracle serves the last-good value (if within ORACLE_MAX_STALE_MS) or fails.
-# Default: 1.
-ORACLE_MIN_SOURCES=1
-
-# How long (ms) a freshly computed price is served from memory before the next
+# How long (ms) a freshly fetched price is served from memory before the next
 # refresh. Default: 60000 (1 min).
 ORACLE_CACHE_TTL_MS=60000
 
-# Longest (ms) the last-good price may be served as a fallback when a refresh
-# cannot gather enough healthy sources. Default: 600000 (10 min).
+# Longest (ms) the last-good price may be served as a fallback while CoinGecko
+# is unavailable. Default: 600000 (10 min).
 ORACLE_MAX_STALE_MS=600000
 
-# Per-source HTTP timeout (ms). Default: 5000.
+# CoinGecko HTTP timeout (ms). Default: 5000.
 ORACLE_FETCH_TIMEOUT_MS=5000
 
-# Drop a source's quote if its own reported update time is older than this (ms).
+# Drop the quote if CoinGecko's own reported update time is older than this (ms).
 # Default: 300000 (5 min).
 ORACLE_MAX_SOURCE_AGE_MS=300000
 
-# Sanity bounds for a single source's AI3/USD price, as plain decimals (NOT
-# exponential notation). A value outside [min, max] is treated as a glitch and
-# dropped. Defaults: 0.0001 and 100.
+# Sanity bounds for the AI3/USD price, as plain decimals (NOT exponential
+# notation). A value outside [min, max] is treated as a glitch and dropped.
+# Defaults: 0.0001 and 100.
 ORACLE_MIN_USD_PER_AI3=0.0001
 ORACLE_MAX_USD_PER_AI3=100
 
--- a/apps/backend/.env.sample
+++ b/apps/backend/.env.sample
@@ -80,3 +80,42 @@ CREDIT_EXPIRY_CHECK_INTERVAL=3600000
 # expired. Users must submit their on-chain transaction within this window
 # after creating an intent. Default: 10.
 INTENT_EXPIRY_MINUTES=10
+
+# ---------------------------------------------------------------------------
+# USDC payments — AI3/USD price oracle & quote margin
+# ---------------------------------------------------------------------------
+
+# Comma-separated price sources for the AI3/USD oracle. Only 'coingecko' is
+# implemented today. Default: coingecko.
+ORACLE_SOURCES=coingecko
+
+# Minimum healthy sources required to compute a fresh price. Below this the
+# oracle serves the last-good value (if within ORACLE_MAX_STALE_MS) or fails.
+# Default: 1.
+ORACLE_MIN_SOURCES=1
+
+# How long (ms) a freshly computed price is served from memory before the next
+# refresh. Default: 60000 (1 min).
+ORACLE_CACHE_TTL_MS=60000
+
+# Longest (ms) the last-good price may be served as a fallback when a refresh
+# cannot gather enough healthy sources. Default: 600000 (10 min).
+ORACLE_MAX_STALE_MS=600000
+
+# Per-source HTTP timeout (ms). Default: 5000.
+ORACLE_FETCH_TIMEOUT_MS=5000
+
+# Drop a source's quote if its own reported update time is older than this (ms).
+# Default: 300000 (5 min).
+ORACLE_MAX_SOURCE_AGE_MS=300000
+
+# Sanity bounds for a single source's AI3/USD price, as plain decimals (NOT
+# exponential notation). A value outside [min, max] is treated as a glitch and
+# dropped. Defaults: 0.0001 and 100.
+ORACLE_MIN_USD_PER_AI3=0.0001
+ORACLE_MAX_USD_PER_AI3=100
+
+# Margin (percent) added on top of the raw oracle-derived USD cost when quoting
+# a USDC payment. The stored AI3/USD rate remains the raw market rate; only the
+# amount the user pays includes this margin. Default: 5.
+USD_QUOTE_MARGIN=5
--- a/apps/backend/__tests__/unit/priceOracle/aggregate.spec.ts
+++ b/apps/backend/__tests__/unit/priceOracle/aggregate.spec.ts
@@ -1,0 +1,93 @@
+import { describe, it, expect } from '@jest/globals'
+import { USD_RATE_SCALE } from '@auto-drive/models'
+import {
+  parseDecimalToScaledBigint,
+  median,
+  isWithinBounds,
+  isQuoteFresh,
+} from '../../../src/infrastructure/services/priceOracle/aggregate.js'
+
+describe('priceOracle/aggregate', () => {
+  describe('parseDecimalToScaledBigint', () => {
+    it('scales integers', () => {
+      expect(parseDecimalToScaledBigint('1')).toBe(USD_RATE_SCALE)
+      expect(parseDecimalToScaledBigint('100')).toBe(100n * USD_RATE_SCALE)
+    })
+
+    it('scales an AI3-range decimal exactly', () => {
+      // 0.00639777 USD/AI3 -> 6_397_770_000_000_000 (scaled 1e18)
+      expect(parseDecimalToScaledBigint('0.00639777')).toBe(
+        6_397_770_000_000_000n,
+      )
+    })
+
+    it('keeps full precision above the float-safe integer range', () => {
+      // 0.05 * 1e18 = 5e16 > 2^53; string scaling stays exact
+      expect(parseDecimalToScaledBigint('0.05')).toBe(50_000_000_000_000_000n)
+    })
+
+    it('truncates fractional digits beyond the scale', () => {
+      const nineteenDecimals = '0.' + '1'.repeat(19) // 19 dp; scale is 18
+      expect(parseDecimalToScaledBigint(nineteenDecimals)).toBe(
+        BigInt('1'.repeat(18)),
+      )
+    })
+
+    it('rejects exponential notation, signs, and non-decimals', () => {
+      expect(() => parseDecimalToScaledBigint('1e-3')).toThrow()
+      expect(() => parseDecimalToScaledBigint('-1')).toThrow()
+      expect(() => parseDecimalToScaledBigint('')).toThrow()
+      expect(() => parseDecimalToScaledBigint('abc')).toThrow()
+    })
+  })
+
+  describe('median', () => {
+    it('returns the middle value for an odd count', () => {
+      expect(median([3n, 1n, 2n])).toBe(2n)
+    })
+
+    it('returns the integer mean of the two middles for an even count', () => {
+      expect(median([1n, 2n, 3n, 4n])).toBe(2n) // (2 + 3) / 2 = 2 (floor)
+      expect(median([10n, 20n])).toBe(15n)
+    })
+
+    it('returns the single value for one element', () => {
+      expect(median([42n])).toBe(42n)
+    })
+
+    it('does not mutate the input', () => {
+      const input = [3n, 1n, 2n]
+      median(input)
+      expect(input).toEqual([3n, 1n, 2n])
+    })
+
+    it('throws on an empty list', () => {
+      expect(() => median([])).toThrow()
+    })
+  })
+
+  describe('isWithinBounds', () => {
+    it('is inclusive of both bounds', () => {
+      expect(isWithinBounds(10n, 10n, 20n)).toBe(true)
+      expect(isWithinBounds(20n, 10n, 20n)).toBe(true)
+    })
+
+    it('rejects values outside the bounds', () => {
+      expect(isWithinBounds(9n, 10n, 20n)).toBe(false)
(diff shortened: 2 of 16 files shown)


EmilFattakhov:
bugbot run

apps/backend/.env.sample
```diff
@@ -80,3 +80,33 @@ CREDIT_EXPIRY_CHECK_INTERVAL=3600000
...
+# Sanity bounds for the AI3/USD price, as plain decimals (NOT exponential
+# notation). A value outside [min, max] is treated as a glitch and dropped.
+# Defaults: 0.0001 and 100.
+ORACLE_MIN_USD_PER_AI3=0.0001
+ORACLE_MAX_USD_PER_AI3=100
+
+# Margin (percent) added on top of the raw oracle-derived USD cost when quoting
+# a USDC payment. The stored AI3/USD rate remains the raw market rate; only the
+# amount the user pays includes this margin. Default: 5.
+USD_QUOTE_MARGIN=5
```
EmilFattakhov, line 112: Open to your suggestions on the default values here Jim, these could be easily adjusted in production later on though.

jim-counter: 5% is a great default and where I hope we end up on live. I'm going to suggest we launch with a 10% override though as we are still early and subject to volatility. Then we adjust down over time.

jim-counter (changes requested):
Code looks good but there seems to be some drift in the PR description that should be fixed up before approving.

EmilFattakhov:
Yeah sorry about that, the implementation in further commits has shifted from the initial one while the PR description stayed the same, adjusted!

jim-counter (approved):
LGTM!

Issue #781, opened 2026-07-10
feat(s3): advertise WORM honestly - Object Lock via versioned-bucket semantics, versionId = CID


jim-counter:
**Depends on:** #778 (soft-delete / mapping-owner groundwork must be merged first)
**Related:** the per-user namespace follow-up proposed on #778 - both want the same append-only mapping migration and should be sequenced together.

## Background

Data on the Autonomys DSN cannot be destroyed - that permanence is one of Auto Drive's differentiators. #778 makes the S3 *namespace* mutable (soft-delete, copy/move, overwrite) and, to avoid promising a lock it no longer enforced at the key level, changed the Object Lock endpoints from advertising COMPLIANCE/forever to reporting "no lock configured".

That's honest for what #778 ships, but it undersells us: we now advertise *less* immutability than we actually have. There's a precedent-backed way to advertise the permanence truthfully.

## Key insight: S3 Object Lock never meant "deletes fail"

In AWS, Object Lock requires **versioning**, and on a versioned bucket a COMPLIANCE-locked object's `DeleteObject` *succeeds* - it writes a delete marker and hides the key. What the lock forbids is destroying a **version** (`DeleteObject` with a `versionId`). Overwrites stack new versions; nothing is ever destroyed.

Auto Drive natively implements exactly these semantics:

| S3 versioned-WORM concept | Auto Drive reality |
|---|---|
| Delete marker hides the key | soft-delete (`deleted_at`), shipped in #778 |
| Overwrite stacks a new version | remap; old CID persists on the DSN |
| Versions are indestructible | DSN permanence, enforced by construction, not policy |
| Version ID | **the CID**, verifiable and permanently retrievable |

The 403-on-versionId case is the differentiator: it isn't a policy refusal, the platform is *incapable* of destroying a version. MinIO and Backblaze B2 implement Object Lock the same versioning-based way; this is also the mechanism behind SEC 17a-4 / FINRA WORM storage.

## Proposed surface

- `GetBucketVersioning` -> `Enabled`
- `GetObjectLockConfiguration` -> `Enabled`, default retention COMPLIANCE (restore the year-9999 "forever" sentinel removed in #778 - this time it's true)
- `GetObjectRetention` -> COMPLIANCE, retain-forever, per version
- `DeleteObject` (no versionId) -> 204 + delete-marker semantics (already shipped)
- `DeleteObject?versionId=` -> 403 AccessDenied (truthful: versions cannot be destroyed)
[...]

PR #782, opened 2026-07-10
feat(s3): scope the S3 key namespace per user

EmilFattakhov:
Stacked on #778 (base branch `feat/s3-soft-delete-rclone`).

This implements the **per-user S3 namespace** proposed by @jim-counter in their review of #778 — the diagnosis and the design are theirs, not mine. Issue #781, which they opened, sequences this same owner-scoped mapping migration alongside the WORM/versioning work; this PR delivers the namespace half of it (with a couple of safety tweaks to the migration SQL, called out below).

## The diagnosis (credit: @jim-counter)

Reviewing #778, @jim-counter pointed out that nearly every hard problem in it — the dedup co-owner gates on delete/copy, the guarded upsert + null-return protocol, `destinationOwnedByAnother`, the cross-owner `AccessDenied` paths, the finalize-before-guard reorder — existed only to referee contention over a single global `(bucket, key)` namespace shared by every user. All of them are symptoms of one root cause: the namespace had no owner dimension while everything layered on top of it did.

## The proposal

Make the mapping identity **`(owner, bucket, key)`** — real S3's model, where bucket namespaces live inside an account. Contention becomes impossible, the guard machinery deletes itself, and the first-writer-wins squatting UX (a second user PUTting `default/test.txt` gets 403; soft-delete never releases the name) disappears. This PR implements exactly that — **net ~480 fewer lines** (+579 / −1061).

## Behaviour

- Two users can each own `default/test.txt` independently — no squatting, no 403-on-collision.
- Every S3 read/write is scoped to the caller (`GET`/`HEAD`/`ListObjectsV2`/`ListBuckets` too, so a user only sees their own keys). **Public sharing by CID (download API) is unaffected.**

## Removed (unreachable once per-user)

Guarded `ON CONFLICT` upsert + null protocol → plain owner-scoped upsert; `destinationOwnedByAnother` + both pre-finalize checks; the owner-vs-CID gate branching in `deleteObject`/`copyObject` (scoped `findByKey` is the boundary); `getMappingOwner`; the dual scoped/unscoped `countActiveMappingsByCid`; the `ForbiddenError → AccessDenied` branches for put/complete/copy. `AbortMultipartUpload` keeps its cross-user 403 (uploads aren't namespaced by key).

## Migration

Swaps the PK to `(owner_oauth_provider, owner_oauth_user_id, bucket, key)`. Remaining NULL-owner legacy rows (multi-admin dedup) are **duplicated** to every admin owner — losing a user's key is worse than surprise clutter they can soft-delete.

Three hardening changes over the SQL sketched in the proposal:
1. **Drop the old PK *before* the duplication INSERT** — inserting duplicate `(bucket, key)` rows while that PK is still enforced would fail.
2. **Fail loud** (`RAISE EXCEPTION`) on NULL-owner rows whose CID has no admin owner, instead of silently dropping them in the `DELETE`.
3. **A reversible `down`** — the forward migration deliberately allows multiple owners per `(bucket, key)`, so the rollback first collapses duplicates (keeping the most-recently-updated row) before restoring the `(bucket, key)` PK, and uses `DROP CONSTRAINT IF EXISTS` for idempotency. Lossy but deterministic, so a rollback — and the test suite's per-suite `down`/`up` reset — never errors on the very data the forward migration was built to allow.

Expected NULL-owner count in production is ~0 (single-admin CIDs were backfilled in #778); a pre-flight `COUNT(*)` confirms whether the legacy path is even exercised.

## Tests

- Unit suite rewritten around the scoped use cases (the cross-owner guard tests are gone; scoping is asserted via the owner args passed to the repo) — **34 passing, Docker-free**.
- Integration test gains a **two-users-same-key isolation** case (each sees their own content; one user's delete doesn't touch the other's identically-named key).
- e2e helpers and call sites updated to the owner-scoped signatures.
- Build, lint, and the full test-inclusive typecheck are clean.

## Verification caveat

The migration (PK swap + row duplication) and the integration/e2e suites need a **staging/CI run against real data + Docker** — testcontainers can't run in my environment. Please run the backend integration/e2e suite and the pre-flight NULL-owner `COUNT(*)` before merging.

### DIFF
--- a/apps/backend/__tests__/e2e/objects/objectRemoval.spec.ts
+++ b/apps/backend/__tests__/e2e/objects/objectRemoval.spec.ts
@@ -146,7 +146,7 @@ describe('Object removal', () => {
 
     it('is listed and retrievable over the S3 API (rclone/aws-cli)', async () => {
       expect(await listS3Keys(owner)).toContain(S3_KEY)
-      const getResult = await S3UseCases.getObject({
+      const getResult = await S3UseCases.getObject(owner, {
         Bucket: S3_BUCKET,
         Key: S3_KEY,
       })
@@ -241,7 +241,7 @@ describe('Object removal', () => {
 
     it('is hidden from S3 listing and not retrievable over S3', async () => {
       expect(await listS3Keys(owner)).not.toContain(S3_KEY)
-      const getResult = await S3UseCases.getObject({
+      const getResult = await S3UseCases.getObject(owner, {
         Bucket: S3_BUCKET,
         Key: S3_KEY,
       })
@@ -307,7 +307,7 @@ describe('Object removal', () => {
 
     it('is listed and retrievable over the S3 API again', async () => {
       expect(await listS3Keys(owner)).toContain(S3_KEY)
-      const getResult = await S3UseCases.getObject({
+      const getResult = await S3UseCases.getObject(owner, {
         Bucket: S3_BUCKET,
         Key: S3_KEY,
       })
--- a/apps/backend/__tests__/e2e/objects/objectRemoval.spec.ts
+++ b/apps/backend/__tests__/e2e/objects/objectRemoval.spec.ts
@@ -25,8 +25,11 @@ const S3_KEY = 'dir/removed.txt'
 
 const FOLDER_S3_BUCKET = 'folder-removal-bucket'
 
-const listS3Keys = async (bucket: string = S3_BUCKET): Promise<string[]> => {
-  const result = await S3UseCases.listObjects({
+const listS3Keys = async (
+  user: UserWithOrganization,
+  bucket: string = S3_BUCKET,
+): Promise<string[]> => {
+  const result = await S3UseCases.listObjects(user, {
     bucket,
     prefix: '',
     delimiter: null,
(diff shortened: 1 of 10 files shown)


jim-counter (approved):
Nice. Less code is almost always a good sign.

## Verification run locally (Docker)

Since the PR flagged that the migration and the integration/e2e suites could not be run in the author's environment, we ran them here before approving.

**Full backend test suite (TestContainers, real Postgres + RabbitMQ): 35/35 suites,
567/567 tests pass.** This includes the new two-user isolation integration tests, the e2e
removal/restore flows, and each suite's migration up/down reset, so the new migration was exercised repeatedly under test cycling.

**Migration duplication path, verified on a scratch Postgres 16 with seeded legacy data**
(a NULL-owner mapping whose CID has two admin owners, one of them trashed, plus a normally owned control row):

- `up`: the NULL-owner key was duplicated to both admin owners, including the trashed one (correct: their key must survive a later restore). The control row was untouched, the NULL originals removed, and the composite PK `(owner_provider, owner_user, bucket, key)` installed.
- `down`: collapsed deterministically to one row per `(bucket, key)`. Lossy, as documented.
- `up` again: completed cleanly with no re-duplication. The full up/down/up cycle is safe.

**Fail-loud orphan guard, verified:** a NULL-owner row whose CID has no admin owner aborts the migration with the intended operator message, and the failed run rolled back atomically. The old `(bucket, key)` PK and all rows were intact afterwards, so no half-migrated state is possible. After resolving the orphan row, `up` completed normally.

PR #784, opened 2026-07-14
feat(admin): reported-files review page + resurface re-reported files

EmilFattakhov:
## Admin reported-files review page

Adds a dedicated moderation screen at `drive/admin/reported` and fixes several
bugs in the report → review → ban/dismiss lifecycle.

### What's new

- **Reported-files page** listing every flagged file, split into two
  independently paginated sections so a large history can never crowd pending
  items out of view:
  - **Needs Review** — reported files awaiting a decision (dismiss/ban per row)
  - **Review History** — previously dismissed or banned files, kept visible so
    either decision can be revisited (including **unban**, which is new)
- Wired into the sidebar and admin dashboard; the inline "Files to be reviewed"
  banner now links here instead of opening a one-off modal, so the count and
  the queue can't diverge. Confirmation modals for ban/unban/dismiss.

### Bug fixes

- **Re-reported files never resurfaced**: a file that was reported, dismissed,
  then reported again stayed hidden forever behind the lingering
  `report-dismissed` tag. `reportObject` now clears the dismissal first.
- **Nested files never surfaced**: the review queue and banner read from the
  roots-only `metadata_roots` view, so reporting a file inside a folder went
  nowhere. Both now read `metadata` directly, with `distinct_on` collapsing
  content-dedup duplicates.
- **Review-queue REST filter leaked reviewed files**: the exclude clause used
  array containment (`NOT tags @> …`), which only dropped rows carrying *all*
  excluded tags — banned-only or dismissed-only files reappeared in
  `GET /objects/to-be-reviewed/list`. Now uses array overlap (`NOT (tags && …)`).
- **Unban failed open**: the `banned` tag was cleared before the file-gateway
  call, so a gateway outage left the DB unbanned while the gateway still
  banned (and background sync silently re-tagged it). Gateway now goes first.
- **Tag updates could silently no-op**: `addTag`/`removeTag` guarded the
  all-rows UPDATE on one *arbitrary* metadata row; with several rows per
  `head_cid` (same content under different roots) holding divergent tags,
  unban/re-report could skip the write. Dedup now lives in the SQL per row.
- **Object details page looked like reporting failed**: the server component
  never refetched, so the reported state only appeared after a manual reload.
- Minor: a refetch failure no longer overwrites the success toast of an action
  that succeeded; banner fetches the count only (was discarding 200 rows per
  admin page load); shared UTC-aware date formatting; fixed copy-pasted log
  messages in `dismissReport`.

### Security / breaking change

- `GET /objects/to-be-reviewed/list` is now **admin-gated** (it was fully
  unauthenticated, unlike its sibling ban/dismiss endpoints, and exposed a
  moderation queue of other users' reported content). ⚠️ Non-admin callers of
  the published SDK's `getToBeReviewedList` will now get 403.

### Tests

- New `objectModeration.spec.ts` (14 tests, real Postgres via TestContainers,
  gateway mocked): full report/dismiss/ban/unban lifecycle — re-report
  resurfacing, fail-closed unban on gateway error, admin gating, multi-row
  head_cids, queue filtering.
- `metadata.spec.ts`: regression tests for per-row tag dedup, all-rows tag
  removal, and include/exclude filter semantics.


EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM.

Accessing the `to-be-reviewed` list via GraphQL without admin access was flagged but I don't think it's a problem.

Issue #787, opened 2026-07-15
S3: rclone integration — remaining work + ignore decisions (tracker)


jim-counter:
## Context

After #778 (soft-delete, server-side copy/move, abort-MPU, mtime) the S3 layer passes the vast majority of rclone's integration suite — the immutability/teardown failures that used to fail ~40 tests are gone, because rclone's per-test cleanup now succeeds via soft-delete.

Re-running the suite against the live endpoint (Autonomys provider: `list_version=2`, `force_path_style=true`, `list_url_encode=true`, `no_check_bucket=true`, `object_lock_supported=false`) leaves a small, well-understood set of failures. This issue is the tracker for the rclone-provider work: it links the remaining fixes and records the deliberate ignores so the eventual upstream rclone PR (see #698) has a stable target.

Related: #757 (Object Lock read contract), #758 (`encoding-type=url`), #759 (multipart ETag), #778 (soft-delete/copy/mtime), #779 (no bucket endpoints), #781 (versioned-WORM), #782 (per-user namespace).

## Remaining work (fixes)

- [x] **Metadata fidelity** — verbatim `Content-Type` + persist system/user metadata. → #786
- [x] **Versioning + honest WORM** — `versionId = CID`, `ListObjectVersions`, `GetBucketVersioning`, `DeleteObject?versionId → 403`, restore COMPLIANCE Object Lock. → #781
  - **Resolved:** `object_lock_supported: false` **stays**. Verified by removing it: 5 `Internal/ObjectLock` subtests fail (Retention, LegalHold, SetAfterUpload, Multipart, Presigned), because the lock is intrinsic — PUT lock ops are 501 and GOVERNANCE/bypass/presigned flows are unsupported. The provider YAML now carries that rationale.

- [ ] **`Content-Encoding` read-path corruption** — GET returns decompressed bytes; no `Content-Length`; weak ETag. → #814
- [ ] **Read-after-write gap during blockstore→nodes migration** — a read landing mid-migration fails with `Chunk not found` and an opaque HTTP/2 stream reset; makes `FsOpenChunkWriter` fail ~2 of 3 runs. → #815

Once both land, the rclone suite is green except the deliberate ignores below.

## Deliberate ignores (won't fix — rationale recorded)

**a. `Rmdir` is a no-op — `FsRmdirFull`, `FsRmdirNotFound`.**
[...]

PR #788, opened 2026-07-15
feat(s3): faithfully round-trip object metadata (#786)

### LINKED ISSUE #786: S3: faithfully round-trip object metadata (verbatim Content-Type + system/user headers)
## Why

The S3 layer persists only `Content-Type` and (since #778) `mtime`, and it mutates `Content-Type` on the way out. rclone's integration suite fails `ObjectMimeType`, `ObjectMetadata`, `FsCopy/Metadata`, and `Internal/Metadata` as a result. S3 clients expect object metadata to round-trip verbatim. Two related parts, likely one or two small PRs by the same owner.

Tracked by #787.

## Part 1 — Return `Content-Type` verbatim (small)

**Problem.** An object stored with `Content-Type: text/csv` comes back as `text/csv; charset=utf-8`. The `; charset=utf-8` is appended by Express: `@autonomys/file-server`'s `handleDownloadResponseHeaders` sets the header with `res.set('Content-Type', …)`, and Express auto-appends a charset to `text/*` values lacking one. It's a framework default, not deliberate.

**S3 standard.** S3 stores and returns `Content-Type` byte-for-byte; it never mutates it.

**Change.** On the S3 `GetObject`/`HeadObject` responses, emit the stored `Content-Type` exactly — e.g. `res.setHeader('Content-Type', storedType)` (Node, no charset mangling) instead of Express's `res.set`. The download path is shared with the browser/gateway download-by-CID, where a charset is mildly useful; scope the verbatim behaviour to the S3 responses, or apply globally (browsers default/sniff UTF-8 for text, so the risk is negligible).

**Files.** `apps/backend/src/app/controllers/s3/s3.ts` (get/head handlers) and/or `@autonomys/file-server` `handleDownloadResponseHeaders`.

**Unblocks.** `TestIntegration/FsMkdir/FsPutFiles/ObjectMimeType`.

## Part 2 — Persist system + user metadata (medium)

**Problem.** `Cache-Control`, `Content-Language`, `Content-Disposition`, and arbitrary `x-amz-meta-*` are dropped on write, so they don't return on GET/HEAD and aren't carried by `CopyObject`. Today only `x-amz-meta-compression`/`-encryption` are read (`getUploadOptions`).

**S3 standard.** Objects carry user metadata (`x-amz-meta-*`) and system metadata (`Cache-Control`, `Content-Language`, `Content-Disposition`, `Content-Encoding`), set once at write and returned verbatim; `CopyObject` copies or replaces metadata per `x-amz-metadata-directive`.

**Change.** Store metadata alongside the CID in the S3 mapping — a `metadata JSONB` column on `S3.object_mappings` (or discrete columns). This is the "metadata beside the content, not in the content hash" pattern S3-over-IPFS gateways use (e.g. Filebase): it keeps content-addressed dedup intact and matches S3's per-key semantics. Then:
- capture the headers in `getUploadOptions` / `putObjectHandler` and `createMultipartUploadHandler`;
- thread through `createMapping` (`putObject`, `completeMultipartUpload`);
- return them on GET/HEAD (read the mapping's metadata, emit the headers);
- carry them through `CopyObject`, honouring the metadata-directive.

**Caveats.**
- Reconcile a client `Content-Encoding` with Auto Drive's internal compression (`x-amz-meta-compression`) so the response doesn't double-advertise encoding.
- Multipart: metadata arrives on `CreateMultipartUpload`; stash it on the upload session and write it at completion.
- Honour a client-supplied `Content-Disposition` when present (today it is always computed for browser UX) — prefer the stored value, fall back to computed.

**Files.** migration (`S3.object_mappings` metadata column); `infrastructure/repositories/s3/objectMappings.ts`; `core/s3/index.ts`; `core/uploads/uploads.ts` (multipart carry); `app/controllers/s3/s3.ts`.

**Unblocks.** `ObjectMetadata`, `FsCopy/Metadata`, `Internal/Metadata` (+`GzipEncoding`).

## Acceptance criteria

- [ ] `GetObject`/`HeadObject` return the stored `Content-Type` byte-for-byte (no injected charset).
- [ ] `Cache-Control`, `Content-Language`, `Content-Disposition`, and arbitrary `x-amz-meta-*` round-trip on PUT→GET/HEAD (single-PUT and multipart).
- [ ] `CopyObject` carries metadata (COPY) and replaces it when directed (REPLACE).
- [ ] Client `Content-Encoding` and Auto Drive's internal compression don't conflict in the response.
- [ ] rclone `ObjectMimeType`, `ObjectMetadata`, `FsCopy/Metadata`, `Internal/Metadata` pass.
- [ ] Unit tests for capture → store → return and for the copy metadata-directive.

EmilFattakhov:
Closes #786. Unblocks the metadata-fidelity item in the #787 tracker.

The S3 layer persisted only Content-Type (as the IPLD node's mimeType) and mtime, and mutated Content-Type on the way out. S3 clients expect object metadata to round-trip verbatim. This fixes both parts.

## Part 1 — verbatim Content-Type
`GetObject`/`HeadObject` now emit the stored Content-Type via Node's `res.setHeader` (after the shared download-header helper), so Express's `; charset=utf-8` injection is overridden. `text/csv` comes back as `text/csv`; an explicitly-stored charset is preserved.

## Part 2 — persist system + user metadata
A `metadata jsonb` column on `"S3".object_mappings` holds `contentType`, `cacheControl`, `contentLanguage`, `contentDisposition`, `contentEncoding`, and a `userMetadata` map of `x-amz-meta-*` (excluding the reserved `mtime`/`compression`/`encryption`/`cid`). It's:
- captured on PutObject and CreateMultipartUpload,
- threaded through `createMapping` and staged for multipart in `"S3".multipart_upload_meta` (mtime relaxed to nullable; the mtime-only staging trio generalised to `setMultipartMeta`/`getMultipartMeta`/`deleteMultipartMeta`),
- returned verbatim on GET/HEAD,
- carried by CopyObject per `x-amz-metadata-directive` (COPY inherits source, REPLACE uses the request).

A stored `Content-Encoding` is only surfaced when the object isn't internally compressed, so it never double-advertises against Auto Drive's own deflate handling.

Migration is additive and reversible (columns dropped, mtime NOT NULL restored on down).

## Tests
Unit (capture/store/return, copy metadata-directive) + integration (verbatim Content-Type incl. explicit charset, system + user metadata, Content-Encoding, COPY/REPLACE, multipart). `tsc` + `eslint` clean; full backend suite green.

### DIFF
--- a/apps/backend/src/app/controllers/s3/http.ts
+++ b/apps/backend/src/app/controllers/s3/http.ts
@@ -119,20 +119,33 @@ export const getS3Method = (req: Request): string => {
   }
 }
 
-// Operations whose request body IS the stored object. S3 stores that payload
-// byte-for-byte and treats Content-Encoding as opaque metadata — it must never
-// inflate the body. Express's parser otherwise gunzips gzip/deflate bodies (so
-// the stored bytes would no longer match the declared encoding) and rejects
-// unknown encodings (br, zstd) with 415 before they can be stored.
-const BODY_STORED_VERBATIM = new Set(['PutObject', 'UploadPart'])
+// The S3 write ops that may carry a client Content-Encoding the body parser must
+// not see. Two reasons, handled identically:
+//  - PutObject / UploadPart: the request body IS the stored object, so it must be
+//    kept byte-for-byte — S3 never inflates a stored body; Content-Encoding is
+//    opaque metadata. Left alone, Express gunzips gzip/deflate bodies (stored
+//    bytes then wouldn't match the declared encoding) or 415s unknown encodings
+//    (br, zstd) before they can be stored.
+//  - CreateMultipartUpload / CopyObject: no stored body, but Content-Encoding
+//    rides along as pure metadata on an empty body the parser still inspects.
+//    body-parser checks the encoding before it notices the body is empty
+//    (Content-Length: 0 counts as "has body"), so 'br' → 415 and 'gzip' → 400
+//    (gunzip on empty input) before the handler could read it as metadata.
+// For all of them, neutralizeContentEncoding moves the header onto a
+// request-scoped stash that getObjectMetadata reads back.
+const CONTENT_ENCODING_WRITE_OPS = new Set([
+  'PutObject',
+  'UploadPart',
+  'CreateMultipartUpload',
+  'CopyObject',
+])
 
 // Move a client-declared Content-Encoding out of the way of the body parser for
-// those operations, preserving it under an internal header for metadata
-// capture, so the exact wire bytes are stored and later returned verbatim under
-// the same encoding (GetObject/HeadObject re-emit it). aws-chunked is AWS
-// transfer framing, not object content: leave it in place so the parser's
-// existing handling stands rather than storing the framing bytes as data.
-const neutralizeObjectBodyEncoding = (
+// those ops (see above), preserving it under a request-scoped stash for metadata
+// capture; GetObject/HeadObject re-emit it. aws-chunked is AWS transfer framing,
+// not object content: leave it in place so the parser's existing handling stands
+// rather than storing the framing bytes as data.
+const neutralizeContentEncoding = (
   req: Request,
   _res: Response,
   next: NextFunction,
@@ -142,11 +155,12 @@ const neutralizeObjectBodyEncoding = (
     typeof encoding === 'string' &&
     encoding.toLowerCase() !== 'identity' &&
     !encoding.toLowerCase().includes('aws-chunked') &&
-    BODY_STORED_VERBATIM.has(getS3Method(req))
+    CONTENT_ENCODING_WRITE_OPS.has(getS3Method(req))
   ) {
     // Stash on a request-scoped side channel (not a header) so it round-trips as
     // metadata without being forgeable over the wire, then hide the real header
-    // from the parser so the body is stored verbatim.
+    // from the parser (which would otherwise inflate a stored body, or 415/400 on
+    // an empty one).
     stashContentEncoding(req, encoding)
     delete req.headers['content-encoding']
   }
@@ -166,9 +180,10 @@ s3Controller.get(
 
 s3Controller.use(
   '/:key(*)',
-  // Runs before the parser: hides a client Content-Encoding on body-storing
-  // ops so the object bytes are stored verbatim (see above).
-  neutralizeObjectBodyEncoding,
+  // Runs before the parser: moves a client Content-Encoding aside for the write
+  // ops that carry it, so the parser neither inflates a stored body nor 415/400s
+  // on an empty one (see above).
+  neutralizeContentEncoding,
   // S3 object bodies are opaque binary payloads. Always read the request body
   // as raw bytes regardless of (or the absence of) a Content-Type header.
   // `type: '*/*'` is insufficient: type-is returns false when no Content-Type
--- a/apps/backend/src/app/controllers/s3/http.ts
+++ b/apps/backend/src/app/controllers/s3/http.ts
@@ -1,4 +1,4 @@
-import { raw, Router, Request, Response } from 'express'
+import { raw, Router, Request, Response, NextFunction } from 'express'
 import { asyncSafeHandler } from '../../../shared/utils/express.js'
 import { createLogger } from '../../../infrastructure/drivers/logger.js'
 import {
@@ -17,6 +17,7 @@ import {
   notImplementedHandler,
   putObjectHandler,
   uploadPartHandler,
+  STASHED_CONTENT_ENCODING_HEADER,
 } from './s3.js'
 
 const s3Controller = Router()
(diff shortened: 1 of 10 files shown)


EmilFattakhov:
Follow-up fix pushed in `0bf04e1b` — closes a gap in the Part 2 `Content-Encoding` handling found during review.

**The gap.** The S3 object route parses bodies with `raw()` at body-parser's default `inflate: true`. So a `PutObject`/`UploadPart` sent with `Content-Encoding: gzip`/`deflate` was **decompressed before storage** while the header survived into metadata — GET/HEAD then re-advertised the stored `Content-Encoding` over the inflated body, mismatching the bytes, `Content-Length` and ETag. Other encodings (`br`, `zstd`) were rejected with **415** before they could be stored. (`inflate: false` isn't the fix — it makes body-parser 415 *every* non-identity encoding.)

**The fix.** S3 stores bytes verbatim and treats `Content-Encoding` as opaque metadata. A pre-parser middleware now moves a client `Content-Encoding` into an internal header for the body-storing ops (`PutObject`/`UploadPart`), so the parser stores the exact wire bytes; `getObjectMetadata` reads it back and GET/HEAD re-emit it. `aws-chunked` (AWS transfer framing, not object content) is left untouched so the parser's existing handling stands rather than storing framing bytes as data.

**Tests.** The gzip test now asserts the GET body is the verbatim gzip bytes (not server-inflated) and gunzips cleanly; added a `br` round-trip that previously 415'd. Full s3-sdk suite (62) and core unit tests (39) pass.

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

apps/backend/src/app/controllers/s3/s3.ts
```diff
@@ -657,24 +780,26 @@ export const copyObjectHandler = async (req: Request, res: Response) => {
...
   const mtime = isReplace ? getMtime(req) : undefined
+  const metadata = isReplace ? getObjectMetadata(req) : undefined
 
   const result = await S3UseCases.copyObject(user, {
     SourceBucket: source.bucket,
     SourceKey: source.key,
     Bucket: bucket,
     Key: key,
     Mtime: mtime,
+    Metadata: metadata,
```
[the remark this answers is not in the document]
EmilFattakhov: I've verified this finding. This one is a false positive — the current behavior is correct and intentional, dismissing it

apps/backend/src/app/controllers/s3/s3.ts
```diff
@@ -77,6 +78,112 @@ const getMtime = (req: Request): string | null => {
...
+  // side channel; read it back from there. It cannot be forged: a client-sent
+  // header only ever lands in req.headers, never in the WeakMap. Other write ops
+  // (CreateMultipartUpload) still carry the real header, metadata-only (no body
+  // to inflate).
+  const contentEncoding =
+    headerString(req.headers['content-encoding']) ??
+    stashedContentEncoding.get(req)
+  if (contentEncoding) metadata.contentEncoding = contentEncoding
+
+  const userMetadata: Record<string, string> = {}
```
jim-counter, line 141: Real S3 enforces a 2 KB limit on total user metadata and rejects the write with `MetadataTooLarge`. Here the only bound is Node's ~16 KB header budget, and everything within it is persisted into the jsonb column and replayed as response headers on every GET/HEAD of the key.

Not a blocker, but a cheap guard here (sum the encoded key/value lengths and reject over 2 KB with a `MetadataTooLarge`-style error) would match client expectations and keep mapping rows small. SDKs and tools already handle that error because real S3 returns it.

apps/backend/src/app/controllers/s3/http.ts
```diff
@@ -118,6 +119,40 @@ export const getS3Method = (req: Request): string => {
   }
 }
 
+// Operations whose request body IS the stored object. S3 stores that payload
+// byte-for-byte and treats Content-Encoding as opaque metadata — it must never
+// inflate the body. Express's parser otherwise gunzips gzip/deflate bodies (so
+// the stored bytes would no longer match the declared encoding) and rejects
+// unknown encodings (br, zstd) with 415 before they can be stored.
+const BODY_STORED_VERBATIM = new Set(['PutObject', 'UploadPart'])
```
jim-counter: This set only neutralizes Content-Encoding for the two body-storing ops, but the `raw({ type: () => true })` parser below runs for every method on `/:key(*)`, and body-parser 1.20 (Express 4.21) checks the encoding before it knows the body is empty (Content-Length: 0 still counts as "has body").

That means the two ops that legitimately carry Content-Encoding as pure metadata never reach the handler:

- `CreateMultipartUploadCommand({ ContentEncoding: 'br' })` fails 415 (unsupported encoding); `'gzip'` fails 400 (gunzip errors on the empty POST body).
- `CopyObjectCommand({ MetadataDirective: 'REPLACE', ContentEncoding: ... })` fails the same way (empty-body PUT).

So the CreateMultipartUpload branch of `getObjectMetadata` (s3.ts:133, "still carry the real header, metadata-only (no body to inflate)") is unreachable for Content-Encoding: the parser fails on the header before noticing there is no body to inflate. The new integration tests skirt this - the multipart test sets ContentType/CacheControl/Metadata but not ContentEncoding, and the REPLACE test omits it too.

Suggestion: add `CreateMultipartUpload` and `CopyObject` to the set (their bodies aren't stored, but the header still needs moving aside; `getObjectMetadata` already falls back to the stash), fix the stale comment in s3.ts, and add the two missing tests:

```suggestion
const BODY_STORED_VERBATIM = new Set([
  'PutObject',
  'UploadPart',
  // No stored body, but these carry Content-Encoding as pure metadata and the
  // parser must not see it (415/400 on the empty body otherwise).
  'CreateMultipartUpload',
  'CopyObject',
])
```

(If you take this, the set name and the "body IS the stored object" comment above it want a small rework too.)

jim-counter (changes requested):
Couple of tweaks to consider.

apps/backend/src/app/controllers/s3/s3.ts
EmilFattakhov: Good suggestion. If the guard only covers single PUT, a client bypasses it via multipart or a REPLACE copy — inconsistent, and the jsonb row still bloats, adding a shared validator applied at all three (put, createMultuplartUpload, copyObject)

apps/backend/src/app/controllers/s3/http.ts
EmilFattakhov: The suggestions makes perfect sense, implemented!

EmilFattakhov:
Both comments addressed @jim-counter!

jim-counter (approved):
Tweaks look good. Thanks Emil!

PR #795, opened 2026-07-22
chore(publishing): add block-by-block confirmation-depth debug logs

EmilFattakhov:
Files were getting stuck in the "publishing" state in production with transactionManager repeatedly logging the same tx hash "in block N — awaiting 12 confirmations" at increasing block numbers, which points at repeated re-inclusion moving the confirmation target in lockstep with the head (and re-arming the timeout) so a tx never confirms nor fails.

Add logger.debug traces to make the confirmation lifecycle observable:
- per-tx submission line recording expected confirmationDepth + timeout
- block-by-block trace: head vs inclusion height, confirmations accumulated vs required, and blocks remaining
- RE-INCLUDED detector logging when the inclusion target advances
- canonical-hash MATCH/MISMATCH decision when depth is reached
- confirmation-watch-open and terminal "settled" lines per tx

Also surface a publish-batch status breakdown (debug + warn) so the previously opaque "Failed to publish nodes" throw explains why an object won't leave the publishing state.

Logging only; no behavior change. Enable with DEBUG_LEVEL=DEBUG.

### DIFF
--- a/apps/backend/src/infrastructure/services/upload/onchainPublisher/index.ts
+++ b/apps/backend/src/infrastructure/services/upload/onchainPublisher/index.ts
@@ -55,10 +55,47 @@ const publishNodes = async (cids: string[]) => {
     }
   })
 
+  logger.debug(
+    'Submitting %d transaction(s) to publish (unique cids=%d, repeated=%d, skipped-null=%d)',
+    transactions.length,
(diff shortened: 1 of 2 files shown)


jim-counter (approved):
LGTM!

PR #797, opened 2026-07-23
feat(uploads): self-heal uploads stranded in migrating

EmilFattakhov:
Migration is driven by a single fire-and-forget migrate-upload-nodes task enqueued once at upload completion, with no re-drive path. If that task is lost or fails all its retries (dead-lettered to the unconsumed frontend-errors queue), the row stays in `migrating` ("Processing" in the UI) forever — publishing and archival already have recovery loops, but migration did not (getPendingMigrations existed but was never scheduled).

That could be addressed manually (script is available on remote machine), but automation is always a way forward. 

Add a periodic migration-recovery loop mirroring publishingRecoveryJob:

- migrationRecoveryJob schedules a recover-migrations task, deferring while the task-manager queue is busy (isTaskQueueBusy) and capped per run.
- processMigrationRecovery re-enqueues migrate-upload-nodes for root uploads stuck in `migrating` past MIGRATION_RECOVERY_STALENESS_MS. Only genuinely idle uploads are touched, so an in-flight migration is never double-driven (processMigration is not concurrency-guarded).
- Each re-drive stamps updated_at=NOW() so a genuinely-failing upload is retried at most once per staleness window rather than every cycle. This relies on the set_timestamp trigger being AFTER UPDATE (a no-op), which leaves updated_at authoritative to explicit writes.
- getStuckRootMigrations / markMigrationRecoveryAttempt repo queries.
- Wired under taskManager.active in frontendWorker (not archiver-gated) — it only needs the task-manager consumer.

Tunables: MIGRATION_RECOVERY_INTERVAL_MS (5m),
MIGRATION_RECOVERY_MAX_PER_CYCLE (50), MIGRATION_RECOVERY_STALENESS_MS (30m).

### DIFF
--- a/apps/backend/src/core/uploads/migrationRecovery.ts
+++ b/apps/backend/src/core/uploads/migrationRecovery.ts
@@ -1,0 +1,82 @@
+import { uploadsRepository } from '../../infrastructure/repositories/uploads/uploads.js'
+import { EventRouter } from '../../infrastructure/eventRouter/index.js'
+import { createTask } from '../../infrastructure/eventRouter/tasks.js'
+import { createLogger } from '../../infrastructure/drivers/logger.js'
+import { config } from '../../config.js'
+
+const logger = createLogger('useCases:uploads:migrationRecovery')
+
+// Concurrency guard — skip if a recovery pass is already running.
+let isRunning = false
+
+/**
+ * Recovers uploads stranded in the MIGRATING state.
+ *
+ * Migration is driven by a single fire-and-forget `migrate-upload-nodes` task
+ * enqueued once at upload completion (see scheduleNodeMigration). If that task
+ * is lost or fails all of its retries — landing in the unconsumed
+ * frontend-errors queue — the row stays MIGRATING forever with nothing to
+ * re-drive it, unlike publishing and archival which already have recovery
+ * loops. This sweep re-enqueues the migrate task for any root upload that has
+ * been MIGRATING past the staleness window.
+ *
+ * Safety:
+ * - Only root uploads idle longer than `stalenessMs` are selected, so a
+ *   healthy in-flight migration is never double-driven (processMigration is
+ *   not concurrency-guarded and saveNodes does plain INSERTs).
+ * - Each re-drive stamps updated_at=now(), so a genuinely failing upload is
+ *   retried at most once per window rather than every cycle.
+ * - A successful migration deletes its upload rows (removeUploadArtifacts), so
+ *   recovered uploads drop out of the sweep automatically.
+ *
+ * Queue deferral (skip while the task-manager queue is busy) is handled by the
+ * scheduler in migrationRecoveryJob.ts, not here, so task-triggered runs
+ * always execute.
+ */
+const processMigrationRecovery = async (): Promise<void> => {
+  if (isRunning) {
+    logger.debug('Migration recovery already in progress, skipping')
+    return
+  }
+
+  isRunning = true
+  try {
+    await runRecoveryBatch()
+  } finally {
+    isRunning = false
+  }
+}
+
+const runRecoveryBatch = async (): Promise<void> => {
+  const { maxUploadsPerCycle, stalenessMs } = config.migrationRecovery
+
+  const stuckUploadIds = await uploadsRepository.getStuckRootMigrations(
+    stalenessMs,
+    maxUploadsPerCycle,
+  )
+
+  if (stuckUploadIds.length === 0) {
+    logger.debug('No stuck migrations found, skipping')
+    return
+  }
+
+  logger.warn(
+    'Found %d uploads stuck in migrating, re-enqueuing migrate-upload-nodes: %s',
+    stuckUploadIds.length,
+    stuckUploadIds.join(', '),
+  )
+
+  for (const uploadId of stuckUploadIds) {
+    EventRouter.publish(
+      createTask({ id: 'migrate-upload-nodes', params: { uploadId } }),
+    )
+  }
+
+  // Rate-limit: stamp updated_at so these uploads are not re-selected for a
+  // full staleness window, whether or not the re-driven migration succeeds.
+  await uploadsRepository.markMigrationRecoveryAttempt(stuckUploadIds)
+}
+
+export const MigrationRecoveryUseCases = {
+  processMigrationRecovery,
+}
--- a/apps/backend/src/core/uploads/migrationRecovery.ts
+++ b/apps/backend/src/core/uploads/migrationRecovery.ts
@@ -21,9 +21,12 @@ let isRunning = false
  * been MIGRATING past the staleness window.
  *
  * Safety:
- * - Only root uploads idle longer than `stalenessMs` are selected, so a
- *   healthy in-flight migration is never double-driven (processMigration is
- *   not concurrency-guarded and saveNodes does plain INSERTs).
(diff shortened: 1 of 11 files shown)


EmilFattakhov:
bugbot run

apps/backend/src/core/uploads/migrationRecovery.ts
```diff
@@ -0,0 +1,82 @@
...
+ * is lost or fails all of its retries — landing in the unconsumed
+ * frontend-errors queue — the row stays MIGRATING forever with nothing to
+ * re-drive it, unlike publishing and archival which already have recovery
+ * loops. This sweep re-enqueues the migrate task for any root upload that has
+ * been MIGRATING past the staleness window.
+ *
+ * Safety:
+ * - Only root uploads idle longer than `stalenessMs` are selected, so a
+ *   healthy in-flight migration is never double-driven (processMigration is
+ *   not concurrency-guarded and saveNodes does plain INSERTs).
```
jim-counter: This bullet is accurate about the hazard, and I think the staleness window alone doesn't
close it. Requesting one line before merge.

`isTaskQueueBusy` can't see an in-flight migration: `checkQueue`'s `messageCount` excludes
unacked messages, so a running `migrate-upload-nodes` doesn't make the queue look busy.
And `RABBITMQ_PREFETCH` defaults to 10 with a non-serialized consume callback
(rabbit.ts:101), so the sweep and a still-running migration coexist happily in one worker.
That means a migration legitimately still working past `stalenessMs` gets a second task.
The loop batches 100 nodes per round trip, so a large object is 10k+ batches and 30 minutes
is reachable.

Before this PR that couldn't happen: `createHandlerWithRetries` only republishes after the
original invocation has thrown, so runs were strictly sequential. This is the first path to
genuine same-upload concurrency, which is why I'd like it covered here rather than in a
follow-up.

The consequence is worse than duplicated work, because the insert isn't idempotent.
`migrateFromBlockstoreToNodesTable` writes through `saveNodes`
(infrastructure/repositories/objects/nodes.ts:45), a plain `INSERT ... VALUES %L` against
`nodes_pkey PRIMARY KEY (cid)`. If two runs insert disjoint subsets and then each hits the
other's rows, both fail, neither reaches `removeUploadArtifacts`, and the upload is left
MIGRATING with a partial set of `nodes` rows that makes every future attempt collide too.
A slow upload becomes a permanently unrecoverable one.

One line fixes it, in `saveNodes`:

    INSERT INTO nodes (cid, root_cid, head_cid, type, encoded_node, piece_index,
                       piece_offset, block_published_on, tx_published_on)
    VALUES %L
    ON CONFLICT (cid) DO NOTHING

`DO NOTHING` rather than `DO UPDATE`: node CIDs are content-addressed, so an existing row
either holds identical content or is a repeated node shared with another object, where
overwriting `root_cid`/`head_cid` would be wrong. Neither run rewrites those columns today,
so this preserves current semantics.

`nodes_pkey` is the only unique constraint on the table and this is the sole production
caller of the repo-level `saveNodes`, so the change is scoped to the migrate path.

Two things come free with it: the migrate task becomes retry-safe, and this recovery loop
starts working for uploads that are already stranded mid-insert. That population can't be
fixed by re-driving as things stand, and it's plausibly a chunk of what you're trying to
heal, since a single mid-insert failure plus the three immediate retries all collide the
same way. Worth a DB-backed assertion that calling `saveNodes` twice with overlapping CIDs
is a no-op.

If you'd rather not touch shared node-writing code in this PR, an advisory lock around
`processMigration` or a `stalenessMs` default set well clear of real migration times would
also close the regression. Any one of the three is fine by me.

jim-counter (changes requested):
One change required to ensure we don't introduce a problem with the same node being processed more than once.

apps/backend/src/core/uploads/migrationRecovery.ts
EmilFattakhov: Good catch, I agree that the staleness window alone doesn't close this. 

I verified the suggested fix with and run it by multiple agents, and confirmed it can't be applied, full explanation as follows:

> On the ON CONFLICT (cid) one-liner — it can't be applied as-is. nodes_pkey PRIMARY KEY (cid) was dropped in migration 20250924123851 and never replaced; cid now has only a non-unique index (idx_nodes_cid). ON CONFLICT (cid) needs a unique index to infer against, so it fails at plan time with 42P10: there is no unique or exclusion constraint matching the ON CONFLICT specification — it would break every saveNodes, not just concurrent ones. (The PRIMARY KEY (cid) in your note is from the init schema, before that drop — easy to confirm with \d nodes.) And because cid isn't unique anymore, the specific "both fail → permanently stranded" outcome doesn't occur either: concurrent runs duplicate rows instead of colliding, which the guard prevents.

Implemented fix: 

An in-process per-upload guard in `processMigration` — if a migration for that upload is already running, the second run is a no-op, so there's no concurrent double-insert. I used this rather than an advisory lock because a DB lock holding a dedicated connection per migration would risk pool exhaustion under prefetch (held connections + their query bodies → deadlock).

EmilFattakhov:
bugbot run

apps/backend/src/core/uploads/migrationRecovery.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  for (const uploadId of stuckUploadIds) {
+    EventRouter.publish(
+      createTask({ id: 'migrate-upload-nodes', params: { uploadId } }),
+    )
+  }
+
+  // Rate-limit: stamp updated_at so these uploads are not re-selected for a
+  // full staleness window, whether or not the re-driven migration succeeds.
+  await uploadsRepository.markMigrationRecoveryAttempt(stuckUploadIds)
```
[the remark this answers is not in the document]
EmilFattakhov: Bugbot over-rated the severity.
The duplicated rows hold identical content, so the object still reconstructs and downloads fine. No corruption, no data loss, nothing unrecoverable.

The heavy fix — the (cid, root_cid) unique index + a production dedupe migration — is over-engineering for this severity, and it carries real prod risk (mutating live data, CONCURRENTLY index build).

The fix I actually made is unusually cheap, and that changes the calculus:

It reuses an existing, currently-unused method (removeNodeByRootCid) — one call.
It's a no-op on the normal path: a first-time migration has nothing to delete, so the happy path is byte-for-byte unchanged. It only does anything on a re-drive.
It makes the self-heal loop actually idempotent — which is the property you want from a recovery loop by definition — and guarantees your 3,190-upload drain stays clean even if a few are partial.

EmilFattakhov:
bugbot run

apps/backend/src/core/uploads/migrationRecovery.ts
jim-counter: Ah, I missed that `nodes_pkey` was dropped. You're right, and my suggestion was wrong on semantics too, not just the missing index: `cid` is legitimately non-unique across `root_cid`s, so a unique index on `cid` shouldn't be added to make `ON CONFLICT (cid)` work. Retracting the rest of that comment along with it, since with no unique constraint there's no collision, no mutual failure, and no permanent MIGRATING stranding, so nothing here was blocking.

The per-upload skip-lock is solid, and rejecting the advisory lock was the right call. The pool is constructed without `max` so it defaults to 10, same as `RABBITMQ_PREFETCH`, so holding a lock connection per handler really would have deadlocked.

Thanks for closing the loop.

jim-counter (approved):
LGTM!

PR #798, opened 2026-07-23
fix(publishing): poll fallback so confirmation survives a silent new-…

EmilFattakhov:
…heads subscription

On-chain publishing marks a tx durably published only after `confirmationDepth` blocks build on its inclusion block, observed via a per-tx `api.rpc.chain.subscribeNewHeads` subscription. A WebSocket reconnect can leave that subscription silently dead -- the socket comes back and one-shot RPC calls keep working, but the subscription never re-delivers -- so `head >= inclusion + depth` is never observed. The tx then hangs until the timeout, is retried with the same nonce, re-included, and stalls again: the "stuck in publishing" incident.

Add a poll fallback alongside the subscription:

- Extract the confirmation/canonical decision into `runConfirmationCheck`, shared by the subscription (`head-sub`) and a new `startConfirmationPoll` that ticks `getHeader()` every `CHAIN_CONFIRMATION_POLL_INTERVAL_MS` (config `confirmationPollIntervalMs`, default 6000ms ~= 1 block). getHeader is a fresh request each tick, so it drives confirmation even when the subscription is dead. `confirmationChecking` serialises the two paths; `source` is logged so the trace shows the poll carrying a tx when `head-sub` lines stop.
- Make a failed `subscribeNewHeads` open non-fatal: it used to resolve ConfirmationError (-> nonce resync -> same-nonce retry, a step in the incident loop); now it logs and relies on the poll. The timeout remains the backstop.
- Log the WS connection lifecycle (connected/disconnected/error) in substrate.ts so a reconnect is visible in the trace.

Unit-tests the confirmation watch against a mocked polkadot api + fake timers: the poll confirms when the subscription is silent, a failed subscription open still confirms via the poll, the subscription path is unchanged, and a non-canonical inclusion block at depth resolves Reorged. `submitTransaction` is exported for these tests.

Scope: does not change the separate re-inclusion-livelock path (target chasing the head); that is a distinct fix if traces ever show `remaining` not reaching 0.

### DIFF
--- a/apps/backend/.env.sample
+++ b/apps/backend/.env.sample
@@ -20,6 +20,13 @@ CHAIN_CONFIRMATION_DEPTH=25
 # margin. Default 300000ms (5 min).
 CHAIN_TRANSACTION_TIMEOUT_MS=300000
 
+# Safety-net cadence (ms) for confirming published transactions. Confirmation is
+# normally driven by a new-heads subscription; if that subscription goes silent
+# after a WebSocket reconnect, the chain head is also polled on this interval so
+# confirmation still completes instead of stranding the tx in "publishing".
+# Default 6000ms (~1 block at 6s/block).
+CHAIN_CONFIRMATION_POLL_INTERVAL_MS=6000
+
 # ---------------------------------------------------------------------------
 # Pay-with-AI3 / purchased credits feature
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


EmilFattakhov:
bugbot run

jim-counter (dismissed):
LGTM. Seems that the insights from logging are pointing us to a number of these scenarios - great!

Something picked up in review that is outside this PR but another failure condition probably worth following up on:

`publishNodes` records nothing if any transaction in the batch fails, and the publishing-recovery sweep only picks up objects that have at least one node already recorded as published. So an object whose only batch fails ends up with zero published nodes, which makes it invisible to the sweep. Once its 3 task retries are exhausted it stays in "publishing" forever. Small files are the ones at risk, since their nodes fit in a single batch.

EmilFattakhov:
No code changes made, requesting a re-review @jim-counter

jim-counter (approved):

Issue #792, opened 2026-07-20
Issue: Backend worker constantly restarts


EmilFattakhov:
## Summary
The production `auto-drive-backend` systemd unit is misconfigured, causing the unit to restart every ~5 seconds (~2.7M restarts observed) and hammering the Docker daemon with `docker compose up -d`. Combined with a logrotate hook that runs `systemctl restart` on the whole stack, this destabilizes `backend-worker`, which restarts repeatedly (graceful `SIGTERM`, exit 0). The restarts interrupt the on-chain publishing pipeline mid-confirmation, so objects stall in the `publishing` state and never progress to `archived`.

## Impact
- `backend-worker` restarts roughly every ~30 min (container `restart: unless-stopped` brings it back after each SIGTERM).
- On-chain `publish-nodes` work is interrupted before `confirmationDepth` is reached, so `block_published_on` is never recorded → objects stuck in `publishing` (observed ~1268 recoverable nodes + 123 partial objects not draining).
- Docker daemon is hit with `docker compose up -d` every 5s continuously.
- `/var/log/auto-drive-backend.log` grew to ~23M lines of pure compose "Running" spam.

## Evidence
`journalctl -u auto-drive-backend.service`:
```
auto-drive-backend.service: Scheduled restart job, restart counter is at 2708404.
Stopped Auto Drive Backend Service.
Started Auto Drive Backend Service.
auto-drive-backend.service: Deactivated successfully.
```
(repeats every ~5s)

`docker inspect auto-drive-backend-worker-1`:
```
restarts=28 lastExit=0 oom=false
```
Exit `0` + `oom=false` = graceful SIGTERM, not a crash or OOM. The only `process.exit(0)` path in `frontendWorker.ts` is the `SIGTERM`/`SIGINT` shutdown handler.

## Root cause

**1. `/etc/systemd/system/auto-drive-backend.service`** has no `Type=`, so it defaults to `Type=simple`. `ExecStart` is `docker compose up -d`, which exits within seconds; systemd interprets that as the service dying and `Restart=always` + `RestartSec=5` re-runs it forever:
```ini
ExecStart=/usr/bin/docker compose ... up -d --remove-orphans
Restart=always
RestartSec=5
```
Also, `CPUQuota=90%` / `MemoryMax=6G` on this unit only bound the short-lived compose CLI cgroup, not the containers (which run under dockerd), so they are ineffective.

[...]

PR #799, opened 2026-07-24
fix(deploy): stop auto-drive-backend systemd restart loop (#792)

EmilFattakhov:
The auto-drive-backend systemd unit had no Type=, so it defaulted to Type=simple with ExecStart=`docker compose up -d`. Because `up -d` returns within seconds, systemd treated the service as dead and Restart=always/RestartSec=5 re-ran it every ~5s (~2.7M restarts), hammering dockerd. Combined with the logrotate postrotate `systemctl restart` hook, this bounced backend-worker repeatedly; each bounce interrupted the on-chain publish-nodes pipeline before confirmationDepth, stranding objects in the `publishing` state.

systemd/systemd:
- Type=oneshot + RemainAfterExit=yes: run compose once, then stay active(exited) instead of being seen as dead and restarted.
- Restart=on-failure, RestartSec=30; add ExecStop; order after docker.service (Requires + After).
- Drop CPUQuota/MemoryMax: they only bounded the short-lived compose CLI cgroup, not the dockerd-managed containers.

systemd/logrotate:
- Replace the stack-restarting postrotate hook with copytruncate, which is correct for systemd `append:` output (it keeps the same inode the service already holds open).

Rollout is manual on the host (daemon-reload + stop/start), per the issue.

Refs #792

### DIFF
--- a/systemd/logrotate
+++ b/systemd/logrotate
@@ -1,12 +1,14 @@
+# Use copytruncate: systemd holds this log open via its `append:` StandardOutput
+# fd, so a rename+create rotation would leave the service writing to the rotated
+# (now-invisible) inode. copytruncate keeps the same inode and needs no service
+# bounce. Do NOT re-add a `postrotate systemctl restart` hook here: it bounced
+# the whole stack daily and interrupted on-chain publishing (see issue #792).
 /var/log/auto-drive-backend.log {
     daily
     rotate 7
     compress
     missingok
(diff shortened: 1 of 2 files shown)


EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM. We should work this into the deploy pipeline when we're getting staging up and running.

PR #803, opened 2026-07-30
fix(admin): show purchasing wallet in full with a copy button

EmilFattakhov:
The Purchasing Wallet column on the admin credit screens was truncated to `0x123456…abcdef`, so the address the refund transfer has to go back to could not be read off the page — only hovered for a title tooltip.

- Render the full 42-char address in the Purchasing Wallet column on both the per-user credit screen and the account list, wrapped in the existing CopiableText atom so it can be copied in one click. Rows stay on one line (whitespace-nowrap) and the table containers already scroll horizontally.
- Show the destination wallet in full, with copy, in the combined-refund action bar (it replaces the truncated wallet that was only shown when batches spanned several refund groups).
- Show it again in the refund confirmation modal, where the admin is about to make the transfer and needs to paste the address into their wallet.

### DIFF
--- a/apps/frontend/src/components/views/AdminPanel/AdminCredits.tsx
+++ b/apps/frontend/src/components/views/AdminPanel/AdminCredits.tsx
@@ -27,6 +27,7 @@ import type {
   OverCapIntent,
 } from '../../../services/api';
 import Link from 'next/link';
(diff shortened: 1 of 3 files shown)


jim-counter (approved):
LGTM

PR #805, opened 2026-07-30
feat(backend): alert to Slack when a task exhausts its retries

EmilFattakhov:
## Problem

The three error queues have no consumer.

`createHandlerWithRetries` publishes a task to `frontend-errors` / `download-errors` / `publish-errors` once its retries are exhausted, but `Rabbit.subscribe` was never called for any of them. So the depth of those queues was never a backlog waiting to be drained — it was a growing tally of failures nobody would ever see.

Production had **186 messages on `download-errors`**. Worth noting what that number is and isn't: the queue is fed by three task types (`async-download-created`, `object-archived`, `populate-cache`), so it was never "186 frustrated users", and purging it would have fixed nothing.

There was also a structural reason this had never been wired up: the error queues were created implicitly by `publish` alone and were absent from the `queues` tuple in `rabbit.ts`, so they were not valid `subscribe` (or `getMessageCount`) targets in the first place.

## What this changes

A consumer that batches permanently failed tasks and posts them to a Slack incoming webhook.

Four design points, roughly in order of how badly getting them wrong would have hurt:

### Batching, not one post per message

The first deploy drains a backlog. Posting per message would have put 186 messages into the channel in a few seconds and trained everyone to mute it — which would leave us exactly where we started, only with more noise.

Failures are collected for `TASK_ERROR_ALERT_WINDOW_MS` (default 30 minutes) and sent as one message with a per-task-type breakdown, listing at most `TASK_ERROR_ALERT_MAX_ITEMS` (default 10) individually before collapsing the rest to `… and N more`.

```
:warning: 50 tasks permanently failed (async-download-created×47, populate-cache×3)
```
```
[download-errors] async-download-created downloadId=dl-8f2c
  reason: Error fetching file header: 404 Not Found
[download-errors] populate-cache cid=bafkr6igab4rn…olyy
  reason: Download stalled: no data received for 300s
… and 47 more
```

### The handler must never throw

`Rabbit.subscribe` nacks with `requeue=true` and **no backoff**, so anything that throws out of the callback spins in a hot redelivery loop — and worse, blocks the drain of the very backlog it is trying to report.

An unparseable message and a failed Slack post are both logged and acked instead. The trade-off is that an alert can be lost if the process dies inside the batching window — which, at a 30-minute window, is worth taking seriously. Both workers now flush pending alerts on `SIGTERM`/`SIGINT`, so an ordinary deploy does not swallow the alerts it may itself have caused; the download worker had no shutdown handler at all before this. A hard crash (OOM, `SIGKILL`) can still lose up to a window's worth. The underlying failures are in the application logs either way — only the Slack notification is at risk.

Two of the eight tests exist purely to pin this behaviour.

### Carry the reason, not just the task

The error-queue payload contained only the task, so a consumer could say *what* failed but not *why*. `error` and `failedAt` now travel on the message.

The consumer validates the payload **loosely on purpose** — a message that no longer parses as a known task is exactly the sort of thing worth alerting on, so re-validating against `TaskSchema` would hide it.

### Make the error queues real queues

Added to the `queues` tuple so they are asserted at channel setup and are valid `subscribe`/`getMessageCount` targets.

## Where it runs

Registered in **both** the download worker and the frontend worker. RabbitMQ delivers each message to a single consumer, so a second reader shares the work rather than double-alerting; the only effect is that a batching window may be split between the two, producing two smaller summaries instead of one.

The download worker's registration is ungated by feature flags, so it is the one process guaranteed to be draining these queues. The frontend worker's registration also sits above its flag checks, for the same reason: a task that exhausted its retries needs reporting regardless of which subsystems a given host runs.

## Config

Alerting is **off unless `SLACK_WEBHOOK_URL` is set**, so dev and test are unaffected (failures are still logged). The channel is encoded in the webhook URL, so there is no separate channel setting.

| Variable | Default | Notes |
| --- | --- | --- |
| `SLACK_WEBHOOK_URL` | — | unset disables alerting |
| `TASK_ERROR_ALERT_WINDOW_MS` | `1800000` | batching window (30 min); also the worst-case alert loss on a hard crash |
| `TASK_ERROR_ALERT_MAX_ITEMS` | `10` | failures listed before collapsing to a count |

## Testing

8 new unit tests in `__tests__/unit/eventRouter/taskErrorNotifier.spec.ts`, covering: task/subject/reason formatting; batching 50 failures into one message; the item cap and overflow count; a missing reason reported rather than omitted; no accumulation when disabled; no throw on an unparseable message; no throw when Slack delivery fails; and a fresh batch after flushing.

`yarn backend lint` and `yarn backend build` pass. 34 tests pass across `__tests__/unit/eventRouter` and `TaskManager.spec.ts`, so the added `error`/`failedAt` fields on the error-queue payload don't disturb existing expectations.

**Not verified:** no live Slack post — this has not been run against a real webhook. Worth one manual check against a test channel before relying on it, ideally by letting a task fail rather than by calling the notifier directly, so the queue wiring is exercised too.

## Note on behaviour change

Consuming a queue drains it. Once these alerts fire, the record of a permanently failed task lives in Slack and the application logs rather than sitting in RabbitMQ. That is the point — but if a durable record is wanted, a `task_failures` table would be a small follow-up.

## Follow-ups (not in this PR)

- The retry path republishes with `retriesLeft - 1` **immediately, with no delay or backoff**, so all retries are consumed within seconds of the first failure. For failures caused by a dependency being down for days (the current DAG indexer lag), retries cannot help by construction — this is why tasks reach the error queue so readily.
- `object-archived` failures deserve more attention than the download ones: `onObjectArchived` marks metadata archived and *then* strips local node data, so archival is a one-way door onto the gateway path with no `DBObjectFetcher` fallback behind it.

### DIFF
--- a/apps/backend/src/app/servers/downloadWorker.ts
+++ b/apps/backend/src/app/servers/downloadWorker.ts
@@ -26,6 +26,9 @@
 
   const shutdown = async () => {
     logger.info('Shutting down download worker...')
+    // Stop consuming before flushing: otherwise failures keep arriving while the
+    // final send is awaited and land in a batch created after the flush finished.
+    EventRouter.stopTaskErrors()
     const { flushTaskErrorAlerts } = await import(
       '../../infrastructure/eventRouter/taskErrorNotifier.js'
     )
--- a/apps/backend/src/app/servers/downloadWorker.ts
+++ b/apps/backend/src/app/servers/downloadWorker.ts
@@ -28,7 +28,7 @@
     logger.info('Shutting down download worker...')
     // Stop consuming before flushing: otherwise failures keep arriving while the
     // final send is awaited and land in a batch created after the flush finished.
-    EventRouter.stopTaskErrors()
+    await EventRouter.stopTaskErrors()
     const { flushTaskErrorAlerts } = await import(
       '../../infrastructure/eventRouter/taskErrorNotifier.js'
     )
--- a/apps/backend/src/app/servers/downloadWorker.ts
+++ b/apps/backend/src/app/servers/downloadWorker.ts
@@ -6,4 +6,10 @@
   await import('../../app/apis/worker.js')
   const { EventRouter } = await import('../../infrastructure/eventRouter/index.js')
   EventRouter.listenDownloadEvents()
+
+  // Consume the error queues here as well as in the frontend worker. This worker
+  // is ungated by feature flags, so it is the one process guaranteed to be
+  // draining them. RabbitMQ hands each message to a single consumer, so the
+  // second reader shares the work rather than double-alerting.
+  EventRouter.listenTaskErrors()
 })()
--- a/apps/backend/src/app/servers/downloadWorker.ts
+++ b/apps/backend/src/app/servers/downloadWorker.ts
@@ -4,7 +4,9 @@
   }
 
   await import('../../app/apis/worker.js')
-  const { EventRouter } = await import('../../infrastructure/eventRouter/index.js')
+  const { EventRouter } = await import(
+    '../../infrastructure/eventRouter/index.js'
+  )
   EventRouter.listenDownloadEvents()
 
   // Consume the error queues here as well as in the frontend worker. This worker
@@ -15,25 +17,44 @@
 
   // Batched alerts are already acked off the queue, so a restart inside the
   // batching window loses them. With a 30-minute window that is worth handling:
-  // deploys are exactly when tasks fail. `flushTaskErrorAlerts` is internally
-  // bounded (the webhook post has its own 10s timeout and never rejects), so it
-  // cannot stall the container's stop grace period.
+  // deploys are exactly when tasks fail.
+  //
+  // Every step is bounded, and the sequence as a whole is too. Until this handler
+  // existed SIGTERM terminated the process outright, so anything unbounded here
+  // is a straight regression: the drain makes up to five sends of 10s each, and
+  // cancelling a consumer and closing a channel are broker RPCs that wait for a
+  // reply the broker may never send. Docker allows 10s before SIGKILL.
   const { Rabbit } = await import('../../infrastructure/drivers/rabbit.js')
   const { createLogger } = await import(
     '../../infrastructure/drivers/logger.js'
+  )
+  const { startShutdownWatchdog, shutdownStep } = await import(
+    '../../shared/utils/shutdown.js'
   )
   const logger = createLogger('servers:downloadWorker')
 
   const shutdown = async () => {
     logger.info('Shutting down download worker...')
+    const stopWatchdog = startShutdownWatchdog(logger)
+
     // Stop consuming before flushing: otherwise failures keep arriving while the
     // final send is awaited and land in a batch created after the flush finished.
-    await EventRouter.stopTaskErrors()
+    await shutdownStep(logger, 'stop error consumers', 2_000, () =>
+      EventRouter.stopTaskErrors(),
+    )
     const { flushTaskErrorAlerts } = await import(
       '../../infrastructure/eventRouter/taskErrorNotifier.js'
     )
-    await flushTaskErrorAlerts()
-    await Rabbit.close()
+    // Ordered after the cancel but not dependent on it: this talks to Slack, so
+    // a broker that has stopped answering must not cost us the alerts.
+    await shutdownStep(logger, 'flush task error alerts', 4_000, () =>
+      flushTaskErrorAlerts(),
+    )
+    await shutdownStep(logger, 'close rabbit channel', 2_000, () =>
+      Rabbit.close(),
+    )
+
+    stopWatchdog()
     logger.info('Download worker shut down successfully')
     process.exit(0)
   }
--- a/apps/backend/src/app/servers/downloadWorker.ts
+++ b/apps/backend/src/app/servers/downloadWorker.ts
@@ -12,4 +12,29 @@
   // draining them. RabbitMQ hands each message to a single consumer, so the
   // second reader shares the work rather than double-alerting.
   EventRouter.listenTaskErrors()
+
+  // Batched alerts are already acked off the queue, so a restart inside the
+  // batching window loses them. With a 30-minute window that is worth handling:
+  // deploys are exactly when tasks fail. `flushTaskErrorAlerts` is internally
+  // bounded (the webhook post has its own 10s timeout and never rejects), so it
+  // cannot stall the container's stop grace period.
+  const { Rabbit } = await import('../../infrastructure/drivers/rabbit.js')
+  const { createLogger } = await import(
+    '../../infrastructure/drivers/logger.js'
+  )
+  const logger = createLogger('servers:downloadWorker')
+
+  const shutdown = async () => {
+    logger.info('Shutting down download worker...')
+    const { flushTaskErrorAlerts } = await import(
+      '../../infrastructure/eventRouter/taskErrorNotifier.js'
+    )
+    await flushTaskErrorAlerts()
+    await Rabbit.close()
+    logger.info('Download worker shut down successfully')
+    process.exit(0)
+  }
+
+  process.on('SIGTERM', shutdown)
+  process.on('SIGINT', shutdown)
 })()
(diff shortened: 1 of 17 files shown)


EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

apps/backend/src/infrastructure/eventRouter/taskErrorNotifier.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  logger.warn(
+    'Task %s on %s failed permanently (%s): %s',
+    parsed.data.id,
+    queue,
+    subjectOf(parsed.data),
+    parsed.data.error ?? '<reason not recorded>',
+  )
+
+  if (!slackNotifier.isEnabled()) {
```
jim-counter, line 256: Should we ever get here if the notifier is disabled? This means we drain the queue when we don't want to alert?

apps/backend/src/infrastructure/eventRouter/taskErrorNotifier.ts
```diff
@@ -0,0 +1,253 @@
...
+      ? `:warning: Task permanently failed: \`${failures[0].task.id}\``
+      : `:warning: ${failures.length} tasks permanently failed (${breakdown})`
+
+  const shown = failures.slice(0, config.slack.alertMaxItems)
+  const lines = shown.map(({ queue, task }) =>
+    [
+      `[${queue}] ${task.id} ${subjectOf(task)}`,
+      task.error ? `  reason: ${task.error}` : '  reason: <not recorded>',
+    ].join('\n'),
+  )
```
jim-counter, line 154: Slack incoming webhooks truncate bodies past [40,000 characters](https://docs.slack.dev/reference/methods/chat.postMessage/#truncating) and this looks unbounded. May not be a problem but worth being aware of.

apps/backend/src/app/servers/downloadWorker.ts
```diff
@@ -6,4 +6,38 @@
...
+  // is ungated by feature flags, so it is the one process guaranteed to be
+  // draining them. RabbitMQ hands each message to a single consumer, so the
+  // second reader shares the work rather than double-alerting.
+  EventRouter.listenTaskErrors()
+
+  // Batched alerts are already acked off the queue, so a restart inside the
+  // batching window loses them. With a 30-minute window that is worth handling:
+  // deploys are exactly when tasks fail. `flushTaskErrorAlerts` is internally
+  // bounded (the webhook post has its own 10s timeout and never rejects), so it
+  // cannot stall the container's stop grace period.
```
jim-counter: That is not quite right. `MAX_DRAIN_PASSES` is 5 and `SEND_TIMEOUT_MS` is 10s, so the flush alone can spend up to ~50s. On top of that, `stopTaskErrors()` awaits `channel.cancel` and `Rabbit.close()` awaits `channel.close()`, neither of which is bounded if the broker is unreachable or the socket is half-open - which is exactly the situation during a coordinated restart. This process previously had no handler at all, so SIGTERM killed it immediately; now it can sit there until SIGKILL.

Suggest a hard bound on the whole sequence. `withTimeout` already exists in apps/backend/src/shared/utils/timeout.ts, or a watchdog is even simpler:
```
const shutdown = async () => {
  setTimeout(() => process.exit(1), 15_000).unref()
  ...
}
```

jim-counter (changes requested):
Some medium items to address. The one below is a pre-existing problem and could be a follow up if you'd prefer:

**`getChannel` re-subscribe loop can crash the worker on reconnect** (`rabbit.ts:66-71`)

The loop discards the promise `subscribe` returns:

```ts
for (const callback of queueSubscriptions) {
  subscribe(queue, callback)
}
```

`ensureConsumer` attaches a `.catch` to the `consumerTag` promise, but `subscribe` re-raises it via `await ensureConsumer(...).consumerTag`, so a `channel.consume` failure during a reconnect ends up as an unhandled rejection. The backend registers no `unhandledRejection` handler and Node 20 defaults to `--unhandled-rejections=throw`, so the worker exits. Confirmed with a throwaway spec: mock `amqplib`, force a keepalive failure, reject `consume` on the new channel.

This predates the PR, but it takes each worker from 1 consumer to 4, so the exposure roughly quadruples. `listenTaskErrors` already guards the direct path the same way, which makes the gap here look accidental:

```ts
subscribe(queue, callback).catch((error) =>
  logger.error(error as Error, 'Failed to re-subscribe %s after reconnect', queue),
)
```

Note that this keeps the process alive but leaves the consumer dead until the next keepalive-driven reconnect, so it is a mitigation rather than a full fix. Fine as a standalone follow-up if you would rather not touch the driver here.

apps/backend/src/infrastructure/eventRouter/taskErrorNotifier.ts
EmilFattakhov: I re-read the file, and I believe we're getting here (confirmed by claude). It is placed after the per-failure `logger.warn` to log permanently failed task unconditionally, and only then hit the slack notifier.

apps/backend/src/infrastructure/eventRouter/taskErrorNotifier.ts
EmilFattakhov: Haven't thought about it, let me put a guard in place

apps/backend/src/app/servers/downloadWorker.ts
EmilFattakhov: Good catch, addressed!

jim-counter (commented):
Re-reviewed at `223d0a00`. `yarn backend lint` and `yarn backend build` pass, 505/505 unit tests across 40 suites.

All three fixes look right. Two notes on the reconnect commit: retrying with backoff rather than just catching is better than what I suggested, and the `subscriptions[queue].includes(callback)` check before each attempt closes a resurrection race I had noticed but not written up. The cached-rejected-`channelPromise` wedge and the `setInterval(keepAlive, ...)` rejection are both real finds, and both were likelier to bite than the case I reported.

Three things left over.

---

**`listenFrontendEvents` / `listenDownloadEvents` / `listenPublishEvents` still float their subscribe promise** (`eventRouter/index.ts:33`, `:36`, `:39`)

```ts
listenFrontendEvents: () => {
  Rabbit.subscribe('task-manager', processFrontendTask)
},
```

Same bug class this commit just fixed in the reconnect loop and the keepalive interval, on the startup path. If the broker is not up when a worker starts, `getChannel` rejects, `subscribe` re-raises, nothing catches, and the worker dies on `--unhandled-rejections=throw`.

Verified with a throwaway spec (mock `amqplib`, reject `connect`): `listenFrontendEvents` produces an unhandled rejection, `listenTaskErrors` logs it and carries on. Worth fixing now rather than later, because clearing a rejected `channelPromise` means a worker that survives startup will actually recover on the next keepalive tick, which was not true before this PR. Same `.catch` `listenTaskErrors` already has:

```ts
Rabbit.subscribe('task-manager', processFrontendTask).catch((error: unknown) =>
  logger.error(error as Error, 'Failed to consume task-manager'),
)
```

**The watchdog exits 1 during an intentional stop** (`shared/utils/shutdown.ts`)

`onExpiry` defaults to `process.exit(1)`. A slow shutdown is then indistinguishable from a crash to whatever is supervising the process, even though it was asked to stop and the steps that mattered are logged either way. Exit 0 with the existing warning reads truer. Minor, and a deliberate choice either way, so ignore if the intent is for a blown budget to show up as a failure.

**`clamp` does not squash newlines** (`taskErrorNotifier.ts`)

The character caps are the important half and they are in. But a reason is still interpolated with its own line breaks intact, so a multi-line `Error.message` breaks the one-line `reason:` layout and pushes the rest of the batch down the message. `task.error.replace(/\s+/g, ' ')` before clamping would keep one failure to one line. Cosmetic only.

---

Still open from the first pass, unchanged: the `slackNotifier.isEnabled()` gate on `listenTaskErrors` (a deploy without the webhook drains the backlog to log lines), and the unbounded `pending` array, which the new character caps do not cover since they bound the message rather than the heap.

EmilFattakhov:
Thanks for your review @jim-counter, your suggestions are implemented.

jim-counter (approved):
LGTM!

PR #806, opened 2026-07-30
fix(uploads): make completeUpload idempotent and stop the frontend-errors dead-letter loop

EmilFattakhov:
## Summary

The `frontend-errors` RabbitMQ queue crossed its 100-message alert threshold in production (332 → 350 while we were investigating), growing by exactly 6 messages every 30 minutes. This PR fixes the cause: **`completeUpload` is not idempotent**, so a retried complete call leaves an upload permanently unmigratable, and the migration recovery sweep then re-drives that broken upload forever, emitting one dead-letter message per upload per staleness window.

No data was lost — the affected payloads are intact in `uploads.blockstore`. This PR makes them migrate on the next sweep tick.

## How we got here

The alert:

```
[FIRING:1] RabbitMQ Queue Exceeds 100 Messages
Broker = auto-drive-rabbitmq-broker-primary
Queue  = frontend-errors      Value: B=332
```

### 1. `frontend-errors` is a terminal sink with no consumer

`Rabbit.subscribe` is typed to `'task-manager' | 'download-manager' | 'publish-manager'`, and `frontend-errors` is not one of them. `publish()` takes a plain `string`, so the code can write to that queue but is type-prevented from ever reading it. Anything landing there stays forever, so the depth is monotonic and the alert re-fires until someone drains it.

Messages arrive via `createHandlerWithRetries`: a failing task is re-published with `retriesLeft - 1`, and at zero it is published to the error queue with `retriesLeft` reset to `errorRetries` (3). Two consequences when reading the queue: the reset means you cannot tell how many attempts a message burned, and the error object is never included in the message — only the task. Only `logger.error(error, 'Task failed')` has the exception.

### 2. What was actually in the queue

Sampled on the production broker with `reject_requeue_true` (messages left in place):

```
324  migrate-upload-nodes   <- 6 uploadIds × 54 copies each
 14  publish-nodes          <- 7 node batches, older, not growing
  4  tag-upload
  1  recover-publishing
  1  reconcile-archival
```

94% is one bug repeated. The 14 `publish-nodes` messages are residue from before #794 moved on-chain publishing to `publish-manager`/`publish-errors`; they are not growing and are unrelated to this fix.

All six uploads are the same user's attempts at a 13-byte `test2.txt`, uploaded 2025-12-01 between 21:03 and 21:33 — six attempts in thirty minutes, i.e. someone retrying an upload that would not finish. All six were still `status = migrating` eight months later. No user was actively affected at the time of the alert; the growth was entirely a background sweep chewing on eight-month-old rows.

### 3. The failure

Each of those six uploads has **three duplicate `uploads.blockstore` rows** — same `cid`, same `node_type = 'File'`, same 13-byte payload. `getFileUploadIdCID` demanded exactly one:

```ts
if (blockstoreEntry.length !== 1) {
  throw new Error(`Invalid number of blockstore entries for file upload with id=${uploadId}`)
}
```

So `processMigration` → `getUploadCID` threw immediately, every time — a permanent data-shape problem where retrying can never help. Every root upload in `migrating` (6 of them; the other 65 `migrating` rows are non-root children) was broken this same way, all with `file_nodes = 3`.

### 4. Why the queue grew forever

1. `migrationRecoveryJob` (added in #797, deployed ~28h before the alert) enqueues `recover-migrations` every 5 minutes.
2. The sweep selects root uploads sitting in `migrating` past the 30-minute staleness window — all 6 qualify, every window.
3. Each `migrate-upload-nodes` fails, retries 3×, then dead-letters to `frontend-errors`.
4. The sweep stamps `updated_at = NOW()` on all six, so they go quiet for exactly one window and then requalify.

This was confirmed live rather than inferred: `updated_at` on all six read `18:46:40.509678`, then `19:16:40.516742` — 30 minutes apart to the millisecond, identical across all six rows — and the queue went 344 → 350 on that tick. 54 copies per upload ≈ 54 windows ≈ 27 hours, matching the #797 deploy.

#797 did not *create* the broken rows; it converted a silent stuck-row backlog into an unbounded visible dead-letter stream. Its docstring assumed "a successful migration deletes its upload rows, so recovered uploads drop out of the sweep automatically" — which never happens for an upload that can't migrate, and the sweep had no give-up counter or unrecoverable-state detection (unlike `publishingRecovery`, which already logs an `unrecoverable` bucket).

### 5. Root cause: `completeUpload` had no status precondition

`completeUpload` fetched the upload, checked permissions, and went straight into `completeUploadProcessing` — **with no check that the upload was still `PENDING`**, unlike `abortUpload` twenty lines above it. `completeFileProcessing` → `processBufferToIPLDFormatFromChunks` writes the root File node via `blockstore.put` → a bare `INSERT` with no `ON CONFLICT`, and `uploads.blockstore` has its primary key on the surrogate `sort_id` with only a **non-unique** btree `blockstore_upload_id_cid_index` on `(upload_id, cid)`.

So calling complete three times on one uploadId wrote the root File node three times. A client timeout retry, a double submit, or a proxy retry is enough — and it matches the evidence precisely (only the `File`-type root row is triplicated, not the chunk rows).

The same missing guard means a repeated complete also re-ran `handleFileUploadFinalization` → `AccountsUseCases.registerInteraction`, i.e. **it double-charged upload credits**. Worth checking `interactions` for duplicate rows on those six CIDs; trivial money at 13 bytes, but the path is general.

## What this PR changes

**1. `completeUpload` claims the upload atomically — the cause fix** (`core/uploads/uploads.ts`, `repositories/uploads/uploads.ts`, `packages/models`)

A repeat call on an already-`MIGRATING` upload is now a no-op that returns the existing CID and re-drives migration, so a client retry *heals* a stuck upload instead of corrupting it. This preserves S3 semantics: `CompleteMultipartUpload` is expected to succeed on retry, and both S3 call sites (`completeMultipartUpload`, `putObject`) get the right behavior. Any other terminal status is rejected. This also removes the duplicate credit debit.

A status *read* is not sufficient on its own, though — **thanks to Bugbot for catching this**: two overlapping calls (a client timeout retry while the first is still in flight) would both observe `PENDING` and both fall through into the processing, recreating the very shape this PR removes. Completion is therefore claimed with a conditional `UPDATE` (compare-and-swap) in `claimUploadForCompletion`, so exactly one caller proceeds no matter how many API processes are running.

Losing the claim is not proof that a completion is still running, though — **thanks to Cursor Bot for catching both of these**. The row is re-read on claim loss, and its state says which of four things happened while this call sat between its status read and the CAS:

| Re-read shows | Meaning | Outcome |
| --- | --- | --- |
| `MIGRATING` | winner finished the whole completion (an easy window to hit for a small upload) | return the existing CID and re-drive migration — same path as the fast path |
| row gone | a successful migration already removed the artifacts | not-found, exactly as the check at the top of the function |
| `PENDING` | winner failed and released the claim (insufficient credits, a gap in the stored parts) | nothing holds the claim, so re-attempt it **once** and do the work |
| `COMPLETING` / terminal | a completion really is in flight, or the upload is finished/cancelled | error, with the message reflecting the observed status |

The `PENDING` re-attempt is deliberately capped at one, so a sibling that keeps failing and releasing cannot turn this into a loop. Both already-completed entry points share `resolveAlreadyCompletedUpload`, so they cannot drift.

The claim uses a new `UploadStatus.COMPLETING` rather than flipping straight to `MIGRATING`, for two reasons:

- The migration recovery sweep selects on `MIGRATING`. A still-completing upload has no root node yet, so if the claim used `MIGRATING` the sweep could enqueue `migrate-upload-nodes` mid-completion — and with change (3) below that would park a perfectly healthy upload as `FAILED`. Any upload whose completion processing outlasts the 30-minute staleness window would hit this.
- `updated_at` must keep marking *`MIGRATING`-entry* for that staleness anchor (the invariant #797 relies on), which it does because the flip to `MIGRATING` still happens exactly where it did before, after finalization.

The claim is released back to `PENDING` when completion fails, so today's retry-after-failure behavior (e.g. `Not enough upload credits`) is unchanged, and a claim older than `UPLOAD_COMPLETION_CLAIM_STALE_MS` (default 1h) is re-claimable so a process dying mid-completion cannot strand the upload — nothing sweeps `COMPLETING`. `uploads.uploads.status` is plain `text`, so the new value needs **no DB migration**.

**2. Root-node CID lookup tolerates same-CID duplicates — recovers the six uploads** (`core/uploads/blockstore.ts`)

`getFileUploadIdCID`/`getFolderUploadIdCID` now resolve through a shared `getRootNodeCID` that keys on **distinct** CIDs rather than row count, logging a warning when duplicates are present. Zero rows, or genuinely different CIDs, still fail.

This is safe because migration already dedupes downstream — `migrateFromBlockstoreToNodesTable` builds `uniqueNodes` keyed by CID before `saveNodes`, so no duplicate `nodes` rows result. Deploying this lets all six production uploads migrate on the next sweep tick, with no manual row deletion and no data mutation.

**Deliberately not done: a unique index on `(upload_id, cid)`.** A file containing two identical chunks legitimately stores two rows with the same CID, and `processBufferToIPLDFormatFromChunks` iterates `getFilteredMany(FileChunk)` to build the DAG links — deduping there would silently corrupt such files (cf. `handleRepeatedNodes`). Tolerating duplicates at the root-node read site is the safe layer; the write path is fixed by (1). Per-chunk reads are untouched.

**3. Permanent failures are parked, not retried forever** (`core/uploads/errors.ts`, `core/uploads/uploads.ts`)

New `UnrecoverableUploadError` marks shapes that can never succeed (missing/ambiguous root node, deleted upload row). `processMigration` catches it, sets the upload tree to `FAILED` so `getStuckRootMigrations` stops selecting it, and acks the task instead of dead-lettering. Transient errors still throw and retry as before. `UploadStatus.FAILED` already exists in `@auto-drive/models`, so this needs no DB migration. Without this, the next differently-broken row reopens this incident.

**4. `FAILED` is not reported as success** (`core/objects/uploadStatus.ts`)

A `FAILED` upload still has its rows (only a successful migration removes artifacts) and no `nodes`, so the counting path would report 0 uploaded of 0 total — which reads as "fully published". `getUploadStatus` now returns the same all-null state it returns for in-progress uploads. Surfacing a genuine failed state to the user needs a new `ObjectUploadState` field plus frontend work; that is out of scope here.

## Tests

New `__tests__/unit/core/uploadCompletionIdempotency.spec.ts` (14 tests): complete-on-MIGRATING skips processing / returns the CID / re-drives migration; complete on a terminal status is rejected; a caller that loses the claim to a still-running winner never reaches the processing; a caller that loses the claim to an *already-finished* winner gets the CID and re-drives migration; a claim lost with the row gone reports not-found; a claim released back to `PENDING` is re-taken and processed; the re-attempt is capped at one; the winner claims *before* processing and only reaches `MIGRATING` after the root node exists; a failed completion releases the claim and never flips the status; duplicate same-CID root rows are accepted; ambiguous and missing root rows raise `UnrecoverableUploadError`; `processMigration` parks-and-acks an unrecoverable failure and still rethrows a transient one.

`__tests__/unit/useCases/uploads.spec.ts` — the existing concurrent-migration guard test asserted that a missing upload *rejects*; updated to the new park-and-resolve contract while still exercising the in-flight guard.

Verified locally: `yarn backend lint` clean, `tsc --noEmit` clean, and 30/37 unit suites pass (421 tests). The 7 failing suites are all DB-backed (`repositories/*`, `OnchainPublisher`, `objectModeration`) and fail identically on a clean tree in this sandbox — Docker/TestContainers is unavailable, so those and the e2e suites need CI.

## Rolling deploy note

An old process that sees `status = 'completing'` during a rolling deploy behaves exactly as it does today: `abortUpload` rejects it (it requires `PENDING`), and old `completeUpload` — which had no guard at all — would re-run processing on it, i.e. the pre-PR behavior for the duration of the rollout. `getUploadStatus` on old code reports a `COMPLETING` upload as 0-of-0 for that window. Nothing is corrupted that isn't already corruptible today, and the window closes when the rollout finishes.

## Deploy / follow-up

- After deploy, the six uploads migrate on the next sweep tick (≤30 min) and the queue stops growing. Only then is it safe to drain the 350 existing messages — purging first would refill within 30 minutes.
- **Ops, not code:** `frontend-errors` / `download-errors` / `publish-errors` are unbounded sinks with no consumer. They want `x-message-ttl` and `x-max-length`. Queue arguments are immutable in RabbitMQ, so on Amazon MQ this means delete-and-recreate after draining.
- Consider whether the alert should fire on queue *growth* rather than absolute depth, since a terminal sink's depth only ever rises.
- Known wart, pre-existing and left alone deliberately: on the REST route, `handleInternalError` flattens every thrown error into `InternalError`, so the new "already being completed" rejection surfaces as a 500 rather than a 4xx (the same is already true of the `BadRequestError` that `assertNoUnprocessedParts` throws). There is no Express error middleware, so on the S3 route it reaches the default handler. Worth fixing by preserving `HttpError` instances in `handleInternalError`, but that changes status codes across every route that uses it and does not belong in this fix.
- Nothing sweeps a stranded `COMPLETING` row before its 1h re-claim window; the equivalent today is a stranded `PENDING` row, which is equally unswept. A recovery pass for both is a reasonable follow-up.
- One correction for anyone debugging on the boxes: `/home/ubuntu/deploy/auto-drive` on the mainnet host is a stale checkout (`e0c47592`, 2026-06-23), which is why the deployed image looks "newer than git HEAD" and appears to contain code that isn't in the repo. `migrationRecovery.js`, `migrationRecoveryJob.js`, `recover-migrations`, and `publishWorker` are all on `main` (#794, #797). There is no unversioned code in production.

### DIFF
--- a/apps/backend/__tests__/unit/core/completionRetryIdempotency.spec.ts
+++ b/apps/backend/__tests__/unit/core/completionRetryIdempotency.spec.ts
@@ -10,11 +10,17 @@ import {
   DEFAULT_MAX_CHUNK_SIZE,
   MetadataType,
   cidToString,
+  stringToCid,
 } from '@autonomys/auto-dag-data'
 import { UploadType } from '@auto-drive/models'
-import { ok } from 'neverthrow'
+import { err, ok } from 'neverthrow'
 import { UploadFileProcessingUseCase } from '../../../src/core/uploads/uploadProcessing.js'
-import { UploadPartsChangedError } from '../../../src/core/uploads/errors.js'
+import {
+  UnrecoverableUploadError,
+  UploadPartsChangedError,
+} from '../../../src/core/uploads/errors.js'
+import { BlockstoreUseCases } from '../../../src/core/uploads/blockstore.js'
+import { ObjectNotFoundError } from '../../../src/errors/index.js'
 import {
   NodesUseCases,
   createNodeDeduplicator,
@@ -29,6 +35,7 @@ import { nodesRepository } from '../../../src/infrastructure/repositories/object
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
 const UPLOAD_ID = 'upload-retry-1'
+const ROOT_CID = 'bafkr6icio4rqi75syx2xkxwmnnnwx3gzmnbjmtnqoydtfcxrjuqvvxxs4u'
 
 // The root node types are the ones blockstore_root_node_unique_idx covers.
 const ROOT_NODE_TYPES = [MetadataType.File, MetadataType.Folder]
@@ -491,5 +498,48 @@ describe('migrateFromBlockstoreToNodesTable over repeated content', () => {
     expect(insertedBatches.every((batch) => batch.length > 0)).toBe(true)
     // Each distinct node written exactly once, across every batch.
     expect(insertedBatches.flat()).toHaveLength(distinctCids.size)
+  })
+
+  // The guard on this used to be `if (!metadata) return` against a neverthrow
+  // Result, which is always truthy — so it never fired and migration proceeded
+  // without metadata, writing nodes that nothing would ever publish. Unrecoverable
+  // is the honest verdict: the metadata is written before the upload reaches
+  // MIGRATING, so a missing row cannot be fixed by retrying.
+  it('parks an upload whose root has no metadata instead of migrating it', async () => {
+    installFakeBlockstore()
+
+    const uploadEntry = {
+      id: UPLOAD_ID,
+      root_upload_id: UPLOAD_ID,
+      type: UploadType.FILE,
+      name: 'orphan.bin',
+      upload_options: null,
+    } as any
+    jest
+      .spyOn(uploadsRepository, 'getUploadEntryById')
+      .mockResolvedValue(uploadEntry)
+    jest
+      .spyOn(uploadsRepository, 'getUploadsByRoot')
+      .mockResolvedValue([uploadEntry])
+    jest
+      .spyOn(BlockstoreUseCases, 'getFileUploadIdCID')
+      .mockResolvedValue(stringToCid(ROOT_CID))
+    jest
+      .spyOn(ObjectUseCases, 'getMetadata')
+      .mockResolvedValue(err(new ObjectNotFoundError('not found')))
+    const removeNodes = jest
+      .spyOn(nodesRepository, 'removeNodeByRootCid')
+      .mockResolvedValue(undefined as any)
+    const saveNodes = jest
+      .spyOn(nodesRepository, 'saveNodes')
+      .mockResolvedValue(undefined)
+
+    await expect(
+      NodesUseCases.migrateFromBlockstoreToNodesTable(UPLOAD_ID),
+    ).rejects.toBeInstanceOf(UnrecoverableUploadError)
+    // Nothing written, and nothing deleted either: the check happens before the
+    // clear-and-reinsert.
+    expect(removeNodes).not.toHaveBeenCalled()
+    expect(saveNodes).not.toHaveBeenCalled()
   })
 })
--- a/apps/backend/__tests__/unit/core/completionRetryIdempotency.spec.ts
+++ b/apps/backend/__tests__/unit/core/completionRetryIdempotency.spec.ts
@@ -14,6 +14,7 @@ import {
 import { UploadType } from '@auto-drive/models'
 import { ok } from 'neverthrow'
 import { UploadFileProcessingUseCase } from '../../../src/core/uploads/uploadProcessing.js'
+import { UploadPartsChangedError } from '../../../src/core/uploads/errors.js'
 import {
   NodesUseCases,
   createNodeDeduplicator,
@@ -210,6 +211,114 @@ describe('completeUploadProcessing retry idempotency', () => {
     // Exactly one root row, and one root CID — the shape getRootNodeCID accepts.
     expect(store.countOf(MetadataType.File)).toBe(1)
     expect(store.distinctCidsOf(MetadataType.File)).toBe(1)
+  })
+
+  // The limit of that resume. Reusing the derived root is right only while the
+  // upload still holds the bytes it was derived from, and `part A, failed
+  // complete, part B, complete` is legal at every step: uploadChunk has no status
+  // guard, the failed completion released the row to PENDING, and S3 permits
+  // UploadPart after a failed CompleteMultipartUpload. Silently returning A's CID
+  // would hand the client a truncated object to store and reference.
+  it('refuses to reuse a root node once more parts have been stored', async () => {
+    installFakeBlockstore()
+
+    // Part A: one full chunk and a tail, i.e. the ordinary mid-upload state.
+    const partA = Buffer.alloc(DEFAULT_MAX_CHUNK_SIZE + 1000, 1)
+    let storedBytes = BigInt(partA.byteLength)
+    let storedInfo: any = {
+      upload_id: UPLOAD_ID,
+      last_processed_part_index: null,
+      pending_bytes: null,
+      created_at: new Date(),
+      updated_at: new Date(),
+    }
+
+    jest
+      .spyOn(fileProcessingInfoRepository, 'getFileProcessingInfoByUploadId')
+      .mockImplementation(async () => storedInfo)
+    jest
+      .spyOn(fileProcessingInfoRepository, 'updateFileProcessingInfo')
+      .mockImplementation(async (info: any) => {
+        storedInfo = { ...info }
+        return storedInfo
+      })
+    jest
+      .spyOn(filePartsRepository, 'getUploadFilePartsSize')
+      .mockImplementation(async () => storedBytes)
+
+    const upload = {
+      id: UPLOAD_ID,
+      root_upload_id: UPLOAD_ID,
+      type: UploadType.FILE,
+      name: 'grew.bin',
+      upload_options: null,
+    } as any
+    jest.spyOn(uploadsRepository, 'getUploadEntryById').mockResolvedValue(upload)
+
+    await UploadFileProcessingUseCase.processChunk(UPLOAD_ID, partA, 0)
+    const firstCid =
+      await UploadFileProcessingUseCase.completeUploadProcessing(upload)
+
+    // Part B arrives while the upload is PENDING again after the failed
+    // completion, then the client completes a second time.
+    const partB = Buffer.alloc(DEFAULT_MAX_CHUNK_SIZE, 2)
+    await UploadFileProcessingUseCase.processChunk(UPLOAD_ID, partB, 1)
+    storedBytes += BigInt(partB.byteLength)
+
+    await expect(
+      UploadFileProcessingUseCase.completeUploadProcessing(upload),
+    ).rejects.toThrow(UploadPartsChangedError)
+    // Not a 500, and not retryable: the client has to abort and start again.
+    await expect(
+      UploadFileProcessingUseCase.completeUploadProcessing(upload),
+    ).rejects.toMatchObject({ statusCode: 400 })
+
+    // The truncated CID is never handed back.
+    expect(cidToString(firstCid)).toBeDefined()
+  })
+
+  // The resume itself must keep working: same parts, same root, no error. This is
+  // the advertised "top up your credits and retry" flow.
+  it('still reuses the root node when the stored parts are unchanged', async () => {
+    installFakeBlockstore()
+
+    const part = Buffer.alloc(DEFAULT_MAX_CHUNK_SIZE + 500, 7)
+    let storedInfo: any = {
+      upload_id: UPLOAD_ID,
+      last_processed_part_index: null,
+      pending_bytes: null,
+      created_at: new Date(),
+      updated_at: new Date(),
+    }
+    jest
+      .spyOn(fileProcessingInfoRepository, 'getFileProcessingInfoByUploadId')
+      .mockImplementation(async () => storedInfo)
+    jest
+      .spyOn(fileProcessingInfoRepository, 'updateFileProcessingInfo')
+      .mockImplementation(async (info: any) => {
+        storedInfo = { ...info }
+        return storedInfo
+      })
+    jest
+      .spyOn(filePartsRepository, 'getUploadFilePartsSize')
+      .mockResolvedValue(BigInt(part.byteLength))
+
+    const upload = {
+      id: UPLOAD_ID,
+      root_upload_id: UPLOAD_ID,
+      type: UploadType.FILE,
+      name: 'unchanged.bin',
+      upload_options: null,
+    } as any
+    jest.spyOn(uploadsRepository, 'getUploadEntryById').mockResolvedValue(upload)
+
+    await UploadFileProcessingUseCase.processChunk(UPLOAD_ID, part, 0)
+    const firstCid =
+      await UploadFileProcessingUseCase.completeUploadProcessing(upload)
+    const secondCid =
+      await UploadFileProcessingUseCase.completeUploadProcessing(upload)
+
+    expect(cidToString(secondCid)).toBe(cidToString(firstCid))
   })
 
   // Guards the other half: even if some future path does re-put the root node,
--- a/apps/backend/__tests__/unit/core/completionRetryIdempotency.spec.ts
+++ b/apps/backend/__tests__/unit/core/completionRetryIdempotency.spec.ts
@@ -1,0 +1,278 @@
+import {
+  jest,
+  describe,
+  it,
+  expect,
+  beforeEach,
+  afterEach,
+} from '@jest/globals'
+import { MetadataType, cidToString } from '@autonomys/auto-dag-data'
+import { UploadType } from '@auto-drive/models'
+import { UploadFileProcessingUseCase } from '../../../src/core/uploads/uploadProcessing.js'
+import { createNodeDeduplicator } from '../../../src/core/objects/nodes.js'
+import { blockstoreRepository } from '../../../src/infrastructure/repositories/uploads/blockstore.js'
+import { fileProcessingInfoRepository } from '../../../src/infrastructure/repositories/uploads/fileProcessingInfo.js'
+import { filePartsRepository } from '../../../src/infrastructure/repositories/uploads/fileParts.js'
+import { uploadsRepository } from '../../../src/infrastructure/repositories/uploads/uploads.js'
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+const UPLOAD_ID = 'upload-retry-1'
+
+// The root node types are the ones blockstore_root_node_unique_idx covers.
+const ROOT_NODE_TYPES = [MetadataType.File, MetadataType.Folder]
+
+/**
+ * In-memory stand-in for uploads.blockstore that reproduces the SQL semantics
+ * the code depends on: rows ordered by an ascending surrogate key, duplicates
+ * permitted for chunk rows, and ON CONFLICT DO NOTHING against the partial
+ * unique index on the root node types.
+ *
+ * Faking at the repository layer rather than the blockstore layer keeps the real
+ * MultiUploadBlockstore in the test, so the DAG is built by the same code that
+ * builds it in production.
+ */
+const installFakeBlockstore = () => {
+  const rows: Array<{
+    sort_id: number
+    upload_id: string
+    cid: string
+    node_type: MetadataType
+    node_size: bigint
+    data: Buffer
+  }> = []
+  let nextSortId = 1
+
+  jest
+    .spyOn(blockstoreRepository, 'addBlockstoreEntry')
+    .mockImplementation(async (uploadId, cid, nodeType, nodeSize, data) => {
+      const violatesRootUnique =
+        ROOT_NODE_TYPES.includes(nodeType) &&
+        rows.some((r) => r.upload_id === uploadId && r.cid === cid)
+      if (violatesRootUnique) return
+
+      rows.push({
+        sort_id: nextSortId++,
+        upload_id: uploadId,
+        cid,
+        node_type: nodeType,
+        node_size: nodeSize,
+        data,
+      })
+    })
+
+  const byUpload = (uploadId: string) =>
+    rows
+      .filter((r) => r.upload_id === uploadId)
+      .sort((a, b) => a.sort_id - b.sort_id)
+
+  jest
+    .spyOn(blockstoreRepository, 'getByType')
+    .mockImplementation(async (uploadId, nodeType) =>
+      byUpload(uploadId).filter((r) => r.node_type === nodeType).map(
+        (r) => ({ ...r, node_size: r.node_size.toString() }) as any,
+      ),
+    )
+
+  jest
+    .spyOn(blockstoreRepository, 'getByCid')
+    .mockImplementation(
+      async (uploadId, cid) =>
+        (byUpload(uploadId)
+          .filter((r) => r.cid === cid)
+          .map((r) => ({ ...r, node_size: r.node_size.toString() }))
+          .at(0) as any) ?? null,
+    )
+
+  jest
+    .spyOn(blockstoreRepository, 'getByCIDWithoutData')
+    .mockImplementation(
+      async (uploadId, cid) =>
+        (byUpload(uploadId)
+          .filter((r) => r.cid === cid)
+          .map((r) => ({ ...r, node_size: r.node_size.toString() }))
+          .at(0) as any) ?? null,
+    )
+
+  jest
+    .spyOn(blockstoreRepository, 'getBlockstoreEntriesWithoutData')
+    .mockImplementation(
+      async (uploadId) =>
+        byUpload(uploadId).map(
+          (r) => ({ ...r, node_size: r.node_size.toString() }) as any,
+        ),
+    )
+
+  // The DAG builder deletes nodes it has folded into a parent. Mirrors the SQL,
+  // which deletes every row for that (upload_id, cid) — not just one.
+  jest
+    .spyOn(blockstoreRepository, 'deleteBlockstoreEntry')
+    .mockImplementation(async (uploadId, cid) => {
+      for (let i = rows.length - 1; i >= 0; i--) {
+        if (rows[i].upload_id === uploadId && rows[i].cid === cid) {
+          rows.splice(i, 1)
+        }
+      }
+    })
+
+  return {
+    rows,
+    countOf: (nodeType: MetadataType) =>
+      rows.filter((r) => r.node_type === nodeType).length,
+    distinctCidsOf: (nodeType: MetadataType) =>
+      new Set(rows.filter((r) => r.node_type === nodeType).map((r) => r.cid))
+        .size,
+  }
+}
+
+describe('completeUploadProcessing retry idempotency', () => {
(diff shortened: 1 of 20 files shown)


EmilFattakhov:
bugbot run

EmilFattakhov:
Confirmed and fixed in d3d72362.

The guard was read-then-act, so two overlapping calls both observed `PENDING` and both fell through into `completeUploadProcessing`, re-INSERTing the root node and re-charging credits. Exactly the shape this PR set out to remove.

Completion is now claimed with a conditional `UPDATE` (compare-and-swap) in `claimUploadForCompletion`, so exactly one caller proceeds regardless of how many API processes are running; the loser gets an "already being completed" error instead of a corrupted upload.

One non-obvious constraint worth recording: the claim deliberately does **not** flip straight to `MIGRATING`. The migration recovery sweep selects on `MIGRATING`, and a still-completing upload has no root node yet — so claiming with `MIGRATING` would let the sweep enqueue `migrate-upload-nodes` mid-completion, and the new unrecoverable-state handling in this same PR would then park a perfectly healthy upload as `FAILED`. Any upload whose completion outlasts the 30-minute staleness window would hit that. It also has to keep `updated_at` marking *MIGRATING-entry* for the staleness anchor #797 depends on. Hence a distinct `COMPLETING` state, with the flip to `MIGRATING` still happening after finalization exactly where it did before.

Failure paths: the claim is released back to `PENDING` when completion fails (so retry-after-failure, e.g. `Not enough upload credits`, behaves as it does today), and a claim older than `UPLOAD_COMPLETION_CLAIM_STALE_MS` (default 1h) is re-claimable, since nothing sweeps `COMPLETING` and a process dying mid-completion must not strand the upload. `uploads.uploads.status` is plain `text`, so no DB migration.

Covered by three new tests: the claim loser never reaches the processing; the winner claims before processing and only reaches `MIGRATING` after the root node exists; a failed completion releases the claim and never flips the status.

EmilFattakhov:
bugbot run

EmilFattakhov:
Confirmed and fixed in 53787bb9.

The claim-loss branch conflated two different situations: a winner still working, and a winner that already finished between this call's status read and its compare-and-swap. For a small upload that second window is easy to hit, and the retry got a spurious error even though completion had succeeded — the opposite of the idempotency this PR is for.

On claim loss the upload is now re-read:

- `MIGRATING` → same already-completed path as the fast path: return the existing CID and re-drive migration.
- row gone (artifacts already cleaned up by a successful migration) → not-found, matching the check at the top of the function.
- anything else (a genuinely held `COMPLETING` claim, or `PENDING` after the winner released it) → the retry-later error.

The already-completed branch is extracted into `resolveAlreadyCompletedUpload` and called from both entry points, so the two cannot drift apart.

Two new tests cover it: claim lost to a finished winner returns the CID and still re-drives migration; claim lost with the row gone reports not-found. The existing still-running case now explicitly re-reads `COMPLETING` rather than relying on a single stubbed status.

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
I think I ran out of bugbot credits or something 😄 This is ready for the initial review, and I'll follow up on any issues

jim-counter (commented):
Reviewed locally on `fix/upload-complete-idempotency`. The diagnosis in the description is accurate and each of the four changes does what it claims. Three findings below; the first two I confirmed with throwaway repro tests rather than by inspection.

## Verification

Worktree set up fully (submodules, install, models build, files-gateway build), Docker available:

- `tsc --noEmit` clean, `yarn backend lint` clean
- Unit: 37/37 suites, 489 tests pass. The 7 DB-backed suites reported as failing in the description pass here, so that was a sandbox artifact rather than anything in the diff.
- E2E: 9/9 suites, 154 tests pass. Integration `s3-sdk` passes.

## 1. `completeUpload` is still non-idempotent on the retry-after-failure path

`core/uploads/uploads.ts`, the `catch` that releases the claim: the release back to `PENDING` happens after `completeUploadProcessing` has already written the root node. `handleFileUploadFinalization` throws `Not enough upload credits` (`uploadProcessing.ts:181`) strictly after that write. So the documented "top up and retry" flow re-enters `completeFileProcessing` and writes a second root node.

Confirmed by test, with two shapes depending on file size:

- Small single-node file: `rows=2, distinctCids=1`. Tolerated by change (2), but still a wasted write.
- File with a leftover chunk: run 2 re-puts the leftover `FileChunk` (`completeFileProcessing` never advances the cursor), so the DAG gains an extra link and the second root has a **different CID**. My repro printed `rows=2, distinctCids=2` with two visibly different root CIDs.

The divergent-CID case is the bad one. `getRootNodeCID` correctly refuses it, so the retry 500s and every further retry appends another root row. If such a row ever reaches `MIGRATING`, change (3) silently parks the whole tree as `FAILED`, and change (4) then reports it as all-null indefinitely: no dead letter, no user-visible error, no operator signal.

This is pre-existing (main fails the same way with `Invalid number of blockstore entries`), so it is not a regression from this PR. But it is the remaining non-idempotent path in a PR whose headline is idempotency, and it is reachable through the exact flow the description says is preserved unchanged. It also makes the 1h stale-claim takeover more dangerous than the config comment suggests: a completion that outruns the window does not merely duplicate, it can diverge.

Fix direction: clear the derived nodes when releasing the claim, or make `blockstore.put` upsert on `(upload_id, cid)` for derived nodes only. Either way it wants a test. The new suite covers every claim-loss branch but not this one.

## 2. The safety argument for tolerating duplicates does not hold above 100 blockstore rows

The description justifies change (2) with "migration already dedupes downstream ... so no duplicate `nodes` rows result". That dedupe is **per batch**: the `uniqueNodes` Map is built inside the `asyncIterableForEach` callback with `BATCH_SIZE = 100` (`core/objects/nodes.ts:171-180`), and `getAllKeys` yields in `sort_id ASC` (insertion) order. A duplicate appended by a later completion pass lands in a different batch from its original whenever the original sits at position <= 100 and the copy at > 100.

Confirmed: a 100-row blockstore plus one duplicate root row produced **101 `nodes` rows, with the root node present twice**. `getCidsByRootCid` has no `DISTINCT` (`core/objects/nodes.ts:233`), so that fans straight into duplicate on-chain publish work, which is the failure mode the `removeNodeByRootCid` comment directly above it warns about.

This does not affect the six 13-byte production uploads (well under 100 rows), so the incident recovery is unaffected. But it undercuts the general tolerance being added.

## 3. Concurrent complete on the S3 route returns an opaque 500

The new `BadRequestError` on claim loss reaches `completeMultipartUploadHandler` (`app/controllers/s3/s3.ts:687`), which has no try/catch around the use case, so it goes `asyncSafeHandler` -> `next(err)` -> Express default handler -> a non-XML 500. An S3 client's own timeout retry, which is precisely the case this PR defends against, gets an unparseable error instead of a retryable one. Worth mapping to a 503 `SlowDown` or a 409 XML error in that handler. The follow-up note covers the REST-side `handleInternalError` flattening but not this path.

## Smaller notes

- Same condition, two error classes: the terminal-status rejection near the top of `completeUpload` throws a plain `Error`, while the claim-loss equivalent throws `BadRequestError`.
- `resolveAlreadyCompletedUpload` enqueues a migrate task on every repeat complete, and each migrate run does `removeNodeByRootCid` plus a full node re-insert. A client retrying in a loop amplifies that against a large object. A staleness guard mirroring `getStuckRootMigrations` would bound it.
- A stranded `COMPLETING` row cannot be aborted for an hour, since `abortUpload` requires `PENDING`, so S3 `AbortMultipartUpload` returns `NoSuchUpload` in that window. The missing sweeper is flagged in the description, but not this consequence of it.

## Checked and correct

`uploads.uploads.status` is plain `text` with no check constraint, so no migration is genuinely needed. `COMPLETING` has no other consumers (no frontend or Hasura references). The compare-and-swap is sound. `releaseUploadCompletionClaim` cannot undo a `MIGRATING` flip. Keeping `COMPLETING` out of `getStuckRootMigrations` is the right call, and the reasoning given for it is correct.

EmilFattakhov:
@jim-counter thanks — all three findings and all three smaller notes verified and addressed. One of them turned out to go deeper than reported, and it invalidated my first attempt at the fix; details below.

Five commits pushed (`74f04e7f`..`afb77f1b`). The description is unchanged, so the finding-1 analysis in it is superseded by this comment.

## 1. Retry-after-failure non-idempotency — confirmed, and worse than described

Your mechanism is exactly right: `getUnprocessedChunkFromLatestFilePart` returns `pending_bytes`, `completeFileProcessing` never clears it, and the credits failure is thrown strictly after the root node write.

What I found while writing a test for it: `processBufferToIPLDFormatFromChunks` is **destructive**. For a single-chunk file it does `get → delete → put(root)` (auto-dag-data `chunker.ts`), consuming the chunk it folded in. Re-derivation therefore cannot reproduce the first result under any circumstances.

That makes the obvious fix wrong. I first just cleared `pending_bytes` — which removes the accidental re-put that was the *only* reason your small-file case produced `distinctCids=1`. With the tail no longer re-flushed, run 2 finds zero chunks and builds a root over an empty link set, so the benign duplicate becomes your divergent-CID case. The test failed with two different CIDs, which is how I caught it.

`d845c2af` fixes it by deriving **at most once**: if a root node already exists, return it instead of re-deriving. It delegates to `getFileUploadIdCID` rather than reading row `[0]`, so duplicate-but-identical rows stay tolerated and genuinely divergent ones stay fatal. The `pending_bytes` clear stays as the complement, covering the narrower window where run 1 flushed the tail and then failed *before* deriving a root.

Checked that the guard cannot misfire on a first completion: `processChunksToIPLDFormat` only ever writes `FileChunk` nodes, so a `File`-type row exists only after a derivation has run.

Remaining hole, documented rather than papered over: a crash *inside* the chunker, after the delete but before the root write, leaves nothing to rebuild from. That now surfaces as an unrecoverable migration — parked, acked, and counted — instead of silent corruption.

## 2. Per-batch dedupe — confirmed

You're right that the description's safety argument was wrong. `35d3c971` replaces the per-batch `uniqueNodes` Map with `createNodeDeduplicator`, which spans the whole upload. I extracted it as a named helper specifically so the cross-batch property is unit-testable — the inline version isn't, since `getUploadBlockstore` is a bare ESM export with no seam to mock. It also subsumes the per-batch Map, so there's one deduplication mechanism instead of two.

## 3. Opaque 500 on the S3 route — confirmed

Verified the handler has no try/catch, the use case doesn't wrap the throw in a `Result`, and there's no Express error middleware. `afb77f1b` introduces `UploadCompletionInProgressError` (409, transient) so the route doesn't have to guess from a message string, and the S3 handler answers **503 SlowDown** — the AWS SDK retries it with backoff until the winning call finishes. Terminal states stay `BadRequestError`.

## Smaller notes

- **Plain `Error` vs `BadRequestError`** — fixed in `32eafe73`.
- **Re-drive amplification** — agreed, and your cost argument is stronger than my own earlier read of it: I'd called the enqueue cheap without accounting for `removeNodeByRootCid` plus a full node re-insert. Now rate-limited by the recovery sweep's staleness window, so a repeat complete can only pull forward a re-drive the sweep was about to do anyway.
- **Stranded `COMPLETING` un-abortable for an hour** — agreed, and it was a real regression against `PENDING`. `abortUpload` now also accepts a `COMPLETING` row whose claim has gone stale, the same judgement `claimUploadForCompletion` makes when it lets a stale claim be re-taken. A *live* completion stays non-abortable deliberately: it's working, not stranded, and racing `removeUploadArtifacts` against it would pull rows out from under a running DAG build.

Both of the last two need the row's age, so `updated_at` is now part of `UploadEntry` — it was always selected (every read is `SELECT *`), just untyped.

## Also in this push, from a separate review pass

`74f04e7f` closes the duplicate-root-node hole at the database level, which the app-level claim can't: a claim that goes stale under a still-running completion, and old processes during a rolling deploy. Partial unique index on `(upload_id, cid) WHERE node_type IN ('File','Folder')` plus `ON CONFLICT DO NOTHING`.

Scoping to root types is what makes it safe — intermediate nodes of a chunked file are written as `FileInlink`, not `File`, so exactly one root row exists per `upload_id`, and chunk rows (where identical content legitimately repeats) are untouched. The `ON CONFLICT` isn't redundant with the index: with the constraint alone, a completion that died after writing the root but before flipping to `MIGRATING` could never be retried, since every retry would raise a unique violation.

The migration deduplicates existing rows first or the index can't build. That's lossless — `cid` is the content hash — and it keeps the lowest `sort_id`, the row the readers already return. Grouping deliberately includes `node_type`: if the one-root-row-per-upload reasoning is ever wrong, the index build fails loudly rather than silently deleting a non-duplicate.

`32eafe73` also adds `auto_drive_upload_migration_unrecoverable` on the park path. Parking stops the dead-letter loop, but it removes the only signal anything is wrong — a `FAILED` upload keeps its rows, is no longer swept, and reports the same in-progress state as a healthy one. Unbounded-but-visible is how this incident got noticed at all, and trading it for unbounded-and-invisible would set up the next one.

One correction to my own description while I'm here: the `getUploadStatus` rationale overstated the risk. `objectStatus()` maps `totalNodes === 0` to `Processing` before it compares `archivedNodes` to `totalNodes`, and the web app's file lists read Hasura aggregates rather than this endpoint, so nothing rendered a false "archived". The zeroes mislead REST/SDK consumers doing their own arithmetic — reason enough to return nulls, but not the reason I gave.

## Testing

Lint and `tsc --noEmit` clean on both `tsconfig.json` and `tsconfig.test.json`. New `completionRetryIdempotency.spec.ts` (6 tests) backs the blockstore with an in-memory fake at the *repository* layer so the real `MultiUploadBlockstore` and the real chunker run — that's what surfaced the destructive delete. Unit: 31/38 suites, 431 tests pass locally; the 7 DB-backed failures are the sandbox artifact you already identified.

**Please re-run e2e and `s3-sdk` if you get a chance.** Your green run predates these commits, and they change `completeFileProcessing`, `migrateFromBlockstoreToNodesTable`, and the S3 complete handler — I don't have Docker here, so those suites are the main thing still unverified. The migration SQL is also unexecuted against a real database for the same reason.

EmilFattakhov:
Verified the changes, they look fine to me, ready for another round

jim-counter (changes requested):
Re-reviewed after the five follow-up commits. My three original findings are addressed, one of them by a fix that introduces a blocking regression. Details below, plus two new correctness issues and two smaller ones.

Requesting changes on the strength of finding 1 only. Everything else is discussable.

## Status of the original findings

**1. Non-idempotent retry-after-failure: fixed, and fixed well.** Three layers (at-most-once derivation in `completeFileProcessing`, the partial unique index on `(upload_id, cid)` scoped to root node types, `ON CONFLICT DO NOTHING`). I re-ran my original repros against the new head: the small-file case now leaves one root row, the chunked case no longer diverges, and a forced second insert of the same root row is absorbed. Calling out in the commit message that clearing `pending_bytes` alone was wrong, and why, is the right read of it.

**2. Cross-batch deduplication: right diagnosis, right shape, broken implementation.** See finding 1 below.

**3. Opaque 500 on the S3 route: fixed.** `UploadCompletionInProgressError` extending `ConflictError`, with the handler answering 503 `SlowDown`, is what an S3 client can actually act on. The four smaller notes are all addressed too: typed `BadRequestError` for terminal states, the re-drive rate-limited to the sweep's staleness window, `abortUpload` accepting a stale claim, and no re-drive on the just-finished-winner path.

## 1. Blocking: `createNodeDeduplicator` can empty a batch, and `saveNodes([])` is invalid SQL

`pgFormat('INSERT ... VALUES %L', [])` renders `INSERT INTO nodes (...) VALUES ` with nothing after it. The old per-batch Map could never return zero for a non-empty batch. A cross-batch filter can, and does.

The repro needs no forced rows, just an ordinary upload: **a 16 MB zero-filled file fails to migrate** with `error: syntax error at end of input` at `core/objects/nodes.ts:217`. 259 blockstore rows, 3 distinct CIDs. Batch 1 absorbs all 100 identical chunk CIDs, batch 2's 100 rows are all already seen, the filter returns `[]`. The same test passes on fb275692 (5 nodes rows). Any run of roughly 200 or more identical consecutive chunks triggers it, so about 13 MB of repeated content: sparse files, zero-padded disk images, padded archives.

The consequence lands exactly where this PR started. A pg error is not an `UnrecoverableUploadError`, so `processMigration` rethrows, the task burns its three retries, dead-letters to `frontend-errors`, stays `MIGRATING`, and the recovery sweep re-drives it every staleness window. That is the unbounded dead-letter loop, reintroduced, and reachable from a normal upload rather than from a legacy broken row.

One line fixes it: `if (uniqueNodes.length === 0) return` before the `saveNodes` call. Worth a test with a batch that fully deduplicates, since the three `createNodeDeduplicator` unit tests exercise the filter in isolation and never let it empty a batch, which is why this got through.

## 2. Parts appended after a failed completion are now silently dropped

The at-most-once guard returns the previously derived root without checking whether chunks arrived since it was derived. `uploadChunk` has no status guard, a failed completion releases the row back to `PENDING`, and S3 permits `UploadPart` after a failed `CompleteMultipartUpload`, so this sequence is legal at every step.

Verified: upload part A, fail the completion, upload part B, complete again. The call returns 200 with the CID of **A alone**. On fb275692 the same sequence raised `Invalid number of blockstore entries ... distinctCids=2`.

So the fix trades a loud unrecoverable error for a silently truncated object that the client stores and references. Loud was wrong, but silent-and-wrong is worse for a content-addressed store. Either reject a completion whose chunk set has grown past what the existing root covers, or refuse `uploadChunk` once a root node exists for the upload.

## 3. The two new staleness checks do timestamp math in JS against a `timestamp without time zone` column

`resolveAlreadyCompletedUpload` and the `abortUpload` stale-claim exception both compute `Date.now() - new Date(upload.updated_at).getTime()`. `uploads.uploads.updated_at` is `timestamp` with no zone, and node-postgres parses that in the process-local zone. Confirmed against the installed `pg-types`: one DB string parses to `12:00:00Z`, `11:00:00Z` or `19:00:00Z` under `TZ=UTC`, `Europe/London` and `America/Los_Angeles` respectively.

Under BST every age is inflated by an hour, so a claim taken seconds ago reads as stale and `abortUpload` will tear down a live completion, which is precisely what the new comment says must not happen. Under a negative offset nothing is ever stale, so a stranded `COMPLETING` row stays un-abortable and the re-drive never fires. Nothing pins `TZ` in the compose files or the Dockerfiles, so this rests on an unstated assumption that the process and the database session agree.

Every existing comparison of this kind (`getStuckRootMigrations`, `claimUploadForCompletion`) is done in SQL and is immune. These two are the first to leave it. Doing them in SQL as well, mirroring `claimUploadForCompletion`, keeps the property.

## 4. `yarn backend lint` fails

`no-duplicate-imports` at `__tests__/unit/core/uploadCompletionIdempotency.spec.ts:26`: `errors.js` is imported on both line 12 and line 26.

## 5. The migration builds the index without `CONCURRENTLY`

`CREATE UNIQUE INDEX blockstore_root_node_unique_idx` blocks writes to `uploads.blockstore` for the duration of the build, and that table holds the node payloads. Worth checking the production row count before deploying, or building the index concurrently outside the transaction.

## Test results on the new head

`tsc --noEmit` clean. Unit 38/38 suites, 499 tests. E2E plus integration 10/10 suites, 230 tests. Lint fails with the single error above.

None of the existing suites cover migrating a file with repeated content, so finding 1 is invisible to CI as it stands.

EmilFattakhov:
Thanks for your careful review @jim-counter, I addressed your feedback point-by-point, ready for another review

jim-counter (approved):
LGTM! Thanks for the fixes.

Issue #748, opened 2026-06-15
Payment manager: chain-parameterized instance watching the Ethereum USDC receiver


EmilFattakhov:
## Context
`apps/backend/src/infrastructure/services/paymentManager/index.ts` currently hard-binds one
viem client + one contract (`config.paymentManager.*`) and watches `IntentPaymentReceived`.
The watcher/poller/orphan-recovery design ports directly to a second chain.

> **Updated 2026-08-07** — unaffected by the oracle change (#746), but the kill switch (#811)
> adds one hard rule, and the balance poller needs a home. Both are below.

## Tasks
- [ ] Parameterize the payment manager by chain config (RPC url, contract address, event ABI,
      confirmations) instead of reading the single global `config.paymentManager`. Support
      running both the Auto EVM (AI3) and Ethereum (USDC) instances.
- [ ] For the USDC instance, parse `IntentTokenPaymentReceived(intentId, token, amount, payer)`
      and call `markIntentAsConfirmed({ intentId, paymentAmount: amount, fromAddress: payer })`.
      Note: `fromAddress` comes from the **event's `payer`** field, NOT `receipt.from`, because
      ERC20 payments can be relayed.
- [ ] Persist the USDC `token_amount` so the pricing conversion (#747) can run.
- [ ] Port watch + poll + orphan-recovery-on-restart to the new instance.
- [ ] Confirmation count: Ethereum post-Merge finality — 2–3 blocks is fine, or keep 6 (config).
- [ ] Wire both instances into the server startup.
- [ ] **The USDC instance keeps running while USDC payments are switched off** (#811). The
      gates block intent *creation*; a payment already on-chain must always be observed,
      confirmed and credited. A pause that strands a paid intent is a refund ticket, and
      possibly a support incident, for every user in flight when it fires.
- [ ] Host the treasury balance poller (#811) alongside this instance — it needs the same
      Ethereum client, and it must run in exactly one process, not once per API worker.

## Files
- `apps/backend/src/infrastructure/services/paymentManager/index.ts`
- `apps/backend/src/infrastructure/services/paymentManager/event-abi.ts`

## Acceptance criteria
- Both managers run side by side; AI3 flow unchanged.
- A USDC payment on Sepolia is detected via event + poll, and survives a backend restart
  (orphan recovery) before confirmation.
- Payments confirm and credit normally with every payment gate closed.

PR #807, opened 2026-08-03
feat(priceOracle): source AI3/USD from a subgraph VWAP of recent pool swaps (USDC epic — step 4 rework)

EmilFattakhov:
Sources the AI3/USD rate for the USDC payment path (#746) from the **volume-weighted average
of the WAI3/USDC pool's most recent realized swaps**, read from its subgraph through The Graph's
gateway.

Nothing in `src/` imports the oracle yet — the only consumers are its specs — so this is all
pre-production.

> **Rewritten 2026-08-07/10.** This PR previously priced from live pool state: a `slot0`
> marginal price, a v4 Quoter call per intent, and a 24h TWAP rebuilt from `Swap` logs. The
> epic pivoted (#742): USDC is no longer swapped per intent but **accumulated and converted
> manually**, so "what would this exact size cost right now" answers a question nobody asks.
> The original description is kept at the bottom. Review comments on the old design are
> resolved by deletion rather than by change — sorry for the churn.

## What changed

What a purchase should be priced against is what the pool has actually been **filling** at —
and a realized fill already carries the swap fee and its own price impact, which is precisely
what the executable quote existed to recover. So:

```
usdPerAi3 = Σ |usdc_i| / Σ |ai3_i|      over the last N swaps (default 10, floor 5)
```

Summing before dividing *is* the volume weighting — a swap's influence is its own size — and it
keeps the whole computation to one division. Weighting by size rather than by count is the
manipulation-resistant choice available here: a count-weighted window is filled with dust wash
trades, while moving a volume-weighted one costs real volume against the pool's own fee.

**No on-chain read remains in the oracle.** `ETH_CHAIN_ENDPOINT` stays for the payment manager
(#748) and the treasury balance check (#811), and **no longer needs to be archive-capable** —
that is the expensive tier, and the TWAP reconstruction was the only thing that needed it.

**One error type.** `OracleUnavailableError` carries a machine-readable `reason` — `gateway`,
`misconfigured`, `insufficient-samples`, `stale-window`, `market-moved`, `narrow-window`,
`thin-volume`, `out-of-bounds`, `indexer-lag`, `indexer-error` — so #747 can map status codes and
#811 can render a status without parsing messages. `QuoteTooLargeError`, `PoolEmptyError`,
`InvalidQuoteAmountError` and `PriceDeviationError` go with the quoter: with no per-intent swap
there is no depth to exceed, and with no live spot read there is nothing for a deviation gate to
compare.

Two of those ten are worth naming here, because both came out of review rather than design.
`market-moved` is described below. `misconfigured` separates the failures a **redeploy** fixes —
no `GRAPH_API_KEY`, a subgraph without our pool, currencies the other way round — from the ones
**waiting** fixes. Folding them into `gateway`, as an earlier commit did, pages whoever is on
call to The Graph's status page to debug a stale constant of ours. It stays narrow on purpose: a
5xx is still an outage, and that is pinned as a test.

**Guards fail closed, cheapest and most structural first:** misconfiguration → indexing errors →
indexer lag → window age filter → sample floor → freshness of the *newest* fill → outlier trim
against the median → the newest fill's veto → minimum window span → surviving volume → absolute
bounds.

Two of those bounds are time, and they answer different questions. `ORACLE_MAX_WINDOW_AGE_MS`
(7d) is the window's **lower** bound and works as a filter: a fill older than this may not vote.
`ORACLE_MAX_SWAP_AGE_MS` (24h) then asks whether the market is alive *now*, judged on the newest
fill — because a window's span is a property of the market rather than of the setting, and at
this pool's ~1.6 swaps/day ten fills is roughly a week.

`getHealth()` exposes the window behind the served rate *and* the current failure — an operator
asking "why is the USDC path shut" needs both facts, and a failure must not evict a usable price.

## Verification against the live gateway

The subgraph is a v4 subgraph indexing the mainnet PoolManager, and it has our pool:
`token0=WAI3` (18dp) / `token1=USDC` (6dp), the same ordering `POOL_KEY` pins. The adapter
re-checks that ordering and those decimals on every read before mapping a single amount — a
reversed pool would invert every price plausibly and undetectably.

**The oracle refuses, correctly.** The newest fill is 2026-08-05 08:12:47 UTC — 130 hours old at
the time of writing, against a 24h bound — so `getPrice()` returns `reason=stale-window`. Pool
liquidity is still `0`, withdrawn 2026-08-06. Two further guards would stop this window
independently: the 25% trim removes three swaps and the survivors total **919.53 USDC** against a
1000 USDC floor, and those three are the three *newest* fills, which is now its own refusal
(below). Every figure here is a property of a dead pool — **re-derive every threshold once it
trades again**.

The window is also ten consecutive **sells** — WAI3 in, USDC out, every one — which is what a
pool being drained looks like rather than a market with two sides. Nothing in the design detects
one-directional exhaustion; staleness and the volume floor are what caught this one. Worth
knowing before the thresholds are re-derived.

## Manipulation controls

The trim measures against a **count-based median**, so whoever supplies most of a ten-swap
window sets the price. Volume alone does not stop that — an attacker who clears the 1000 USDC
floor with their own fills clears it at whatever price they like, which on this 1% pool cost
about $20 in fees for a 64x discount. `ORACLE_MIN_WINDOW_SPAN_MS` (2h) closes it: the surviving
fills must have spanned real time. A burst in one block is printable on demand; the same fills
held across hours must be defended against everyone else trading in between, and that delay is
what the treasury alerting (#811) needs in order to react.

**The limit that remains, pinned as a test:** an attacker willing to sustain a wrong price
across hours can still *deny* the payment path, because trimming their fills leaves too few
samples. Denial, not mispricing — the direction this must fail in. If denial ever becomes the
concern, the answer is a deeper pool or a second source, not a looser trim.

## When the market moves, the trim throws away the news

The count-based median cuts both ways, and the second way was a real defect. It drops the lone
absurd print it was built for — but when the market genuinely **re-prices**, the fills carrying
the new price are the minority, so those are the ones it discards. What survives is a majority
that has not caught up, and the average it produces is calmly wrong in the direction of the old
regime with every other guard satisfied: enough samples, enough volume, spanning hours, in bounds.

This pool's own history demonstrates it. Take the ten real fills above, make them fresh and put
their volume over the floor — change nothing about the prices — and the oracle serves
**0.004698 USD/AI3 while the pool's most recent fill was 0.002393**, a rate 96% above anything
that had traded, backed by 7 samples and 1,839 USDC spanning 6h. At ~1.6 swaps/day the window
needs *days* to catch up, and every purchase in between is charged at a price nobody traded.
Which side loses depends on the direction: a crash over-charges the buyer, a rally hands the
margin back as free storage and is straightforwardly arbitrageable — buy credits right after AI3
moves up.

So **the newest fill in the window now has a veto**. If the trim discarded it, the window
describes a regime the market has left, and the oracle refuses with `market-moved` rather than
quoting. Denial instead of mispricing, the same direction every other guard here fails in, and it
costs nothing on a merely volatile market because then the newest fill sits inside the band —
pinned by tests in both directions. `ORACLE_MAX_SWAP_DEVIATION` now sets two things: which fills
are dropped, and how far the market may move before the path shuts. Widening it makes refusals
rarer and the stale-price window wider, so it wants deriving from how far this pool actually moves
between consecutive fills.

The regression test is built from the ten real fills rather than invented ones, so it keeps
failing for the reason it was written.

## Tests

93 unit tests, none of which touch the network or depend on what is in a developer's `.env`.
Both properties are enforced rather than assumed: an unstubbed `fetch` fails the test — in both
oracle suites, not just the adapter's — and the adapter is exercised through
`fetchRecentSwapsFrom`, which takes its endpoint as an argument. Earlier drafts got this wrong
twice in the same way — `config` snapshots env at import, so one draft quietly queried the live
gateway and another passed locally while failing every case in CI, which is what drove the seam.
Coverage: the VWAP/median/trim/volume math; every guard (gateway failure, misconfiguration,
indexing errors, indexer lag, ancient-majority windows, sample floor, staleness, a re-priced
market, burst windows, trim-below-floor, thin volume, out-of-bounds); TTL cache, last-good
fallback, throttling, concurrent-refresh collapsing, health reporting across a failure *and* a
recovery; the adapter's mapping, sign handling, zero-leg dropping and pool-identity assertion; and
`keccak256(abi.encode(POOL_KEY)) === POOL_ID`, which is the only thing pinning currency order.

`tsc --noEmit` and ESLint are clean. The repository and e2e suites need Docker (TestContainers)
and were not run locally; CI covers them.

## Review rounds

**First round.** An independent review of the first version found two real defects on the money
path — the missing lower time bound, and the burst-of-fills case above — plus a nullable
`_meta.block.timestamp` that would have pinned the oracle at a permanent, misattributed
`indexer-lag`, and `subgraphError` defaulting to `deny`, which made the `indexer-error` reason
unreachable. All fixed in the second commit; each claim was checked against the live gateway
rather than argued, and one finding (a supposed GraphQL variable-type violation) was wrong.

**Second round**, against the rewritten version, and re-verified against the live gateway rather
than argued — including re-running the adapter's exact query, which confirmed the first round's
variable-type finding really was a false positive. One defect on the money path, the
`market-moved` case above, found by replaying this pool's real fills through the real guards. Plus
five smaller things, all fixed:

- **Misconfiguration read as an outage** — a stale `POOL_ID`, reversed currencies or a missing
  `GRAPH_API_KEY` all surfaced as `reason=gateway`. Same class of misdiagnosis as the
  `indexer-error` bug the first round caught, one layer down. Now `misconfigured`, via a typed
  `SubgraphConfigError`.
- **`ORACLE_MAX_SWAP_AGE_MS` vs `ORACLE_MAX_WINDOW_AGE_MS` was unchecked.** Two other config
  relationships are validated at import; this one was not, and inverting it *weakens* the oracle
  silently instead of refusing — the window filter runs first, so a wider freshness bound can
  never fire and a market that stopped six days ago prices as current. Now validated too.
- **10^30 was defined twice.** `pricing.ts` exported `USDC_CONVERSION_FACTOR` with a comment
  saying the oracle used it; `quote.ts` re-derived it from hardcoded `18`/`6` literals instead.
  Two definitions of one factor on the path that decides what a user is charged. It now imports.
- **A dashboard could render a failure time with no reason.** A success cleared
  `lastFailureReason` but not `lastFailureAt`. They are now an always-paired historical record,
  with a separate `currentFailureReason` answering "is it degraded right now" — the two facts #811
  asks for, in two fields instead of one overloaded one. `OracleHealth`'s shape is unchanged.
- **Two misleading messages:** the index-lag warning compared seconds against a limit printed in
  milliseconds, and the sample-count refusal blamed the 7d window even when nothing had been
  filtered and the pool simply had too few fills.

Also reverted: prettier reformatting of six untouched `config.ts` blocks. `eslint-config-prettier`
only *disables* rules, so formatting is not enforced here and the churn bought nothing but review
surface.

## Unrelated, included to unblock CI

`__tests__/e2e/uploads/folder.spec.ts` fails intermittently on `main`: its `.map()` has a block
body with no `return`, so `Promise.all` awaits an array of `undefined` and the assertion races
the writes it was meant to wait for (CI saw `uploadedNodes: 1` of 3). Fixed by returning the
promises. Not a sleep — a delay would leave the race in place and hide the fact that nothing was
being awaited.

<details>
<summary>Superseded description (live pool state, v4 Quoter + TWAP gate)</summary>

Reworks the AI3/USD oracle from step 4 (#746) to price from a DEX pool instead of a CEX aggregate, and hardens it for the binding-charge path that step 5 (#747) will build on.

Nothing in `src/` imports the oracle yet — the only consumers are its specs — so this is all pre-production.

## What changed

The oracle now reads the **Uniswap v4 WAI3/USDC pool on Ethereum mainnet** over our own RPC. WAI3 is the bridged representation of AI3, and the pool is USDC-quoted, so the pool price *is* the USD price — no ETH/USD second hop. The previous HTTP source adapter is deleted; there is no secondary feed or fallback source.

The pool is hardcoded ([`POOL_KEY` / `POOL_ID`](https://github.com/autonomys/auto-drive/blob/feat/price-oracle-dex-source/apps/backend/src/infrastructure/services/priceOracle/uniswapV4.ts)), not configurable — a wrong RPC gives a failed connection, but a wrong pool would give a plausible-looking wrong price on a money path. A unit test asserts `keccak256(abi.encode(POOL_KEY)) === POOL_ID` so the two forms cannot drift.

Two reads, both against the **same block**, because they answer different questions:

| Read | Question | Used for |
|---|---|---|
| `StateView.getSlot0` | what's the price? | `getPrice()` — display, and the rate persisted as `usd_rate_at_creation` |
| `V4Quoter.quoteExactOutputSingle` | what would converting *this amount* cost? | `getExecutableQuote()` — the amount actually charged |

**Slippage is priced in, not refused.** `usdcAmount` carries it, and `priceImpactBps` reports how much of the cost is the size's own doing so callers can display it, alert on it, and reconcile against it. There is no slippage threshold in the oracle at all: how much a user may buy is business policy, enforced against the per-user credit cap before an intent exists. The oracle prices whatever size it is handed, and fails only when the pool genuinely cannot fill it.

The pool has no oracle hook, so v4 exposes no TWAP for it. A **spot-deviation gate** stands in: the current price is compared against the median of sampled historical blocks, and a quote is refused when they diverge. `getExecutableQuote` never serves a cached or stale price — it backs a binding charge.

## Verification worth noting

The `PoolKey` components were derived by brute-forcing the hash against the live poolId, not read off a page — which is also how `hooks = 0x0` was confirmed, the fact that rules out a TWAP and motivates the sampling gate.

The pool carries a governance-set **0.1% protocol fee on top of its 1% LP fee**, which v4 composes to 10,990 pips. Assuming the static `POOL_KEY.fee` left ~10bps of phantom slippage in every measurement (a 100 AI3 quote measured 11bps where it should measure 1bps), so the fee is read from the same `slot0` as the price.

Measured against the live pool (spot $0.005361/AI3):

| Purchase | Cost | Slippage | Total premium over spot |
|---|---|---|---|
| 1,000 AI3 | $5.43 | 13 bps | 124 bps |
| 10,000 AI3 | $54.93 | 133 bps | 246 bps |
| 50,000 AI3 | $290.15 | 705 bps | 824 bps |
| 200,000 AI3 | $1,472.17 | 3,579 bps | 3,730 bps |

Every quote carries ~112 bps minimum from the swap fee alone, before `USD_QUOTE_MARGIN`.

## Config

All new keys are optional and default off — existing deploys boot unchanged. `ETH_CHAIN_ENDPOINT` is only required when quoting in USDC; without it the oracle is never invoked, and with the feature enabled but the endpoint missing it fails fast naming the variable. Kept separate from `paymentManager.url`, which points at Auto EVM.

New: `ORACLE_SPOT_SAMPLES`, `ORACLE_SPOT_SAMPLE_SPACING`, `ORACLE_MAX_SPOT_DEVIATION`. Sampling depth defaults to 100 blocks so it fits inside a standard full node's 128-block state window; raising either knob needs an archive-capable RPC, since the gate fails closed.

## Follow-ups, deliberately not in this PR

- **Slippage at cap-sized purchases needs a product answer.** The quote will be accurate, but on this pool a large purchase carries double-digit-percent slippage. Someone has to decide whether the buyer pays it or the treasury absorbs it, and `USD_QUOTE_MARGIN=5` is currently sized as though neither.
- **Manipulation exposure is contained, not closed.** The deviation gate raises the cost of a single-block price push; it does not eliminate it at current pool depth. `ORACLE_MAX_SPOT_DEVIATION=10` is a guess made without data on this pool's 20-minute volatility.
- **#747 needs a `quoted_bytes` column on `intents`.** Deriving USDC credits by inverting `usd_rate_at_creation` would over-credit by exactly fee + slippage + margin, since the user pays the *executable* amount while the stored rate is the *marginal* one. Credits should come from a stored byte count.
- **The per-user credit cap needs gating at intent creation for USDC.** The existing `OVER_CAP` path fires after payment, which for an irreversible transfer means a manual refund.
- **No telemetry on slippage yet.** Worth wiring `priceImpactBps` to `vmetrics` before launch.
- `ORACLE_MIN/MAX_USD_PER_AI3` are still the wide `[0.0001, 100]` defaults.

Part of #742. Supersedes the source used in #780.

</details>


EmilFattakhov:
bugbot run

apps/backend/src/config.ts
```diff
@@ -197,18 +205,42 @@ export const config = {
...
     maxStaleMs: positiveIntEnv('ORACLE_MAX_STALE_MS', 600000),
-    // CoinGecko HTTP timeout.
+    // RPC timeout for both the pool read and the quoter simulation.
     fetchTimeoutMs: positiveIntEnv('ORACLE_FETCH_TIMEOUT_MS', 5000),
-    // Drop the quote if CoinGecko's own reported update time is older than this.
-    // Default: 5 minutes.
+    // Drop the quote if the block the pool state came from is older than this,
+    // which catches a lagging RPC node. Default: 5 minutes.
     maxSourceAgeMs: positiveIntEnv('ORACLE_MAX_SOURCE_AGE_MS', 300000),
+    // Manipulation gate. The pool has no oracle hook, so every read is
```
jim-counter: My feeling is the manipulation gate should cover something like the last 24 hours while we're in this early stage.

apps/backend/.env.sample
```diff
@@ -121,6 +125,22 @@ ORACLE_MAX_SOURCE_AGE_MS=300000
 ORACLE_MIN_USD_PER_AI3=0.0001
 ORACLE_MAX_USD_PER_AI3=100
 
+# Manipulation gate. The pool exposes no TWAP, so the current price is compared
+# against the median of ORACLE_SPOT_SAMPLES readings taken every
+# ORACLE_SPOT_SAMPLE_SPACING blocks, and a quote is refused when it diverges by
+# more than ORACLE_MAX_SPOT_DEVIATION percent.
+#
+# Sampling starts one spacing BEHIND the block being judged, so it reaches
+# (samples * spacing) blocks back — 100 blocks (~20 min at 12s) by default. A
```
jim-counter: The manipulation check window is too short. I think we should make it a requirement we have access to an archive node.

apps/backend/src/infrastructure/services/priceOracle/index.ts
```diff
@@ -188,16 +231,296 @@ const getPrice = async (): Promise<
...
+  const sorted = [...values].sort((a, b) => (a < b ? -1 : a > b ? 1 : 0))
+  const mid = Math.floor(sorted.length / 2)
+  return sorted.length % 2 === 1
+    ? sorted[mid]
+    : (sorted[mid - 1] + sorted[mid]) / 2n
+}
+
+const absDeviationBps = (value: bigint, reference: bigint): bigint => {
+  const delta = value > reference ? value - reference : reference - value
+  return (delta * BASIS_POINTS) / reference
```
jim-counter: `reference` is unguarded here, and a zero reference throws `RangeError: Division by zero` straight out of `getExecutableQuote`, breaking the `Result` contract on a path an attacker can influence.

It is reachable: `sqrtPriceX96ToUsdPerAi3` truncates to `0` for any `sqrtPriceX96` below ~7.9e13, which is far above v4's `MIN_SQRT_PRICE` of 4295128739. So a pool pushed toward the bottom of its range reports a zero *price*, while `sampleUsdPerAi3`'s filter only rejects a zero *sqrtPrice*. The median of those is 0. Confirmed by driving the module with a `[0n, 0n, 0n]` sample set.

Suggest treating a non-positive reference as `OracleUnavailableError` rather than dividing. A zero baseline means the sampled history is unusable, not that the deviation is infinite.

apps/backend/src/infrastructure/services/priceOracle/uniswapV4.ts
```diff
@@ -0,0 +1,411 @@
...
+        'price from the Uniswap v4 pool',
+    )
+  }
+  // viem's actions take no per-call AbortSignal, so the request bound is set on
+  // the transport. The oracle additionally races its own `withTimeout` around
+  // these calls; this is what actually aborts the in-flight HTTP request rather
+  // than leaving it to run out its own default.
+  client = createPublicClient({
+    chain: mainnet,
+    transport: http(url, { timeout: config.priceOracle.fetchTimeoutMs }),
```
jim-counter: `fetchTimeoutMs` is doing two incompatible jobs. It is the per-request transport timeout here, and it is also the budget for the whole multi-call sequence at `index.ts:396` (three round trips: `getBlock`, `getSlot0`, `simulateContract`) and `index.ts:301` (up to five more). Both default to 5000ms, so the outer race always fires first: the transport timeout is effectively dead code, and a healthy-but-slow RPC at ~2s per call fails every quote. Worth splitting into a per-request bound and a total budget, with the total a multiple of the per-request one. This gets more pressing if the sampling window widens.

Separately, the comment above overstates what the outer race does. `fetchUniswapV4Quote` takes no parameter, so the `AbortSignal` that `fetchQuote` passes it is silently discarded, and the two quote paths never construct a controller at all. Nothing aborts an in-flight viem request anywhere, so a timed-out call runs to completion in the background. The comment at `index.ts:108-111` has this right and this one contradicts it.

apps/backend/src/infrastructure/services/priceOracle/index.ts
```diff
@@ -159,11 +199,14 @@ const refresh = async (): Promise<
...
  *
- * Serves the cached value while fresh; otherwise refreshes from CoinGecko, with
- * concurrent callers sharing one in-flight request. When a recent fetch failed,
+ * Serves the cached value while fresh; otherwise re-reads the pool, with
+ * concurrent callers sharing one in-flight request. When a recent read failed,
  * subsequent calls within `cacheTtlMs` serve the last-good value (or error)
  * without re-hitting upstream, so a degraded source is not hammered. Returns
  * `err(OracleUnavailableError)` when no trustworthy price is available.
+ *
+ * This is the pool's *marginal* price. For anything that will actually be
```
jim-counter: Worth saying explicitly that `getPrice` is not only the marginal price, it is also the *ungated* one. `checkSpotDeviation` runs only inside `getExecutableQuote`, so this path's sole guard is the `[0.0001, 100]` bounds check. Since the PR description has this feeding display and the persisted `usd_rate_at_creation`, a manipulated read reaches the database even in the case where `getExecutableQuote` would have refused it.

It also gets written to both `cache` (60s) and `lastGood`, and `lastGood` is then served for up to `ORACLE_MAX_STALE_MS` of 10 minutes, so a manipulated value outlives the manipulation by ten minutes.

Two asks: have #747 persist the gated `ExecutableQuote.usdPerAi3` rather than this, and consider not populating `lastGood` from a read that has not been through the gate.

jim-counter (commented):
The implementation quality here is high, and the findings below are structural rather than defects in the code. The parts that are easy to get wrong hold up: the protocol fee composition matches v4's `ProtocolFeeLibrary.calculateSwapFee`, the `>> 12` correctly selects the oneForZero half, `removePoolFee` correctly inverts an input-side fee, and the poolId hash test is load-bearing rather than decorative. Typecheck is clean and all 69 oracle and pricing tests pass.

Two things need resolving before #747 puts a binding charge on this. Smaller items are inline.

## 1. The manipulation window is too short, and shorter than it looks

`sampleUsdPerAi3` reads `getSlot0` at five block heights and takes their median. Moving a median of five needs three samples, which means the three newest, so the push has to be in place at `H-60`. **Holding a push for ~60 blocks defeats the gate entirely, at any magnitude the absolute bounds permit.**

Three reasons this is weaker than "N blocks of defence" suggests:

- **No trades are sampled.** `slot0.sqrtPriceX96` changes only on a swap, so on a pool that trades infrequently every sample collapses to the price left by the last swap. The median becomes one observation counted five times, and the outlier resistance it was chosen for does not exist.
- **Holding is free.** Because `slot0` is sticky between swaps, a push persists on every subsequent block until somebody else trades. In the situation this is the only venue AI3 trades on, no arbitrageur pulls it back, so waiting carries neither cost nor risk. The attacker's only unavoidable expense is the round-trip swap fee.
- **It bounds rate, not level.** The gate is purely relative, so the ratchet above works: each step is legal and the median catches up between steps. The only absolute anchor is `ORACLE_MIN/MAX_USD_PER_AI3`, and at this asset's volatility no band narrow enough to constrain an attacker is wide enough to avoid tripping on real movement. Those bounds are an integration-error filter, not a manipulation control.

Two suggestions for the design discussion:

- **Read swap events rather than sampling state.** The `PoolManager` `Swap` event carries the post-swap `sqrtPriceX96` indexed by pool id, so a couple of `eth_getLogs` calls give the real trade history with exact prices, and time-weighting them yields an exact TWAP rather than a sampled approximation. `eth_getLogs` is served off the log index rather than archival state, so this may reduce the RPC requirement rather than expand it, subject to your provider's range limits. Prefer time-weighting to a trade count: "last N trades" is attacker-controlled, since dust wash trades can fill the window.
- **Consider the average as a floor on the rate rather than a gate.** The only profitable direction is pushing the rate down. If the charge can never price AI3 below the trailing average, a downward push earns nothing, no refusal is needed, and the fail-closed outage on the revenue path disappears. Upward pushes harm users rather than the treasury and can keep a refuse path on a shorter window. One complication: `usdcAmount` comes from the quoter rather than from `rate × size`, so a floor cannot be a simple `max()` without a model for the expected fee-and-impact component at that size.

## 2. Concurrent intents under-collect, with no manipulation involved

Slippage is convex: buying a fraction φ of the reserve carries a premium of φ/(1-φ). Each intent is priced against instantaneous pool state, so intents quoted before any of them convert are each priced as though they were the only one. Ten intents each taking 1/40th of the reserve are quoted at ~2.6% premium apiece, while acquiring the combined 25% costs ~33%: a ~30 point shortfall on the notional against a 5% margin. No manipulation is involved, and the per-user credit cap cannot see it, being denominated in bytes with no knowledge of pool depth.

Conversion is manual and batched, so there is no depleted state for a later quote to see and serialising intents does not help. Carrying this as a known risk for now, bounded by:

- **A configurable threshold on outstanding un-converted AI3**, alerting well below the level at which it restricts new quotes. Denominated in AI3 rather than USD, since the gap tracks volume against depth. Checked at intent creation rather than after payment, for the reason follow-up 4 gives, and counting pending intents still inside their price-lock window as well as confirmed ones. Needs somewhere to record that a conversion has happened, since `COMPLETED` means credits granted rather than AI3 acquired.
- **Sizing `USD_QUOTE_MARGIN` for the expected batch**, rather than leaving it at a figure the description notes was sized as though neither party absorbs slippage.
- **Per-batch reconciliation** of actual conversion cost against the sum of quoted `usdcAmount`s. `ExecutableQuote` already carries the fields needed for it.

Worth noting in the code that `usdcAmount` is a floor on conversion cost rather than an estimate: a batch operator pays the batch cost, not the sum of the individual costs. This bounds the convexity gap only, not manipulation exposure.

## One process note

Every slippage figure is a function of pool depth at the moment it was measured, and depth moves independently of price when liquidity is added or withdrawn. Any threshold tuned against the numbers in the description needs re-deriving against live state at the time it is set, and probably wants to be expressed relative to reserves rather than as an absolute.

apps/backend/src/infrastructure/services/priceOracle/index.ts
EmilFattakhov: Confirmed and fixed. A non-positive reference is now treated as an unusable baseline (`OracleUnavailableError`), not as an infinite deviation — the division is unreachable with a zero denominator, and the reference is never cached when non-positive.

Your reasoning about how it is reached carried over to the new implementation unchanged, so the comment explaining the guard cites it: the conversion truncates to `0` well above `MIN_SQRT_PRICE`, so a zero *price* survives a filter that only rejects a zero *sqrtPrice*.

apps/backend/src/infrastructure/services/priceOracle/uniswapV4.ts
EmilFattakhov: Both confirmed and fixed.

Split into `ORACLE_RPC_TIMEOUT_MS` (5s, the transport bound on one request) and `ORACLE_FETCH_TIMEOUT_MS` (30s, the budget for a whole multi-call sequence), with a load-time check that the total exceeds the per-request bound so the degenerate configuration you describe cannot be set silently. This mattered more than it did before the rework: building the average is one state read plus one log query per chunk, so the sequence is now longer, not shorter.

The comment was wrong and is rewritten. It now says plainly that the transport timeout is the only thing that stops an in-flight request, that the `withTimeout` races bound how long a caller waits rather than aborting anything, and that a request given up on runs to completion in the background.

apps/backend/src/infrastructure/services/priceOracle/index.ts
EmilFattakhov: Confirmed, and fixed structurally rather than by convention: `refresh()` now runs the same gate before anything is written to `cache` or `lastGood`. A read that fails it is not remembered at all, and the caller gets the previous good value (or an error). There is a test that asserts the pushed value does not become the new last-good.

On the first ask — the docstring now says outright that `getExecutableQuote` is where the rate to persist comes from, and why: deriving `usd_rate_at_creation` from here stores a number the pool was never asked to honour at the intent's size. #747 should take `ExecutableQuote.usdPerAi3`.

Worth flagging that gating this path is not free. The gate fails closed, so during a move large enough to trip it the displayed price freezes at the last good value and then errors after `ORACLE_MAX_STALE_MS`. Given what this pool did over the last 45 days, that will happen occasionally. I think that is the right trade for a value that reaches the database, but it is a real availability cost rather than a free win.

apps/backend/src/config.ts
EmilFattakhov: Done — the window is now 24h (`ORACLE_TWAP_WINDOW_BLOCKS=7200`), and it is a genuine time-weighted average over swap events rather than a median of sampled state. Details in the summary comment; the short version is that the reference is built from the `Swap` log history plus one seed read, so a push has to be held for a large fraction of a day to move it.

One consequence worth knowing about before this ships: with a window this wide, the threshold has to absorb a full day of genuine drift. Measured over 45 days of this pool's real history, the 10% in this file would have refused every quote for 21% of all hours. The defaults are now 25% down / 50% up, derived from that history rather than picked.

apps/backend/.env.sample
EmilFattakhov: Agreed, and documented as a requirement here rather than as a caveat: the sample now says the endpoint must be archive-capable, because the gate reads pool state one full window back and queries logs across the same span, and it fails closed.

It needs less archival *state* than the sampling version implied, though. The history now comes from `eth_getLogs` against the `Swap` topic, which is served off the log index; the only archival state read is the single seed at the window's start. Verified against two public endpoints that a 7200-block filtered log query is served fine, while a 21600-block one is rejected — hence `ORACLE_TWAP_LOG_CHUNK_BLOCKS`, which splits the window into requests small enough for the common 10k-block provider cap.

apps/backend/src/infrastructure/services/priceOracle/uniswapV4.ts
```diff
@@ -0,0 +1,411 @@
...
+  }
+  const reverted = error.walk((e) => e instanceof ContractFunctionRevertedError)
+  if (!(reverted instanceof ContractFunctionRevertedError)) {
+    return false
+  }
+  return (
+    reverted.data?.errorName === NOT_ENOUGH_LIQUIDITY ||
+    reverted.reason === NOT_ENOUGH_LIQUIDITY
+  )
+}
```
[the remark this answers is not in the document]
EmilFattakhov: Confirmed exactly as described, and fixed.

Reproduced against mainnet: with no error fragments in the ABI, viem leaves both `data` and `reason` undefined, so `isQuoterRevert` returned `false` for every revert. The actual top-level revert is `UnexpectedRevertBytes(0x7a5ed734… )` — `NotEnoughLiquidity(poolId)` wrapped — which is what the quoter's `try`/`catch` around `poolManager.unlock` re-throws.

The fragments are now declared, the wrapper is unwrapped and the inner error decoded, and the unwrapped form is still matched in case a future periphery release stops wrapping. The unit tests were the reason this got through: they hand-set `data.errorName`, which asserts that the classifier reads a field rather than that the field is ever populated. They now build real encoded revert data.

One thing the fix surfaced that this report did not cover: an *empty* pool reverts identically to an oversized trade, so the corrected classifier would have told users to buy less when no size at all could be filled. The liquidity read now separates the two, and an empty pool is reported as an outage.

apps/backend/src/infrastructure/services/priceOracle/index.ts
jim-counter: Failing closed is the preference here.

EmilFattakhov:
bugbot run

EmilFattakhov:
This has mostly taken shape, ready for another round of reviews - please note that there's been significant changes made to the PR, so it requires a full re-review

apps/backend/src/infrastructure/services/priceOracle/subgraph.ts
```diff
@@ -0,0 +1,318 @@
...
+        id
+        decimals
+      }
+      token1 {
+        id
+        decimals
+      }
+    }
+    swaps(
+      first: $first
```
jim-counter, line 141: The count truncates at the source, and `ORACLE_MAX_WINDOW_AGE_MS` is applied afterwards as a filter over the result (`index.ts:233`), so it can only ever shrink these ten. It can never reach back for an eleventh fill.

That makes a burst *evict* history rather than compete with it. Six fills are the six newest, so `first: 10` returns them plus the four next-newest and everything older is never fetched. Six buys a majority whether the pool has traded 11 times this week or 11,000, which is why the cost is flat rather than proportional to the market.

Verified against the module with identical attacker fills:

| Window | Result |
|---|---|
| Last 10: 6 attacker + 4 organic | serves, 40% below mid |
| All fills in 7d: 6 attacker + 11 organic | refused `market-moved` |

Suggest swapping the roles of the two settings already here: `timestamp_gt` in the `where` clause selects the window, and `first` becomes a generous cap on response size. A burst then dilutes instead of displacing, and the attacker has to out-number organic flow, which is what the `ORACLE_MIN_WINDOW_SPAN_MS` guard already assumes it forces.

One case to handle explicitly: a full page means the window was truncated and is a count window again, so it needs detecting rather than silently accepting.

apps/backend/src/infrastructure/services/priceOracle/subgraph.ts
```diff
@@ -0,0 +1,318 @@
...
+      token0: { id: string; decimals: string }
+      token1: { id: string; decimals: string }
+    } | null
+    swaps: GraphSwap[]
+  }
+  errors?: { message: string }[]
+}
+
+// A BigDecimal amount in whole tokens, signed by direction, as the base-unit
+// magnitude the price math works in. The sign carries no price information —
```
jim-counter: Half of this is right: `amount0` and `amount1` always carry opposite signs, so one of them is redundant. The remaining bit is not. The pool takes its fee from whichever leg is the input, so a buy prints above mid and a sell prints below, separated by roughly twice the fee.

Measured on these functions at a 1.1% swap fee: an all-buy window prices +1.10% against mid, an all-sell window -1.10%, and a mixed one lands near it. Sell-heavy is the under-collecting direction, and it needs no attacker, only a seller.

Not asking for a pricing change. It sits inside `USD_QUOTE_MARGIN`, it averages out as flow balances, and correcting it properly needs a hardcoded composed fee, which is the assumption the old `slot0` read existed to avoid.

Worth keeping the sign though, and exposing a buy/sell count on `SwapWindow` and `getHealth`. Nothing currently records whether a window was lopsided, so there is no way to check that the balanced-flow assumption holds, and that data only accumulates going forward. It also unlocks the single-direction option for the volume floor (see the comment on `quote.ts:183`).

Note that 1.1% is only the fee component, which is depth-independent. Each fill's own price impact adds to it and scales inversely with depth, so treat the measured figure as the floor of the bias rather than its expected size.

apps/backend/src/infrastructure/services/priceOracle/quote.ts
```diff
@@ -35,10 +37,148 @@ export const isWithinBounds = (
...
+  }
+  if (ai3Total <= 0n) {
+    throw new Error('Cannot average a window whose AI3 volume is zero')
+  }
+  return (usdcTotal * PRICE_CONVERSION_FACTOR) / ai3Total
+}
+
+// Total USDC base units traded across a window — the weight behind its average,
+// and the figure the thin-volume guard judges.
+export const windowVolumeUsdc = (samples: SwapSample[]): bigint =>
```
jim-counter, line 183: A buy and a sell each contribute their full USDC leg, so a round trip counts twice. Buy $500 of AI3 and sell it back: two fills, roughly $1000 of recorded volume, net position unchanged, capital committed $500.

That makes this a fee barrier of about 1.1% of its own nominal value rather than a capital barrier. At this pool's swap fee the 1000 USDC default costs an attacker roughly $11 to clear, and raising it to 10,000 costs them roughly $110.

The asymmetry is what makes the knob hard to tune: honest flow has to genuinely trade the full amount to clear a floor of N, while an attacker clears it for around 1% of N. Raising the floor until it deters the attacker raises the bar for organic flow a hundred times faster.

It matters most in combination with the `first: 10` window: the attacker supplies the majority of the fills, so the volume backing the average is entirely their own churn, and this guard certifies that churn as evidence of a real market. Fixing the window defangs most of it, since their volume then has to compete with organic volume rather than being all of it.

Two suggestions, in order of strength:

- `pool(id: $pool)` is already queried for token identity, and the standard Uniswap subgraph schema carries depth on that same object (`liquidity`, `totalValueLockedUSD`). Volume is a proxy for depth, and depth is right there in a query you are already making. Churn cannot manufacture it. Worth introspecting for the exact field names.
- If the sign is carried through (see the comment on `subgraph.ts:128`), counting a single direction here stops a round trip double-counting. That roughly doubles the cost of clearing the floor rather than removing the technique, so I would treat it as secondary to the depth check.

apps/backend/src/infrastructure/services/priceOracle/index.ts
```diff
@@ -121,49 +407,88 @@ const serveStaleOrError = (): Result<OraclePrice, OracleUnavailableError> => {
...
+  let window: Awaited<ReturnType<typeof buildWindow>>
+  try {
+    window = await internal.buildWindow()
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error)
+    logger.error(error, 'Price oracle: unexpected failure building the window')
+    window = err(
+      unavailable(
+        `unexpected failure building the window (${message})`,
+        'gateway',
```
jim-counter: This catch exists for throws from the statistics helpers, which are our own invariant violations rather than anything upstream. Labelling them `gateway` sends an operator to The Graph's status page to debug our bug, which is the same misdiagnosis `misconfigured` was split out to prevent.

None of the ten reasons currently means "the oracle itself failed". Worth adding one, since #811 renders these and this is the case where the answer is a stack trace rather than a wait.

apps/backend/src/infrastructure/services/priceOracle/subgraph.ts
```diff
@@ -0,0 +1,318 @@
...
+  apiKey: string | undefined,
+): SubgraphEndpoint => {
+  if (!override && !apiKey) {
+    throw new SubgraphConfigError(
+      'GRAPH_API_KEY is not set — the AI3/USD oracle cannot query the pool ' +
+        'subgraph through the gateway, so USDC payments cannot be quoted ' +
+        '(set GRAPH_SUBGRAPH_URL instead to point at an unauthenticated mirror)',
+    )
+  }
+  return { url: override || defaultSubgraphUrl(), apiKey }
```
jim-counter: When both are set the key is sent to the override URL. The docstring above already has the reasoning: the credential authenticates against the gateway and nothing else, and the override exists for local mirrors and test doubles that need no credential.

Suggest attaching `apiKey` only when the default gateway URL is in use. One line, and it makes the documented override safe to point anywhere.

apps/backend/src/infrastructure/services/priceOracle/subgraph.ts
```diff
@@ -0,0 +1,318 @@
...
+    throw new Error('Subgraph query returned no swaps collection')
+  }
+  assertPoolIdentity(body.data.pool)
+
+  // A swap with a zero leg has no price and would divide by zero downstream.
+  // Dropping it here rather than throwing is deliberate: it is one unusable
+  // row, not an unusable response, and the sample-count guard already decides
+  // whether what remains is enough.
+  const samples: SwapSample[] = body.data.swaps.flatMap((swap) => {
+    const ai3Amount = toBaseUnits(swap.amount0, AI3_DECIMALS)
```
jim-counter: `parseDecimalToScaledBigint` rejects anything that is not a plain decimal, and a throw here propagates out of the adapter into `buildWindow`'s catch as `gateway`. So one malformed amount fails the whole window and reports it as a network outage.

Worth confirming what graph-node actually emits for BigDecimal at the extremes, since scientific notation would trip this. Either way the handling looks inconsistent: a zero leg is dropped per-row a few lines below and left to the sample floor, while an unparseable one is fatal. Whichever is right, this should not surface as `gateway`.

apps/backend/src/infrastructure/services/priceOracle/index.ts
```diff
@@ -47,70 +52,351 @@ if (minScaled > maxScaled) {
...
+  // actually filled at. At 1.6 swaps/day the window needs days to catch up, and
+  // every purchase in between is charged at a price nobody traded.
+  //
+  // So: the newest fill in the window gets a veto. If it is itself an outlier
+  // against the rest, the window is describing a regime the market has left,
+  // and the answer is to stop quoting until enough fills agree again. Denial,
+  // not mispricing — the same direction every other guard here fails in. It
+  // costs nothing when the market is merely volatile-but-continuous, because
+  // then the newest fill sits inside the band.
+  if (newestSwapMs < windowNewestMs) {
```
jim-counter: This compares the newest surviving timestamp against the newest in-window one. If two fills share a timestamp, which same-block swaps do, and only one is trimmed, the maxima stay equal and the veto does not fire even though the newest fill was discarded.

Cheap to make exact by tracking the newest sample itself rather than its timestamp.

jim-counter (commented):
The design is shaping up nicely here.

One higher priority item item around the way we restrict samples based on count or date/time and a number of smaller items to look at.

Feeling like this is pretty close.

apps/backend/src/infrastructure/services/priceOracle/subgraph.ts
EmilFattakhov: Good finding, verified and confirmed! Implemented your suggestion

apps/backend/src/infrastructure/services/priceOracle/subgraph.ts
EmilFattakhov: Agreed on both halves. The legs always disagree in sign — checked across all 236 fills this pool has recorded — so one sign is redundant, but the direction isn't, and I've added it: SwapSample.direction, read off the USDC leg, with buyCount/sellCount on SwapWindow and therefore on getHealth. Recorded, not judged, and no pricing change.

apps/backend/src/infrastructure/services/priceOracle/quote.ts
EmilFattakhov: Both suggestions taken, in your order of strength.

Depth, primary: introspected the schema, and pool(id:) carries totalValueLockedToken1 — the pool's own USDC balance — so the new ORACLE_MIN_POOL_USDC_DEPTH floor costs no extra round trip and refuses with a new thin-liquidity reason. Deliberately not totalValueLockedUSD, which is a valuation through the subgraph's pricing paths (the same objection that keeps amountUSD out of the query), and not liquidity, since an in-range v4 L isn't a unit anyone can write a floor in. 

Volume, secondary: now judged on the window's larger side, so a round trip can't count twice. It never overstates committed capital, costs one-directional flow nothing, and at worst doubles the effective floor for perfectly balanced flow. The total is still reported, it just no longer decides.

apps/backend/src/infrastructure/services/priceOracle/index.ts
EmilFattakhov: Agreed, adjusted

apps/backend/src/infrastructure/services/priceOracle/subgraph.ts
EmilFattakhov: On what graph-node actually emits: all 236 fills this pool has recorded are plain decimals, at most 18 fractional digits. Exponent form isn't ruled out, BigDecimal renders through Rust's Display, which uses it for scales far from zero — but dropping such a row costs nothing real, since the only amounts extreme enough to reach it are dust that truncates to zero base units anyway or magnitudes neither token can represent.

EmilFattakhov:
Verified the recent changes and they look right to me, thanks for your review and this is ready for another one!

jim-counter (approved):
LGTM

PR #808, opened 2026-08-03
feat(intents): reject over-cap purchases before payment

EmilFattakhov:
Prerequisite step of epic #742 (USDC payments). Enforces the per-user credit cap **before** the user pays, instead of after an irreversible on-chain payment.

## Problem

`createIntent` takes no size, so the per-user cap is only enforceable after payment. `onConfirmedIntent` → `addCreditsToAccount` fails the cap check and parks the intent in `OVER_CAP` for admin review — but the money is already on-chain and irreversible, so resolution is a manual refund or a cap bump.

## What this does

- `POST /intents` accepts an optional `requestedBytes` (decimal string canonical; a JSON number is accepted while it is a safe integer).
- When supplied, a creation-time cap pre-check rejects a purchase with no headroom **before the user pays**, as `403 CREDIT_CAP_EXCEEDED` with the cap and current balance in `message`.
- The size gates creation and is then discarded. Nothing is persisted.
- **The purchase screen sends it**, so the check is live for the real flow rather than waiting on a later PR.
- **It is required when paying in USDC**, where the charge is the rate times the size and there is nothing to quote without it. Optional on AI3, which locks a per-byte rate and lets any payment settle.

## Why the size is optional on AI3 but required on USDC

Not a stylistic asymmetry — the two paths owe different things at creation. An AI3 intent can be created before anyone has decided how much to buy. A USDC intent has to name an amount to transfer, and that amount is the rate times the size.

Enforced in the use case rather than by making the field required on the endpoint: `POST /intents` is a documented API-key flow for third-party integrators and every one of them pays in AI3 today. Requiring it outright would break them to close a hole they cannot reach. The consequence, stated plainly: **an AI3 caller using an API key can still omit the size and skip the pre-check.** Closing that is a breaking API change and is not attempted here.

## A cap check already existed — client-side

`GET /credits/summary` returns `canPurchase` and `maxPurchasableBytes`, and the purchase screen disables over-cap packages and blocks over-cap custom amounts. That is a greyed-out button, not a server-side gate, and it leaves three holes: API-key callers never load the screen; `maxPurchasableBytes` is a snapshot taken when the screen mounts, so a second tab or a stale page guards with the wrong number; and the client validates the *package* while credits come from what is actually paid.

This PR is the server-side half.

## Why no column

`requestedBytes` is a parameter, not a record. Three uses were considered and only one would ever have read a stored copy:

| Use | Needs storage? |
|---|---|
| The cap pre-check | No — runs inside `createIntent`, in-memory |
| The USDC quote (#747) | No — computed in the same request; its *output* persists as `quoted_token_amount` |
| Rebuilding the effective rate at confirmation | Yes — different process, minutes later |

That third use is real, but it belongs to #747, and #747 is better served persisting the effective rate as **its own column** than reconstructing it from three others. What was left was a column written by one code path, read by none, and shaped exactly like a balance it can never agree with: credits come from `paymentAmount / shannonsPerByte`, so any over- or underpayment makes a recorded size and the granted bytes diverge silently.

## The pre-check is a fast-fail, not a reservation

Nothing is held. Two concurrent intents for the same account can both pass it, and their sizes may sum past the cap. That is accepted — which is why the advisory-lock-protected check stays where it is, in `createPurchasedCreditWithCapCheck`. A reservation would need an expiry sweep, release-on-failure and its own contention story; far more than the problem warrants.

It measures the same aggregate as the authoritative check and uses the same `>` boundary, so a purchase landing exactly on the cap is allowed by both. Not by coincidence — it reaches that query through `getRemainingCredits` rather than restating it, because a pre-check that waved through purchases the real check then rejects is the precise failure this exists to prevent.

It also does not bound what the user ends up with, because credits follow the payment, not `requestedBytes`. Paying more than quoted grants more than was pre-checked. **That drift is accepted by design**, and the authoritative check bounds it: re-measuring the real balance under the advisory lock, an overpayment can walk an account up to the cap but never past it, with the excess landing in `OVER_CAP` exactly as an unchecked purchase would.

## Also in here: a latent BigInt serialization bug

`res.json()` throws on a raw BigInt, and the three intent response sites stringified two of the six bigint fields while spreading the whole object. Replaced with one `serializeIntent` helper. The `token_*` fields are latently broken today only because nothing writes them yet — this fixes them before #747 makes them reachable.

## Two more bugs, surfaced by making the 403 reachable

Both are in the frontend, and both were invisible until a `createIntent` call could actually fail:

- **The error had nowhere to go.** Step 3's catch was a bare `console.error` with a comment saying wagmi would surface it — but `writeError` only covers the wallet call, so a failure before that made the button silently do nothing. Now surfaced.
- **The client read the wrong field.** It parsed `message` off the error body, which is the shape coded errors use; `HttpError`'s default is `{ error: <message> }`. Reading only `message` would have shown every 400 as its status text.

`mibToBytes` is shared with `isMibOverCap` so the client's own cap verdict and the number the server re-checks come from one conversion — and it equals what the payment grants, since `formatCreditsInMbAsValue` applies the same 1024² before `shannonsPerByte` and credits are `paymentAmount / shannonsPerByte`.

## Validation is split by concern

`parseRequestedBytes` handles wire shape; `createIntent` handles range, so a caller reaching the use case directly (i.e. #747) gets the same rules. A decimal string is canonical — every other size on an intent already crosses the wire as one, and it keeps working if the cap ever moves past 2^53. A JSON number is accepted while it is a safe integer, so no lossy float reaches the byte path and an oversized number is told to use a string rather than silently rounded.

## Forward note for #747

`usd_rate_at_creation` stores the **raw** rate the oracle reported. The user is charged that rate plus `USD_QUOTE_MARGIN`, so converting a received payment back to AI3 at the raw rate returns the margin as free storage on every purchase — the whole margin, exactly, since a rate is a scalar and the error scales with the amount. With no size persisted on the intent, there is nothing on the type to recover an effective rate from: **#747 must persist one.**

The field comments in `packages/models/src/users/intent.ts` state that contract rather than whichever mechanism currently prices it — the epic has had three rate sources and the mechanism-naming comments went stale on each.

## Verification

| Check | Result |
|---|---|
| `yarn models build && yarn s3 build && yarn backend build` | clean |
| `yarn backend lint` | clean |
| `yarn backend test` — unit suite | 659 passed / 659, 45 suites |
| `yarn frontend lint` + `tsc --noEmit` | clean |
| Cursor Bugbot | no issues |

Coverage: the no-`requestedBytes` regression path (asserting the balance is not even read), rejection of zero/negative/over-cap without pricing or a balance read, the exact-cap boundary, 18 `parseRequestedBytes` shape cases, a guard that the size is not persisted, and both halves of the USDC/AI3 asymmetry.

---


EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

apps/frontend/src/services/api.ts
```diff
@@ -133,11 +133,35 @@ export const createApiService = ({
...
-      throw new Error(`Network response was not ok: ${response.statusText}`);
+      // The cap rejection is the one failure here a user can act on, and it
+      // names the cap and their balance. Surfacing statusText instead would
+      // turn "you have 2 GiB of room left" into "Forbidden".
+      //
+      // Two body shapes, because the backend has two: errors carrying a
+      // machine-readable code send { error: CODE, message }, while the
+      // HttpError default sends { error: <the message> }. Preferring `message`
+      // and falling back to `error` reads the human-readable half of either.
+      const detail = await response
```
jim-counter, line 171: The new error-body passthrough forwards the server's message verbatim for every non-ok status. `handleInternalErrorResult` builds 500 bodies as `Failed to create intent: <raw exception>`, so a downed consensus WebSocket puts `connect ECONNREFUSED 10.0.3.7:9944` on screen under the Send button. Previously only response.statusText was used and pre-wallet failures rendered nothing at all.

Gate the passthrough to 4xx; 5xx falls back to the generic string. Keeps the `CREDIT_CAP_EXCEEDED` message.

apps/frontend/src/utils/credits.ts
```diff
@@ -3,6 +3,16 @@
...
  */
 
+/**
+ * Whole MiB to bytes.  The single conversion used both to decide locally
+ * whether a purchase fits under the cap and to tell the backend what size to
+ * check — the client's own verdict and the number the server re-checks must
+ * come from the same arithmetic, or a purchase can pass here and be rejected
+ * there.
+ */
+export const mibToBytes = (mib: number): bigint =>
```
jim-counter, line 13: `BigInt(mib)` has no integer guard, and index.tsx:80 re-hydrates `sizeMB` from the query string with a regex that accepts decimals. `/purchase?step=3&sizeMB=0.5` now throws `RangeError: The number 0.5 cannot be converted to a BigInt`, surfaced raw to the user. The same URL worked before the PR, because `formatCreditsInMbAsValue(0.5)` yields `BigInt(524288)`.

`BigInt(Math.round(mib))` matches the function's documented "whole MiB" contract and removes the class.

apps/backend/__tests__/unit/useCases/intents.spec.ts
```diff
@@ -63,6 +106,238 @@ describe('IntentsUseCases', () => {
...
+      .mockImplementation(async (intent) => intent)
+
+    const result = await IntentsUseCases.createIntent(orgUser, 1_073_741_824n)
+
+    expect(result.isOk()).toBe(true)
+    // The size gates creation and is then discarded. Persisting it would store a
+    // number that reads like a balance and never agrees with one, since credits
+    // come from paymentAmount / shannonsPerByte.
+    const created = createSpy.mock.calls[0][0]
+    expect(Object.keys(created)).not.toContain('quotedBytes')
```
jim-counter: **This assertion can never fail.** `quotedBytes` was removed from the model, row type, mapper and migration in 2f519fb, so `Object.keys(created)` cannot contain it regardless of what `createIntent` does.

That matters because this is the test the PR cites as the guard that the size is not persisted. Adding `requestedBytes` to the object passed to `intentsRepository.createIntent` (`core/users/intents.ts:283`) leaves the suite green.

Asserting on the value tests the actual invariant and survives future fields:

```suggestion
    // Nothing derived from the size may reach the row, under any key name.
    expect(Object.values(created)).not.toContain(1_073_741_824n)
```

apps/backend/src/app/controllers/intents.ts
```diff
@@ -2,16 +2,30 @@ import { Router } from 'express'
...
 import { IntentsUseCases } from '../../core/users/intents.js'
-import {
-  handleInternalError,
-  handleInternalErrorResult,
-} from '../../shared/utils/neverthrow.js'
+import { handleInternalErrorResult } from '../../shared/utils/neverthrow.js'
 import { handleError } from '../../errors/index.js'
 import { config } from '../../config.js'
 import { hasGoogleAuth } from '../../core/featureFlags/index.js'
+import { Intent } from '@auto-drive/models'
```
jim-counter, line 9: Does the Auto SDK or it's documentation need any update to support this change in behaviour?

jim-counter (commented):
A few low priority points.

apps/backend/src/app/controllers/intents.ts
EmilFattakhov: 100%, creating an issue that it's worth addressing after #808 is merged

apps/frontend/src/services/api.ts
EmilFattakhov: Great catch, fixed and also improved the test coverage 👍

apps/frontend/src/utils/credits.ts
EmilFattakhov: There's a drawback in using `BigInt(Math.round(mib))`, so I used the ` normaliseMib` instead to essentially eliminate the same problem

apps/backend/__tests__/unit/useCases/intents.spec.ts
EmilFattakhov: Good catch, addressed!

apps/frontend/src/utils/credits.ts
jim-counter: Nice, this is a better fix!

jim-counter (approved):
LGTM

PR #810, opened 2026-08-07
feat(intents): quote and lock a USDC purchase at creation

EmilFattakhov:
Prices a USDC intent at creation and locks the rate the payment converts back at. Step #747 of epic #742. **Stacked on #808** — review that first; this diff is only the commits on top.

**Off by default.** `usdc_eth` sits behind `payWithUsdc` (`PAY_WITH_USDC_ACTIVE`), with admins exempt. Nothing settles a USDC intent yet — the payment manager subscribes to the AI3 receiver only, and no watcher reads `intentTokenPaymentReceivedAbi` — so a quote created today would be a binding amount the backend cannot observe payment of, returned without an address to send it to. Merging this lands the pricing half; opening the flag waits on #748.

## The problem

An AI3 intent stores one number and credits are `payment_amount / shannons_per_byte`. That works because the user pays in the asset the price is denominated in.

USDC does not. The user pays dollars, and converting back to bytes needs the AI3/USD rate — but **the rate they were charged at is not stored anywhere.** `usd_rate_at_creation` is the raw oracle rate, and the user pays that plus `USD_QUOTE_MARGIN`. Convert a received payment at the raw rate and the margin comes back as free storage on every purchase — the whole margin, exactly, since a rate is a scalar so the error scales with the amount. It also grants more bytes than #808's cap pre-check was run against.

## `quoted_ai3_shannons`, and why a rate column would not do

`quoted_token_amount` is what the user was charged. This new column is what they were charged **for**. The pair *is* the effective rate.

It is a pair rather than a single rate column because a rate here cannot be stored exactly. USDC carries 6 decimals against byte counts near 1e11, so USDC-per-byte is ~0.0027 base units — deeply sub-unit, and once scaled to survive as an integer it is a rounded ratio that no longer round-trips the quote. Two exact integers do:

```
bytes = token_amount * quoted_ai3_shannons / quoted_token_amount / shannons_per_byte
```

Pay exactly what was quoted and the ratio cancels: the user receives exactly the size they asked for, with no rounding in either direction. Overpay and it scales proportionally at the same effective rate.

Worked example — at 1 USDC = 1 AI3 = 1 GB with a 5% margin, a 1 GB purchase is quoted at 1.05 USDC and grants **1 GB, not 1.05 GB**. The margin is part of the price, not a payment for extra storage; it exists to cover FX drift until an operator converts the batch, and handing it back as bytes would make it cover nothing. Both behaviours are pinned by tests — one asserting the exact grant, one asserting that the raw-rate conversion over-credits.

## Also in the quoting path

- **`paymentMethod` on `POST /intents`**, defaulting to AI3 for body-less requests. An unrecognised value is rejected rather than defaulted: silently treating `'usdc'` as AI3 would quote in the wrong asset and the caller would find out at payment time.
- **Ordering**: flag, then validation, then cap pre-check, then the rate. A purchase that cannot be granted never costs a subgraph round-trip, and a failed quote writes no row.
- **Oracle failures are 503s**, `PRICE_UNSTABLE` distinguished from `PRICE_ORACLE_UNAVAILABLE`. There is deliberately no size-related code — see the rebase note.
- **A last-good (stale) rate still prices a binding quote.** Deliberate: the number is a days-long average, and refusing through every subgraph blip would shut the path further than the drift justifies.
- **An intent at a zero per-byte price is refused at creation**, on both payment methods. Every payment against one converts to zero credits and lands in FAILED with the money kept; on the USDC path the quote itself computes to 0 — a binding charge of "nothing" for a purchase that will never be granted. A 503, not a 4xx: the request was fine, the deployment is not.

## Rebased onto the subgraph-VWAP oracle

This branch was written against the oracle #807 replaced. Three things had to change, and none was mechanical:

| Written against | Now |
|---|---|
| `getExecutableQuote(ai3)` — priced a specific size | `getPrice()` — one size-independent rate, applied by multiplication |
| `409 QUOTE_TOO_LARGE` / `400 QUOTE_AMOUNT_INVALID` | deleted — no oracle failure is about the requested size, so no 4xx can tell the user to ask for less |
| over-credit test asserting >600bps (fee + impact + margin) | asserts against the configured margin, now the entire wedge |

The swap fee and price impact did not disappear — they moved inside the rate, which averages realized fills, and those fills paid both. What no longer exists is *this* purchase's own impact, because the treasury does not swap per intent.

## A payment bug found in review

`payIntent(bytes32)` on the AI3 receiver takes **any** intent id from anyone — no registry, no allowlist, just a non-zero value check — and the watcher reports it as `paymentAmount` regardless of what the intent expects. So an AI3 payment against a USDC intent arrived as a well-formed confirmation.

The row went CONFIRMED with `token_amount` NULL; `onConfirmedIntent` looked for the USDC column, found nothing, and returned an error **without writing a terminal status** — so the 30-second poller retried it forever. Payment kept, no credits, no admin row. Worse, the idempotency guard then treated the intent as settled, so the user's real USDC payment would be silently discarded on arrival.

Griefable rather than merely accidental: the intent id is visible in the calldata of a pending Ethereum tx, and racing Auto EVM against Ethereum confirmations is easy.

Fixes, defence in depth:

1. **`markIntentAsConfirmed` refuses an asset mismatch** before writing. The intent stays PENDING and expires on its own schedule. Ordered after the idempotency guard, so re-delivery for a settled intent stays a no-op.
2. **A confirmed-but-amountless intent is marked FAILED** rather than retried. Nothing writes that column after confirmation, so the state can never resolve itself.
3. **`getIntentCredits` returns 0 when `shannonsPerByte` is 0** — BigInt division by zero throws.
4. **`_checkConfirmedIntents` catches per intent.** It checked `result.isErr()`, which only covers what `onConfirmedIntent` *returns*; a **thrown** exception escaped the loop and was swallowed by the `safeCallback` on the interval, abandoning every intent queued behind it — from users who paid correctly, on every tick, with nothing written to explain why. The guard fixes the cause we found; the catch bounds the blast radius of the ones we did not. The existing batch tests only ever returned errors, which were always handled; the new one throws mid-batch and asserts the intent *after* the thrower still runs.

## Refusing is not resolving: `intent_mispayments`

Refusing a mispayment is correct, but it settles nothing on chain — the transfer happened, and a log line is not something an admin queries. Refused payments are now recorded, with the tx hash threaded from the watcher: an amount and a sender describe a payment, but only the hash finds it again. `GET /intents/mispayments` (admin) reads it, mirroring `/intents/over-cap` — that queue is for payments we accepted and could not convert, this one for payments we never accepted at all.

Two cases, and they are not equally new:

- **`ASSET_MISMATCH`** — the intent exists but is denominated in the other asset. New with USDC.
- **`UNKNOWN_INTENT`** — the id matches no row. **This is reachable on `main` today**, by anyone with a wallet, and until now produced only a log line. It is also the case with the least evidence anywhere else, since no intent row shows the money arrived at all.

Design notes: no foreign key to `intents`, because the unknown-intent case has nothing to point at — a FK would make the table incapable of recording precisely the case that needs it most. `ON CONFLICT DO NOTHING` on `(tx_hash, intent_id)`, because the watcher replays: reorgs re-emit events and the startup sweep re-runs every PENDING intent carrying a tx hash, so without it one mispayment becomes a fresh row per restart. Recording never throws — failing to file the paperwork must not change which error actually happened, and on the startup-sweep path a throw would abort the recovery of unrelated transactions.

## Down-migration is destructive

Credit derivation routes purely on `payment_method` and does **not** fall back to the AI3 formula, so a `usdc_eth` row without `quoted_ai3_shannons` yields 0 credits and is marked FAILED. Running the down-migration while paid USDC intents are in flight strands them. Drain or settle first.

## Verification

| Check | Result |
|---|---|
| `yarn models build && yarn backend build` | clean |
| `yarn backend lint` | clean |
| intents + PaymentManager + intentMispayments suites | 135 passed |
| `yarn frontend lint` + `tsc --noEmit` | clean |

---

### DIFF
--- a/apps/backend/migrations/sqls/20260806000000-intent-quoted-ai3-shannons-up.sql
+++ b/apps/backend/migrations/sqls/20260806000000-intent-quoted-ai3-shannons-up.sql
@@ -1,0 +1,37 @@
+-- Record the AI3 amount a USDC quote was priced for.
+--
+-- A USDC intent needs to convert the payment it receives back into storage
+-- bytes, and the only rate that makes "pay the quote, receive the quote" true is
+-- the rate the user was actually quoted at. Neither existing column carries it:
+--
+--   usd_rate_at_creation is the pool's MARGINAL price. The user pays the
+--   executable quote, which is that price plus the pool swap fee, plus the
+--   price impact of their own size, plus the quote margin. Converting a received
+--   payment at the marginal rate hands all three back as free storage — 5-8% on
+--   a realistic purchase — and grants more bytes than the pre-payment cap check
+--   was run against.
+--
+--   quoted_token_amount is one half of the rate: what was charged. This column
+--   is the other half: what it was charged FOR.
+--
+-- Together the pair IS the effective rate, held exactly as two integers rather
+-- than as a rounded ratio. That matters because USDC has 6 decimals against
+-- byte counts of ~1e11, so a usdc-per-byte or usdc-per-shannon rate is deeply
+-- sub-unit (~0.0027 base units per byte at 100 GiB) and cannot be stored as an
+-- integer without scaling — and once scaled it is a rounded ratio that no longer
+-- round-trips the quote exactly. Storing the two quantities keeps the
+-- confirmation-path conversion exact:
+--
+--   bytes = token_amount * quoted_ai3_shannons
+--             / quoted_token_amount / shannons_per_byte
+--
+-- With token_amount = quoted_token_amount the ratio cancels and the result is
+-- the requested size exactly, with no rounding in either direction.
+--
+-- numeric(78,0) to match the bigint base-unit convention already used by
+-- payment_amount / shannons_per_byte / quoted_token_amount. NULL for AI3_NATIVE
+-- intents and for rows created before USDC support, both of which derive credits
+-- from payment_amount / shannons_per_byte and never read this.
+
+ALTER TABLE intents
+  ADD COLUMN IF NOT EXISTS quoted_ai3_shannons numeric(78,0);
--- a/apps/backend/migrations/sqls/20260806000000-intent-quoted-ai3-shannons-up.sql
+++ b/apps/backend/migrations/sqls/20260806000000-intent-quoted-ai3-shannons-up.sql
@@ -30,8 +30,16 @@
 --
 -- numeric(78,0) to match the bigint base-unit convention already used by
 -- payment_amount / shannons_per_byte / quoted_token_amount. NULL for AI3_NATIVE
--- intents and for rows created before USDC support, both of which derive credits
--- from payment_amount / shannons_per_byte and never read this.
+-- intents, which derive credits from payment_amount / shannons_per_byte and never
+-- read this.
+--
+-- No existing row needs a backfill: nothing could create a usdc_eth intent before
+-- this column existed. Note that credit derivation routes purely on
+-- payment_method and does NOT fall back to the AI3 formula, so a usdc_eth row
+-- without this column yields 0 credits and is marked FAILED for admin review.
+-- That is the intended outcome for a corrupt row, but it is also why the
+-- down-migration is destructive in a way a column drop usually is not: running it
+-- while paid USDC intents are in flight strands them. Drain or settle them first.
 
 ALTER TABLE intents
   ADD COLUMN IF NOT EXISTS quoted_ai3_shannons numeric(78,0);
--- a/apps/backend/migrations/sqls/20260806000000-intent-quoted-ai3-shannons-up.sql
+++ b/apps/backend/migrations/sqls/20260806000000-intent-quoted-ai3-shannons-up.sql
@@ -4,12 +4,11 @@
 -- bytes, and the only rate that makes "pay the quote, receive the quote" true is
 -- the rate the user was actually quoted at. Neither existing column carries it:
 --
---   usd_rate_at_creation is the pool's MARGINAL price. The user pays the
---   executable quote, which is that price plus the pool swap fee, plus the
---   price impact of their own size, plus the quote margin. Converting a received
---   payment at the marginal rate hands all three back as free storage — 5-8% on
---   a realistic purchase — and grants more bytes than the pre-payment cap check
---   was run against.
+--   usd_rate_at_creation is the RAW rate the oracle reported. The user pays that
+--   rate plus USD_QUOTE_MARGIN, so converting a received payment at the raw rate
+--   hands the margin back as free storage on every purchase — the whole margin,
+--   exactly, since a rate is a scalar and the error scales with the amount — and
+--   grants more bytes than the pre-payment cap check was run against.
 --
 --   quoted_token_amount is one half of the rate: what was charged. This column
 --   is the other half: what it was charged FOR.
--- a/apps/backend/migrations/sqls/20260812000000-intent-mispayments-up.sql
+++ b/apps/backend/migrations/sqls/20260812000000-intent-mispayments-up.sql
@@ -1,0 +1,49 @@
+-- Record on-chain payments that were refused, so an irreversible transfer is
+-- never left with only a log line pointing at it.
+--
+-- Both receivers accept any intent id from anyone — payIntent(bytes32) on Auto
+-- EVM, payIntentWithToken(bytes32, uint256) on Ethereum — so a payment can name
+-- an intent that does not exist, or one denominated in the other asset.
+-- markIntentAsConfirmed refuses both, which is correct: confirming a mispayment
+-- strands the intent in the polling loop AND makes the idempotency guard discard
+-- the user's real payment when it arrives. But refusing resolves nothing on
+-- chain. The money has moved; the only open question is whose it is, and that
+-- question needs a row.
+--
+-- No foreign key to intents on purpose. The unknown-intent case has no row to
+-- point at, and it is exactly the case with the least other evidence attached.
+
+CREATE TABLE IF NOT EXISTS intent_mispayments (
+  id bigserial PRIMARY KEY,
+  -- The id named by the on-chain event, which may match no intent.
+  intent_id text NOT NULL,
+  -- IntentMispaymentReason: 'unknown_intent' | 'asset_mismatch'. Text rather
+  -- than an enum type so a new reason is a code change, not a migration.
+  reason text NOT NULL,
+  -- What the named intent was denominated in; NULL when there is no such intent.
+  expected_payment_method text,
+  -- Whichever the watcher reported. Exactly one is set, and which one is itself
+  -- the evidence: an AI3 amount against a USDC intent is the asset mismatch.
+  -- numeric(78,0) to match the bigint base-unit convention used across intents.
+  payment_amount numeric(78,0),
+  token_amount numeric(78,0),
+  from_address text,
+  tx_hash text,
+  created_at timestamptz NOT NULL DEFAULT NOW()
+);
+
+-- The watcher re-runs: chain reorgs re-emit events, and the startup sweep
+-- replays every PENDING intent that has a tx_hash. Recording is therefore
+-- idempotent per (transaction, intent) rather than append-only, or one restart
+-- loop would turn a single mispayment into a page of them.
+--
+-- Partial because a NULL tx_hash cannot participate in the constraint anyway
+-- (NULLs are distinct in a unique index) — stating that in the predicate keeps
+-- the index honest about what it enforces, and small.
+CREATE UNIQUE INDEX IF NOT EXISTS intent_mispayments_tx_intent_uniq
+  ON intent_mispayments (tx_hash, intent_id)
+  WHERE tx_hash IS NOT NULL;
+
+-- The admin listing reads newest-first and nothing else reads this table.
+CREATE INDEX IF NOT EXISTS intent_mispayments_created_at_idx
+  ON intent_mispayments (created_at DESC);
--- a/apps/backend/src/core/users/intents.ts
+++ b/apps/backend/src/core/users/intents.ts
@@ -1,5 +1,6 @@
 import {
   Intent,
+  IntentMispaymentReason,
   IntentStatus,
   PaymentMethod,
   User,
@@ -7,6 +8,7 @@ import {
   UserWithOrganization,
 } from '@auto-drive/models'
 import { intentsRepository } from '../../infrastructure/repositories/users/intents.js'
+import { intentMispaymentsRepository } from '../../infrastructure/repositories/users/intentMispayments.js'
 import { purchasedCreditsRepository } from '../../infrastructure/repositories/users/purchasedCredits.js'
 import { EventRouter } from '../../infrastructure/eventRouter/index.js'
 import { MAX_RETRIES } from '../../infrastructure/eventRouter/tasks.js'
@@ -20,12 +22,14 @@ import {
   QuoteErrorCode,
   QuoteFailedError,
   ServiceUnavailableError,
+  UsdcPaymentsDisabledError,
 } from '../../errors/index.js'
 import { err, ok, Result } from 'neverthrow'
 import { config } from '../../config.js'
 import { randomBytes } from 'crypto'
 import { createLogger } from '../../infrastructure/drivers/logger.js'
 import { AccountsUseCases } from './accounts.js'
+import { FeatureFlagsUseCases } from '../featureFlags/index.js'
 import { transactionByteFee } from '@autonomys/auto-consensus'
 import { ApiPromise, WsProvider } from '@polkadot/api'
 import { priceOracle } from '../../infrastructure/services/priceOracle/index.js'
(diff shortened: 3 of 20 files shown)


EmilFattakhov:
bugbot run

EmilFattakhov:
The idea behind that PR was rather simple and it since has grown significantly, and now cover some of the edge-cases especially around payments with no intents. I've chatted with multiple agents about it and tried simplifying the design, but all highlighted that this bit is well worth handling. Curious to hear if you think differently, or prefer me to split them up in a few different PRs.

apps/backend/migrations/sqls/20260806000000-intent-quoted-ai3-shannons-up.sql
```diff
@@ -0,0 +1,44 @@
+-- Record the AI3 amount a USDC quote was priced for.
+--
+-- A USDC intent needs to convert the payment it receives back into storage
```
jim-counter, line 3: The user selects the amount of storage they want to buy when they create the intent.

apps/backend/src/core/users/intents.ts
```diff
@@ -389,17 +709,122 @@ const markIntentAsConfirmed = async ({
...
+  //
+  // Confirming it anyway is what makes it dangerous. The row would go CONFIRMED
+  // with payment_amount set and token_amount NULL, onConfirmedIntent would look
+  // for the USDC column, find nothing, and the intent would sit in the 30-second
+  // polling loop indefinitely — payment kept, no credits, and no terminal row for
+  // an admin to find. Worse, the idempotency guard above would then treat the
+  // intent as settled, so the user's real USDC payment would be silently
+  // discarded when it arrived.
+  //
+  // Refusing leaves the intent PENDING so it expires on its own schedule. The
```
jim-counter: The comment here says refusing "leaves the intent PENDING so it expires on its own schedule". That holds only when `tx_hash` is NULL.

- `isIntentExpired` returns `false` for any PENDING intent carrying a `txHash` ([intents.ts:89](apps/backend/src/core/users/intents.ts:89)).
- `getExpiredPendingIntents` filters on `AND tx_hash IS NULL` ([repositories/users/intents.ts:144](apps/backend/src/infrastructure/repositories/users/intents.ts:144)), so `cleanupExpiredIntents` never sees the row.

**Scenario.** A user creates a `usdc_eth` intent, pays in AI3 by mistake, then calls `POST /intents/:id/watch` with that hash. `triggerWatchIntent` writes `tx_hash` to the row with no on-chain validation that the transaction pays this intent, or that it exists ([intents.ts:562](apps/backend/src/core/users/intents.ts:562)). The AI3 event then reaches the asset-mismatch guard and is refused, leaving the intent PENDING.

From that point the row can reach neither EXPIRED nor CONFIRMED:

- `getIntent` keeps returning it as live and payable, indefinitely past its price-lock window.
- `getPendingWithTxHash` ([repositories/users/intents.ts:172](apps/backend/src/infrastructure/repositories/users/intents.ts:172)) hands it to `_recoverOrphanedTransactions` ([paymentManager/index.ts:162](apps/backend/src/infrastructure/services/paymentManager/index.ts:162)) on every service restart, to be re-watched and re-refused.

**Reachability.** Admin-only today, since it needs a `usdc_eth` intent to exist. It cannot fire on an AI3 intent at all, because `watchTransaction` always supplies `paymentAmount`. It becomes user-reachable the moment `PAY_WITH_USDC_ACTIVE` opens, which is why it is worth closing now rather than at #748.

**Two ways out.**

- *Targeted:* clear the intent's `tx_hash` on refusal when it matches the refused `txHash`, so the expiry sweep can reclaim the row.
- *General:* let the sweep reclaim PENDING rows whose `expires_at` passed by some grace margin regardless of `tx_hash`. This covers every submitted-but-never-confirming hash rather than just this one, including an RPC outage that outlives the grace window.

I would lean to the general fix. The current exclusion assumes a `tx_hash` means "actively being watched and will resolve", and a refused payment is a standing counterexample.

apps/backend/migrations/sqls/20260812000000-intent-mispayments-up.sql
```diff
@@ -0,0 +1,49 @@
...
+
+-- The watcher re-runs: chain reorgs re-emit events, and the startup sweep
+-- replays every PENDING intent that has a tx_hash. Recording is therefore
+-- idempotent per (transaction, intent) rather than append-only, or one restart
+-- loop would turn a single mispayment into a page of them.
+--
+-- Partial because a NULL tx_hash cannot participate in the constraint anyway
+-- (NULLs are distinct in a unique index) — stating that in the predicate keeps
+-- the index honest about what it enforces, and small.
+CREATE UNIQUE INDEX IF NOT EXISTS intent_mispayments_tx_intent_uniq
```
jim-counter: `payIntent(bytes32)` is payable and callable from a contract, so a single transaction can emit `IntentPaymentReceived` more than once for the same intent id with different values.

`watchTransaction` parses every matching log and calls `markIntentAsConfirmed` for each inside a `Promise.all` ([paymentManager/index.ts:57](apps/backend/src/infrastructure/services/paymentManager/index.ts:57)), so `record` runs twice with an identical key. The second hits `ON CONFLICT DO NOTHING` ([intentMispayments.ts:64](apps/backend/src/infrastructure/repositories/users/intentMispayments.ts:64)) and returns null, and `recordMispayment` deliberately logs nothing on a null return, treating it as an expected replay.

**Consequence.** The admin queue records one payment, of whichever of the two amounts won the race, when both arrived. The recorded total understates the refund owed, in the one table whose stated purpose is that an irreversible transfer is never left with only a log line pointing at it.

**Reachability.** The variant reachable today is UNKNOWN_INTENT, where the money is the sender's own, so real harm waits on the flag. The variant that loses a genuine user's second transfer needs a `usdc_eth` intent.

**Why now.** The key is a unique index in a migration. Changing it before any rows exist is free; changing it afterwards is a data migration over records that are, by construction, the only durable evidence of unresolved on-chain payments.

**Suggested fix.** Carry the event's log index through `markIntentAsConfirmed` alongside `txHash`, and key on `(tx_hash, log_index) WHERE tx_hash IS NOT NULL`. viem already exposes `log.logIndex` on the parsed logs, and a log index is unique within a transaction, so `intent_id` is not needed in the key. Replay de-duplication still works, because a re-emitted event carries the same log index.

jim-counter (commented):
After reading this through I was asking myself if there's a possibility we're over-complicating things a little here. The user asks to purchase a certain amount of storage when they create the intent. We could simplify the onward process to be:

1. Convert storage amount to AI3
2. Convert AI3 amount to USDC
3. Accept USDC payment if it is at or above the amount quoted
4. Reject USDC payment if it is below the amount quoted

This feels more natural to the way the user commits to the purchase. More detailed breakdown below. What do you think?

# Design proposal: commit to the purchase, not to the rate

## The observation

The USDC intent stores an effective rate and reconstructs the grant from it at settlement: `getIntentCredits` computes `tokenAmount * quotedAi3Shannons / quotedTokenAmount / shannonsPerByte` ([intents.ts:822](apps/backend/src/core/users/intents.ts:822)). The "pay the quote, receive the quote" property is not a stored fact, it is an algebraic accident of that expression: with `tokenAmount == quotedTokenAmount` the ratio cancels and you land back on `requestedBytes`. Any other amount is extrapolated linearly.

The product commitment is not a rate, though. It is a fixed price: the API contract advertises `quotedTokenAmount` as "the exact USDC amount to pay, locked until `expiresAt`". So the settlement model and the promise are different shapes.

Two things suggest the storage decision was made one level too low:

1. **`quoted_ai3_shannons` is strictly derivable.** It is computed as `requestedBytes * shannonsPerByte` ([intents.ts:461](apps/backend/src/core/users/intents.ts:461)), and `shannonsPerByte` is already on the row. It is a denormalisation of `requested_bytes`, not independent information.
2. **The "no reader" argument is circular.** [intents.ts:438](apps/backend/src/core/users/intents.ts:438) declines to persist `requestedBytes` on the grounds that nothing downstream reads it. That is only true because the grant is derived from a rate. Under a fixed-price commitment, `requested_bytes` is the thing you grant, and it becomes the primary reader.

**Proposal:** persist `requested_bytes` instead of `quoted_ai3_shannons`. At settlement, if `tokenAmount` is within a configured tolerance of `quotedTokenAmount`, grant the stored `requested_bytes`. Outside the tolerance, refuse and file an `intent_mispayments` row.

## Comparison

| Concern | As implemented | Proposed |
|---|---|---|
| What the intent commits to | An effective rate, held as the `quotedTokenAmount` : `quotedAi3Shannons` pair | The purchase: `requested_bytes` for `quotedTokenAmount` |
| Column added by the migration | `quoted_ai3_shannons numeric(78,0)` | `requested_bytes numeric(78,0)` |
| Settlement calculation | `tokenAmount * quotedAi3Shannons / quotedTokenAmount / shannonsPerByte` | Tolerance check, then return stored `requested_bytes` |
| Exact payment | Exactly `requestedBytes`, by cancellation, pinned by a regression test | Exactly `requested_bytes`, by construction |
| Underpayment | Silent proportional grant, intent COMPLETED, no signal the quote was missed | Outside tolerance: refused, filed for admin, intent stays PENDING |
| Overpayment | Proportional extra storage, past the amount the cap pre-check ran against | Within tolerance: excess kept as margin. Beyond: policy required |
| Precision reasoning required | 6 decimals against byte counts near 1e11, ceiling direction, division order, no intermediate floors | None at settlement. Only the creation-side quote rounds |
| "Wrong rate at settlement" bug class | Representable, and the reason this PR exists (`usdRateAtCreation` vs the effective rate) | Not representable. No rate is applied at settlement |
| Cap pre-check | Indicative. The grant can exceed what was checked ([intents.ts:199](apps/backend/src/core/users/intents.ts:199)) | Exact for this path |
| USDC dust guard | Required. A conversion can floor to 0 credits, routing to FAILED | Not applicable. A stored positive byte count cannot convert to 0 |
| Corrupt-row surface | Three columns must all be present, or 0 credits and FAILED | One column |
| New tuneables | None | Payment tolerance in USDC base units, plus a minimum quote floor derived from it |
| AI3 path | Rate-based, any payment settles | Unchanged |

## The trade, stated plainly

This is not a free simplification. The rate model's real virtue is that every payment converts to *something*, so no payment needs human resolution. The proposal gives that up: an out-of-band payment becomes an admin row and a user with no storage until someone acts.

So the trade is **silent-wrong-amount for loud-no-grant**. I think that is the right way round. A silent partial grant is neither detectable nor recoverable unless the user notices their balance looks short, whereas a recorded refusal is both, and `intent_mispayments` already exists for exactly the "money arrived, we would not attach it" case. But it should be an explicit decision, not a side effect.

It also moves complexity from arithmetic into policy. Policy is reviewable in a way that implicit arithmetic is not, which is part of the appeal, but the far-above-band case genuinely needs writing down, and does under the current design too.

## Tolerance shape

An absolute amount in USDC base units, not a percentage, and asymmetric.

- USDC is not fee-on-transfer, and `payIntentWithToken(bytes32, uint256)` records the amount actually transferred, so nothing mechanically shaves value in transit. The below-quote side of the band exists only to absorb user and UI error, which is an absolute-size problem rather than a proportional one.
- A symmetric 1% band would also quietly reduce the 5% `USD_QUOTE_MARGIN` to about 4% on every payment landing under the quote. That may be acceptable, but it should be chosen rather than inherited from the band's shape.
- Suggested rule: accept `tokenAmount >= quotedTokenAmount - tolerance`, generous above up to whatever the refund threshold is.

### Minimum quote floor

An absolute tolerance assumes the quote is large relative to it, and small purchases break that assumption. A 1-byte purchase quotes at roughly 2 base units after `ai3ShannonsToUsdcBaseUnits` ceilings to 1 and the margin rounds up to 2. A tolerance of 0.01 USDC is then five thousand times the price, and effectively any payment settles the intent for the full grant.

Refuse a `usdc_eth` intent at creation whose `quotedTokenAmount` falls below some multiple of the tolerance, and derive the floor from the tolerance rather than configuring both independently so the two cannot be tuned into contradiction. This subsumes the separate zero-`quotedTokenAmount` hole as a special case rather than adding a second guard for it.

## Consequences for this PR

- The migration becomes a column swap, not an addition, which is why this is worth settling before it lands rather than after.
- Retired: the pair-of-integers reasoning, the exactness commentary, the over-credit regression test, the USDC branch of `getIntentCredits` ([intents.ts:810](apps/backend/src/core/users/intents.ts:810) to :826), and the separate zero-quote guard.
- Unaffected: the feature flag and admin exemption, the oracle error mapping and its 503s, the asset-mismatch refusal and `intent_mispayments`, the per-intent catch in `_checkConfirmedIntents`, and the zero `shannonsPerByte` guard.
- Still worth fixing independently, since neither depends on the settlement model: the stranded PENDING intent at [intents.ts:727](apps/backend/src/core/users/intents.ts:727), and the mispayment dedup key.
- Adjacent hole, present under either model: a second payment against an already-settled intent is swallowed by the idempotency guard at [intents.ts:690](apps/backend/src/core/users/intents.ts:690), which returns `ok(intent)` before reaching the refusal and recording path. Pay-in-two-transfers today confirms on the first and files nothing for the second. The proposed model makes this more visible, since each below-band transfer is refused and recorded rather than the first silently confirming a partial grant.

EmilFattakhov:
Thanks for your review Jim! I had a rather interesting conversion with two agents, and actually had two of them defend their own case for my and other agent judgement. Whats changes (below is a simplified version taken from Claude notes):

Most of this is on the branch now:

- `af615050` — dedup key moved to `(tx_hash, log_index)`
- `d47fc4e8` — expiry grace, so a stale `tx_hash` can't exempt a row forever
- `177cabab` — a second payment against a settled intent gets recorded
- `9b836bb1` — off-quote payments get recorded
- `ec7e23f0` — dust payments get recorded

Two of those need a word. Expiring `tx_hash` rows on its own would have made
things worse: the idempotency guard treated EXPIRED as "already processed", so a
payment arriving after expiry just vanished. That's reachable on `main` today, so
it's fixed in the same commit.

And your detectability point is the one that landed. Silent was the problem, not
proportional. An off-quote payment now files a row and still gets credited, which
is ~20 lines instead of a new settlement model.

On the column swap: not doing it since  quoted_ai3_shannons` is `requestedBytes * shannonsPerByte` and
`shannons_per_byte` is on the row, so `requested_bytes` comes back exactly.
Switching later is code only, no backfill. You're right that it's derivable — it
just goes both ways.

Why I'd rather keep the rate model: you can't bounce a transfer, so every amount
needs an answer that doesn't involve a person, and `intent_mispayments` has no
grant path out today (`reprocessOverCapIntent` only takes OVER_CAP).
Band-or-refuse also splits settlement per asset. And I couldn't find a tolerance
I liked — 0.01 USDC is ~3.5MB at realistic prices, and a few base units makes it
a cliff.

Happy to revisit at #748. If the payment amount ends up as immutable calldata
built from the quote, none of this fires anyway.

apps/backend/migrations/sqls/20260806000000-intent-quoted-ai3-shannons-up.sql
EmilFattakhov: Fair point, added a note that the size is known at creation and `requested_bytes` and `quoted_ai3_shannons` recover each other exactly since `shannons_per_byte` is on the same row. So the choice is about what
settlement reads, not about what we can reconstruct later.

apps/backend/src/core/users/intents.ts
EmilFattakhov: Good catch! Fixed in d47fc4e8, went with the general version you preferred.

apps/backend/migrations/sqls/20260812000000-intent-mispayments-up.sql
EmilFattakhov: Implemented the suggestion 👍

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

EmilFattakhov:
bugbot run

apps/backend/migrations/sqls/20260806000000-intent-quoted-ai3-shannons-up.sql
jim-counter: This was more a pointer to the suggestion that we could avoid complexity by delivering the amount of storage requested. This was the best I could come up with based on your unease about the drift from a simple solution.

apps/backend/src/core/users/intents.ts
```diff
@@ -379,27 +745,353 @@ const markIntentAsConfirmed = async ({
...
+    paymentAmount,
+    tokenAmount,
+    fromAddress,
+    txHash,
+  })
+
+  // Lost the transition: something else moved the intent out of PENDING between
+  // the read above and this write. Filed the same way as the guard above, reached
+  // a different way. Nothing is retried and nothing is overwritten.
+  if (!confirmed) {
```
jim-counter: This branch treats losing the transition as proof a second payment arrived. It also fires when the *same* payment is delivered twice and one delivery loses the race, filing a mispayment for money that arrived once.

`paymentManager.start()` runs in both [frontend.ts:25](apps/backend/src/app/servers/frontend.ts:25) and [frontendWorker.ts:72](apps/backend/src/app/servers/frontendWorker.ts:72), so two processes each run an independent `watchContractEvent` over the same contract, and [processors/frontend.ts:39](apps/backend/src/infrastructure/eventRouter/processors/frontend.ts:39) calls `watchTransaction` again for the queued `watch-intent-tx` task. All of them call `markIntentAsConfirmed` with identical arguments, same `txHash` and same `logIndex`, unserialised across processes. Both read PENDING, one wins `confirmIntentIfPending`, the loser lands here.

Reproduced: two concurrent calls with identical `txHash` and `logIndex` leave the intent correctly CONFIRMED at 100 and file

```
{ reason: "already_settled", paymentAmount: "100", txHash: "0xtx", logIndex: 0 }
```

The `(tx_hash, log_index)` dedup cannot suppress it, because the winner files nothing to collide with.

**Fix:** apply the settled guard's comparison against the row read back before choosing a reason. If `current.txHash === txHash` and the settled amount equals the incoming amount, it is re-delivery, not a second payment, and nothing should be filed.

The existing test for this invariant (`leaves a true replay of the settling payment silent`) starts at `status: COMPLETED`, so it only covers the sequential path. Worth a same-log-index case next to the existing different-index one.

jim-counter (commented):
One small bug to address and I think we're good here.

apps/backend/src/core/users/intents.ts
EmilFattakhov: Good catch, reproduced with a test and fixed 👍

EmilFattakhov: The initial fix was over-engineered, simplified as per suggestion in the latest commit

jim-counter (approved):
LGTM!

PR #813, opened 2026-08-13
Treat a failed payment intent as terminal so the purchase flow stops polling

### LINKED ISSUE #634: Frontend polling never stops for failed intents
[apps/frontend/__tests__/unit/hooks/useTransactionConfirmation.spec.ts](https://github.com/autonomys/auto-drive/pull/633/files/ca24477fadc831afb028e0d516c9a6a6d27860be#diff-e317851d59b2137bb12136ff497c20aba10c95a1375b5c6cb356403672a493c4)

High Severity

The evaluateIntentStatus function (which mirrors the hook's polling decision logic) does not treat failed as a terminal state — it returns shouldContinue: true. The backend now marks dust payments as FAILED (a permanent, terminal status that will never change), but the frontend polling loop keeps retrying indefinitely. This leaves users stuck on a loading screen with no way to proceed after a too-small payment. The over_cap status is correctly handled as terminal, but failed was missed.

eastagiletracker:
This PR proposes treating a `failed` payment intent as a terminal state so the credit-purchase flow stops polling and surfaces the failure to the user instead of spinning on the confirmation screen forever. We include this PR work along with a full history of your repo at https://eastagiletracker.com/projects/266. You can sign in with your GitHub ID to claim ownership of the project.

Fixes #634.

## The defect

At current `main` (`12ab0654`), the backend permanently marks a dust payment `IntentStatus.FAILED` — `apps/backend/src/core/users/intents.ts` sets `status: IntentStatus.FAILED` and then groups it with `COMPLETED`/`OVER_CAP`/`EXPIRED` as "already processed", so it is terminal and will never change. The frontend polling loop in `apps/frontend/src/hooks/useTransactionConfirmation.ts`, however, only treats `completed`, `over_cap`, and the 410-expired error as terminal. A `failed` status matches none of those branches, so it falls through to `timer = setTimeout(poll, 2000)` and the loop retries forever. In `Step3_TransferTokens.tsx` that leaves the user on "Waiting for backend to update credits…" with the Continue button stuck on "Finalizing…" and no way to proceed after a too-small payment — the High-severity symptom described in the issue.

The existing unit test actually encoded this defect: `apps/frontend/__tests__/unit/hooks/useTransactionConfirmation.spec.ts` asserted `failed` → `shouldContinue: true`. It passed only because it tested a copy of the decision logic that mirrored the hook — the drift between that mirror and the real hook is what let the missing terminal case ship unnoticed.

## The change

The try-branch decision is lifted out of the hook into a small, dependency-free module `apps/frontend/src/hooks/intentPolling.ts` (`evaluateIntentStatus`), where `failed` is now a terminal outcome. The hook imports that function, adds an `isFailed` state that stops the loop, and returns it; `Step3_TransferTokens.tsx` surfaces `isFailed` with a "Payment failed" message and disables Continue, exactly as it already does for `over_cap` and `isExpired`. The test now imports the real `evaluateIntentStatus` instead of a mirror, so the hook and its tests share one source of truth and cannot drift apart again. The 410-expired catch-branch is unchanged.

Backward compatibility: the only public change is the additive `isFailed` field on the hook's return; the other consumer (`Step4_Success.tsx`) does not read the terminal flags and is unaffected.

## How I verified it (replayable in a minute)

Reproduction at HEAD and the regression guard both run under `yarn frontend test`:

```
# Revert only the new terminal line in intentPolling.ts (failed -> shouldContinue: true)
# and the regression test goes red, reproducing the poll-forever behaviour:
✕ marks failed and stops polling for "failed"
  - Expected  "shouldContinue": false
  + Received  "shouldContinue": true

# With the fix in place the whole frontend suite is green:
Test Suites: 7 passed, 7 total
Tests:       152 passed, 152 total
```

Baseline before the change had the same suites passing (plus two pre-existing suites that only fail until `@auto-drive/models` is built), so this adds no new failures. `tsc --noEmit` and `next lint` are both clean on the changed files.

## How this work was managed

I imported your repository's issues and pull requests into a live agile board (772 stories, 21 labels) and used it to plan and track this fix. This contribution maps to the imported story for #634:

- Story: https://eastagiletracker.com/projects/266/stories/156126
- Board: https://eastagiletracker.com/projects/266

board

If you'd rather not receive contributions like this, reply `no-more-prs` on this pull request and we won't open any further ones on your repositories.

---

**Lawrence W. Sinclair**
CEO / East Agile
[linkedin.com/in/lwsinclair/](https://linkedin.com/in/lwsinclair/)
[eastagile.com](https://eastagile.com)

### DIFF
--- a/apps/frontend/__tests__/unit/hooks/useTransactionConfirmation.spec.ts
+++ b/apps/frontend/__tests__/unit/hooks/useTransactionConfirmation.spec.ts
@@ -1,8 +1,10 @@
 /**
- * Unit tests for the intent-polling decision logic extracted from
- * useTransactionConfirmation.  These tests verify the correct terminal-state
- * handling for completed, over_cap, and expired intents without requiring a
- * React rendering environment.
+ * Unit tests for the intent-polling decision logic used by
+ * useTransactionConfirmation.  The try-branch decision now lives in a real,
+ * exported module (hooks/intentPolling) that both the hook and these tests
+ * import, so there is a single source of truth to verify — the earlier
+ * hook/test mirror is what let a terminal `failed` status go unhandled.  These
+ * tests still need no React rendering environment.
  *
  * IMPORTANT: The backend returns HTTP 410 Gone for expired intents instead of
  * an `{ status: 'expired' }` response body.  The `getIntent` call throws an
@@ -10,8 +12,11 @@
  * loop detects this via the catch block, not via `intent.status === 'expired'`.
  */
 
+import { evaluateIntentStatus } from 'hooks/intentPolling'
+
 // ---------------------------------------------------------------------------
-// Minimal ApiError replica (mirrors apps/frontend/src/services/api.ts)
+// Minimal ApiError replica (mirrors apps/frontend/src/services/api.ts) — used
+// only by the catch-branch tests below, which exercise thrown-error handling.
 // ---------------------------------------------------------------------------
 
 class ApiError extends Error {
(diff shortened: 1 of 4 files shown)


EmilFattakhov:
Thank you for your PR, it won't be merge as-is but we appreciate you taking a look at our project.

EmilFattakhov:
no-more-prs

